PR #20, opened 2015-05-02
Adds handling of already buffered data

AVVS:
Before that using already buffered data lead to max call stack errors

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -114,17 +114,25 @@ Pipeline.prototype.exec = function (callback) {
   }
 
   return promise.then(function () {
-    var data = '';
+    var bufferList = [];
+    var bufferLength = 0;
     var writePending = _this.replyPending =  _this._queue.length;
+
     // TODO optimize buffer
     var stream = {
       write: function (writable) {
-        data += writable;
+        if (!Buffer.isBuffer(writable)) {
+            writable = new Buffer(writable, 'utf-8');
+        }
+
+        bufferList.push(writable);
+        bufferLength += writable.length;
+
         if (!--writePending) {
           if (isCluster) {
-            _this.redis.slots[sampleSlot].stream.write(data);
+            _this.redis.slots[sampleSlot].stream.write(Buffer.concat(bufferList, bufferLength));
           } else {
-            _this.redis.stream.write(data);
+            _this.redis.stream.write(Buffer.concat(bufferList, bufferLength));
           }
         }
       }


luin:
Could you please provide an example to reproduce the max call stack errors?

AVVS:
I'm currently writing redis backend for https://github.com/pgte/abstract-skiff-persistence, which is basically a copy of https://github.com/pgte/skiff-level/blob/master/lib/index.js with a different backend, and without the proposed change `pipeline.exec(callback);` would crash when not using `hiredis` with max call stack error, and when using hiredis it would crash with smth saying that it can't read `.name` of the command or smth. I'm pretty sure it tries to parse the buffer, that was encoded as a msgpack and fails somehow during serialization/deserialization 

To simulate this you can run the tests as proposed in https://github.com/pgte/abstract-skiff-persistence - basic payload is a bit incorrect - `:` instead of `,` - but otherwise it's fine to test.

```js
import msgpack5 from 'msgpack5';
import SkiffPersistence from 'abstract-skiff-persistence';
import _ from 'lodash';
import async from 'neo-async';

let msgpack = msgpack5();

function encode(val) {
  return msgpack.encode(val).slice();
}

function decode(buf) {
  return msgpack.decode(buf);
}

class SkiffRedis extends SkiffPersistence {

    constructor(options={}) {

        super(options);

        let { redis, namespace } = options;

        // in case we use redis cluster, make sure that namespace is hashed
        this._namespace = namespace || '{skiff-redis}';

        this._redis = redis;

        this._redis.on('error', this._onError);

        this.nodes = {};
    }

    _onError(err) {
        this.emit('error', err);
    }

    _key(...parts) {
        if (parts.length === 0) {
            throw new Error('parts must include at least one value');
        }

        return this._namespace + parts.join('~');
    }

    _saveMeta(nodeId, state, callback) {
        let redis = this._redis;
        let pipeline = redis.pipeline();
        let log = state.log;
        let logEntries = log.entries;

        // do not enter hashtable mode
        state.log = _.without(log, ['entries']);

        pipeline.set(this._key(nodeId, 'meta'), encode(state));

        let minLogIndex = logEntries.length && logEntries[0].index || 0;
        let maxLogIndex = logEntries.length && logEntries[logEntries.length - 1].index || Infinity;
        let logNamespace = this._key(nodeId, 'logs');
        let cursor = 0;
        let maxReadIndex = 0;

        async.doUntil(
            function zscanRedis(next) {
                redis.zscanBuffer(logNamespace, cursor, 'count', 10, function zscanResponse(err, response) {
                    if (err) {
                        return next(err);
                    }

                    // update cursor
                    cursor = parseInt(response[0], 10);

                    // process log entries
                    let entries = response[1];
                    for (let i = 0, l = entries.length; i < l; i += 2) {
                        let index = parseInt(entries[i + 1], 10);
                        let entry = entries[i];

                        if (index > maxReadIndex) {
                            maxReadIndex = index;
                        }

                        if (index < minLogIndex || index > maxLogIndex) {
                            pipeline.zrem(logNamespace, entry);
                        } else {
                            let correspondingNewEntry = logEntries[index - minLogIndex];
                            let decodedEntry = decode(entries[i]);
                            if (correspondingNewEntry.uuid !== decodedEntry.uuid) {
                                pipeline.zrem(entry);
                                pipeline.zadd(logNamespace, correspondingNewEntry.index, encode(correspondingNewEntry));
                            }
                        }

                    }

                    next();

                });
            },

            function isIterationComplete() {
                return cursor === 0;
            },

            function saveMeta(err) {
                if (err) {
                    return callback(err);
                }

                if (maxReadIndex < maxLogIndex) {
                    logEntries.slice(maxReadIndex - minLogIndex + 1).forEach(function (entry) {
                        pipeline.zadd(logNamespace, entry.index, encode(entry));
                    });
                }

                pipeline.exec(callback);
            }

        );
    }

    _loadMeta(nodeId, callback) {

        this._redis
            .pipeline()
            .getBuffer(this._key(nodeId, 'meta'))
            .zrangebyscoreBuffer(this._key(nodeId, 'logs'), '-inf', '+inf')
            .exec(function (___, results) {
                // err is always null

                let [err, meta] = results[0];
                let [logerr, logEntries] = results[1];

                if (err || logerr) {
                    return callback(err || logerr);
                }

                if (meta) {
                    meta = decode(meta);
                    meta.log.entries = logEntries.map(decode);
                }

                callback(null, meta);

            });

    }

    _applyCommand(nodeId, commitIndex, command, callback) {

    }

    _lastAppliedCommitIndex(nodeId, callback) {

    }

    _saveCommitIndex(nodeId, commitIndex, callback) {

    }

    _createReadStream(nodeId) {

    }

    _createWriteStream(nodeId) {

    }

    _removeAllState(nodeId, callback) {

    }

    _close(callback) {
        this._redis.disconnect();
        this._redis.removeListener('error', this._onError);
        callback();
    }

}


export default SkiffRedis;

```

AVVS:
The code I've posted is by no means complete, but this is where the bug appeared

luin:
I ran the tests successfully and get the same error as you. I create a new branch to fix the error: https://github.com/luin/ioredis/commit/b81b625a4ab93d71e997f32a291b2c956710ff6a. It's faster than yours when all arguments are strings. However it seems that this change still doesn't pass all your tests. Could you please check it out and see whether it works correctly?

AVVS:
I'm still working on the code, so its fine that it doesn't pass my tests. As long as you got rid of the error this should be great

luin:
Released in 1.1.3. Thank you for the pull request!

PR #40, opened 2015-05-20
Add missing Promise definition in node 0.10.x

AVVS:


AVVS:
@luin can you update package version and release this on npm? Kind of using that in production :)

luin:
Done. I just ran the tests and it turns out all tests are passed in node 0.10.x. I forget the reason why I state ioredis only works with Node.js whose version >= 0.11.16.

PR #53, opened 2015-06-01
fix typo

xroot:


luin:
It's a glob pattern. `?` means any char.

PR #63, opened 2015-06-04
Resolves a bug with cluster where a subscribe is sent to a disconnected node

devaos:
The issue came up when connecting to a load balancer, sitting in front of the redis cluster nodes, and attempting to subscribe as soon as the 'ready' event was fired.  The load balancer's IP:port is not in redis cluster's list of nodes so the connection was disconnected and removed, causing the subscribe to be sent to the disconnected node.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -162,10 +162,11 @@ Cluster.prototype.createNode = function (port, host) {
 
     var _this = this;
     this.nodes[key].once('end', function () {
-      if (_this.subscriber === _this.nodes[key]) {
+      var deadNode = _this.nodes[key];
+      delete _this.nodes[key];
+      if (_this.subscriber === deadNode) {
         _this.selectSubscriber();
       }
-      delete _this.nodes[key];
       if (Object.keys(_this.nodes).length === 0) {
         _this.setStatus('close');
       }
@@ -179,6 +180,10 @@ Cluster.prototype.selectRandomNode = function () {
 };
 
 Cluster.prototype.selectSubscriber = function () {
+  if (Object.keys(this.nodes).length === 0) {
+    this.subscriber = null;
+    return;
+  }
   this.subscriber = this.selectRandomNode();
   if (this.subscriber.status === 'wait') {
     this.subscriber.connect().catch(function () {});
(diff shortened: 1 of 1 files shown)


luin:
When a node is disconnected(https://github.com/luin/ioredis/blob/master/lib/cluster.js#L407), ioredis will re-select a new subscriber automatically(https://github.com/luin/ioredis/blob/master/lib/cluster.js#L164). So you don't need to call `selectSubscriber` manually.

devaos:
Oh, I see what you mean.  However this commit did resolve the issue we were having.  I'll comment shortly with the debug logs so you can see what was happening.

devaos:
Here is a snippet of the debug output before the patch.  I added a couple console lines to clarify which command was being sent and whether the stream was writable.

      ioredis:redis status[127.0.0.1:6381]: close -> end +1ms
      ioredis:redis status[10.240.14.47:6375]: connect -> ready +0ms
      ioredis:redis status[10.240.46.240:6375]: connect -> ready +0ms
      ioredis:redis status[10.240.75.76:6375]: connect -> ready +1ms
      ioredis:redis status[127.0.0.1:6381]: ready -> close +0ms
      ioredis:connection skip reconnecting since the connection is manually closed. +0ms
      ioredis:redis status[127.0.0.1:6381]: close -> end +0ms
      ioredis:cluster getting slot cache from 10.240.46.240:6375 +2s
    sendCommand: command.name=cluster
    sendCommand: stream.writable=undefined
    sendCommand: this.stream.writable=true
      ioredis:redis write command[0] -> cluster(slots) +1ms
    sendCommand: command.name=subscribe
    sendCommand: stream.writable=undefined
    sendCommand: this.stream.writable=false
      ioredis:cluster getting slot cache from 10.240.46.240:6375 +2s
    sendCommand: command.name=cluster
    sendCommand: stream.writable=undefined
    sendCommand: this.stream.writable=true
      ioredis:redis write command[0] -> cluster(slots) +0ms
    sendCommand: command.name=subscribe
    sendCommand: stream.writable=undefined
    sendCommand: this.stream.writable=false

devaos:
I updated the commit with what I believe is a more appropriate fix based on your feedback.  Thanks

luin:
Awesome! Thank you for the patch!

PR #69, opened 2015-06-10
Adds ReadOnly mode for Standalone/Cluster mode

nakulgan:
- Cluster slots are now served with both Master and Slave instances.

The purpose is to use the client in a `readonly` mode to scale reads.

Each client individual issues a `readonly` once `ready`.
Each node in the cluster is connected to, maintaining a pool of Master nodes and SlaveNodes.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -164,6 +164,12 @@ Cluster.prototype.createNode = function (port, host) {
     }));
 
     var _this = this;
+    if (this.options.readOnly) {
+      this.nodes[key].once('ready', function () {
+        debug('sending readonly to %s', key);
+        _this.nodes[key].readonly();
+      });
+    }
     this.nodes[key].once('end', function () {
       var deadNode = _this.nodes[key];
       delete _this.nodes[key];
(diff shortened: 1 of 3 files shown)


luin:
Awesome! I'll check it out ASAP.

lib/redis.js
```diff
@@ -163,7 +164,8 @@ Redis.defaultOptions = {
   enableReadyCheck: true,
   autoResubscribe: true,
   autoResendUnfulfilledCommands: true,
-  lazyConnect: false
+  lazyConnect: false,
+  readOnly: false
```
luin: Since `readOnly` only works in cluster mode, we'd better only have it in cluster.js and listen to the `ready` event in the `createNode` method of cluster.js

lib/cluster.js
```diff
@@ -394,20 +410,32 @@ Cluster.prototype.getInfoFromNode = function (redis, callback) {
...
+      var master = items.shift();
+      var masterNodeKey = master[0] + ':' + master[1];
+      var masterNode = _this.createNode(master[1], master[0]);
+      _this.masterNodes[masterNodeKey] = masterNode;
+      allNodes.push(masterNode);
+      delete oldNodes[masterNodeKey];
+      items.forEach(function(item) {
+        var host = item[0];
+        var port = item[1];
+        allNodes.push(_this.createNode(port, host));
```
luin: I think we don't need to connect to the slaves when `readOnly` is disabled.

luin:
Just added some comments

nakulgan:
Cool, I agree on both of them. I'll update the PR with changes accordingly later today.

PR #83, opened 2015-06-28
Pass stream via sync callback

### LINKED ISSUE #80: If network goes down during connection to a redis instance, process crashes
I wouldn't say that this is really going to happen in the real-life, and probably if it did, the machine is not reachable from the outside, but I think it's not the worst idea to report.

To replicate this - start redis process, make sure it tries to reach something via wi-fi (ip might not exist, but should be valid one), turn off wi-fi and enjoy the crash

```js
events.js:141
      throw er; // Unhandled 'error' event
            ^
Error: connect ENETUNREACH 10.10.100.80:6998 - Local (0.0.0.0:53496)
    at Object.exports._errnoException (util.js:846:11)
    at exports._exceptionWithHostPort (util.js:869:20)
    at connect (net.js:834:14)
    at lookupAndConnect (net.js:927:5)
    at Socket.connect (net.js:896:5)
    at Object.exports.connect.exports.createConnection (net.js:61:35)
    at Connector.connect (/Users/vitaly/projects/mservice-utils/node_modules/arklogger/node_modules/ioredis/lib/connectors/connector.js:28:34)
    at Redis.<anonymous> (/Users/vitaly/projects/mservice-utils/node_modules/arklogger/node_modules/ioredis/lib/redis.js:228:20)
    at tryCatcher (/Users/vitaly/projects/mservice-utils/node_modules/arklogger/node_modules/ioredis/node_modules/bluebird/js/main/util.js:24:31)
    at Promise._resolveFromResolver (/Users/vitaly/projects/mservice-utils/node_modules/arklogger/node_modules/ioredis/node_modules/bluebird/js/main/promise.js:427:31)
    at new Promise (/Users/vitaly/projects/mservice-utils/node_modules/arklogger/node_modules/ioredis/node_modules/bluebird/js/main/promise.js:53:37)
    at Redis.connect (/Users/vitaly/projects/mservice-utils/node_modules/arklogger/node_modules/ioredis/lib/redis.js:214:10)
    at null._onTimeout (/Users/vitaly/projects/mservice-utils/node_modules/arklogger/node_modules/ioredis/lib/redis/event_handler.js:76:12)
    at Timer.listOnTimeout (timers.js:89:15)
```

AVVS:
That way we avoid errors firing between error listener is defined. Fixes process crash during network device up/down. Fixes #80

### DIFF
--- a/lib/connectors/connector.js
+++ b/lib/connectors/connector.js
@@ -26,9 +26,7 @@ Connector.prototype.connect = function (callback) {
   }
 
   var stream = this.stream = net.createConnection(connectionOptions);
-  process.nextTick(function () {
-    callback(null, stream);
-  });
+  callback(null, stream);
 };
 
 module.exports = Connector;


AVVS:
https://github.com/luin/ioredis/blob/master/lib/connectors/sentinel_connector.js#L73-L74 - for sentinel this is already a sync operation

luin:
Your solution makes sense. However since `SentinelConnector#connect` invokes callback asynchronously, we'd better keep the interface same in `Connector#connect`, so `process.nextTick` is used here.

The problem is we should keep `net.createConnection` and `callback` running in the same event loop to ensure all connection errors can be caught. I submitted a pull request #84 which hopefully fixes the issue. Could you please help me review it? :-)

luin:
Closing the issue since #84 should fix it.

PR #84, opened 2015-06-29
Fix connection error emitting before listening to `error` event.

luin:


AVVS:
This looks good, gonna test it shortly and report if problem goes away

PR #88, opened 2015-07-03
pmessage now properly returns pattern, channel, message

klinquist:
In Redis Cluster mode, arg3 (message) was not getting passed to pmessage/pmessageBuffer.


luin:
Awesome! Thanks for the patch.

PR #105, opened 2015-07-21
Add support for transparent key prefixing. Close #95

### LINKED ISSUE #95: Implement transparent key prefixing
Hi, I noticed that you have this feature on the roadmap, and I'd like to help implement it. Do you have any ideas about how it will work?

dguo:


lib/command.js
```diff
@@ -43,6 +43,11 @@ function Command(name, args, options, callback) {
   this.replyEncoding = options && options.replyEncoding;
   this.errorStack = options && options.errorStack;
   this.args = args ? _.flatten(args) : [];
+  if (options && options.keyPrefix) {
```
AVVS: this is costly to access options + keyPrefix each time, imo it shoud be cached in a var. Also options should be always an object?

lib/command.js
```diff
@@ -106,22 +129,37 @@ Command.prototype.getKeys = function () {
...
         break;
       case 'zunionstore':
       case 'zinterstore':
+        if (transform) {
+          this.args[0] = transform(this.args[0]);
+        }
         this._keys.push(this.args[0]);
         keyStop = parseInt(this.args[1], 10) + 2;
         for (i = 2; i < keyStop; ++i) {
+          if (transform) {
```
AVVS: check in a cycle is not the best thing, imo we should create a function which is either a noop (gives back the input) or the transformed input. that way check is only performed once. this an be applied to all the other if (transform)

luin: Yes, and I prefer the latter way to create a function for `transform` and `_keys.push`:
```javascript
function (arg, keys)
  keys.push(transform ? transform(arg) : arg);
}
```
However I'm not sure whether that would impact the performance.

lib/command.js
luin: There's no place we can cache the `keyPrefix` option. `options` is possible to be `undefined` since it's optional. We may put

```javascript
if (typeof options === 'undefined') {
  options = {};
}
```

at the top of the constructor so that we won't need to test options everytime.

luin:
Awesome! Just checked out the changes and they're really nice! I'm going to do more tests and we'll soon have this feature. 🍺

lib/command.js
dguo: @luin, I don't think that would work because the tranformation needs to be saved in `this.args`.

How about doing what @AVVS suggested in this way:
```javascript
// at the top
if (typeof transform !== 'function') {
  transform = function (key) {
    return key;
  };
}
```
And then I can remove all the checks and always apply the transform.

luin: Aha, you're right. The way works for me.

dguo:
Okay, I made the two changes.

lib/command.js
```diff
@@ -39,10 +39,18 @@ var commands = require('../commands');
...
+  if (typeof options === 'undefined') {
+    options = {};
+  }
   this.name = name;
-  this.replyEncoding = options && options.replyEncoding;
-  this.errorStack = options && options.errorStack;
+  this.replyEncoding = options.replyEncoding;
+  this.errorStack = options.errorStack;
   this.args = args ? _.flatten(args) : [];
+  if (options.keyPrefix) {
```
AVVS: should be:

var keyPrefix = options.keyPrefix;
if (keyPrefix) {
  this._iterateKeys(function (key) {
    return keyPrefix + key;
  });
}

Otherwise each iteration it will look for a pointer from options to .keyPrefix and we want to avoid it

dguo: Good catch. Thanks.

luin:
Released in 1.7.0 🍺

dguo:
:smile: Awesome! Thanks for the quick responses.

PR #109, opened 2015-07-23
Auto resend unfulfilled commands for cluster

klinquist:
Attempting to add autoResendUnfulfilledCommands in cluster mode.  Please test :)

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -44,6 +44,8 @@ function Cluster (startupNodes, options) {
   this.retryAttempts = 0;
   this.options = _.defaults(options || {}, this.options || {}, Cluster.defaultOptions);
   this.offlineQueue = [];
+  this.prevCommandQueue = [];
+  this.commandQueue = [];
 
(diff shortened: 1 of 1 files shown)


luin:
`autoResendUnfulfilledCommands` option is passed to the `Redis` instances of the cluster(https://github.com/luin/ioredis/blob/master/lib/cluster.js#L167), so it should work. I tried the following code:

```javascript
var cluster = new Redis.Cluster([
  { port: 6380 },
  { port: 6381 },
  { port: 6382 }
]);

function loop(err, res) {
  (err || res) && console.log(err, res);
  cluster.blpop('qqq', 0, loop);
}

loop();
```

When I `lpush` a item to `qqq`, it prints it, and when I stop the cluster and restart it, `lpush` a item to `qqq` again, it still prints it correctly.

klinquist:
d'oh.  I thought it was autoResend that wasn't working in cluster mode - turns out it's auto-resubscribe.

PR #114, opened 2015-07-27
Fixes status debug message in case of Unix socket path

tkalfigo:
When server uses Unix socket, the debug message of the status changes
still indicated [host:port] even though the connector was using the
socket's path correctly.


luin:
Merged! Thank you for the patch 😄

PR #119, opened 2015-08-03
Handle watch-exec rollback correctly

suprememoocow:
When using [WATCH/EXEC](http://redis.io/topics/transactions) transactions in Redis, if the transaction rolls back (due to one of the watched keys being modified), the EXEC command does not return a result (instead of returning an array).

Currently this leads to `utils/wrapMultiResult` failing with a `TypeError: Cannot read property 'length' of null`. 

This PR adds tests for WATCH/EXEC (normal and rollback) and fixes the problem by making `wrapMultiResult` a little more defensive.

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -70,6 +70,9 @@ exports.convertBufferToString = function (value, encoding) {
  * @private
  */
 exports.wrapMultiResult = function (arr) {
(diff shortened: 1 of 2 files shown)


luin:
Cool! Released in 1.7.3 🍺

suprememoocow:
Wow, that was quick. Thanks so much for your help.

PR #120, opened 2015-08-03
Add clarification about commands arguments

lpinca:
This tweaks a comment to make clear that when the number of arguments is big, using an array results in better performances.

Correct me if I'm wrong.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,8 @@ redis.get('foo').then(function (result) {
   console.log(result);
 });
 
-// Arguments to commands are flattened, so the following are the same:
+// Arguments to commands are flattened, so the following are the same.
+// When the number of arguments is big the latter is slightly faster.
 redis.sadd('set', 1, 3, 5, 7);
 redis.sadd('set', [1, 3, 5, 7]);
 ```


luin:
Actually they should have similar performance since ioredis passes arguments to the `flatten` method of Lodash without checking whether they are arrays: https://github.com/luin/ioredis/blob/master/lib/command.js#L48

lpinca:
Yes, but using an array of arguments will "remove" this [loop](https://github.com/luin/ioredis/blob/1e7961e9c91155fc780e86b64ca5323e9099234d/lib/commander.js#L104-L107).

luin:
Sounds interesting. I wrote a tests for it: https://gist.github.com/luin/d06770c736866e0571d0. The result shows the former form is little faster. It looks like `_.flatten` would be faster if the arguments are all plain strings/numbers: https://github.com/lodash/lodash/blob/3.10.0/lodash.src.js#L2122

lpinca:
That's indeed interesting, I would have never expected that.
Feel free to close this.

lpinca:
Here is my own [test case](https://gist.github.com/lpinca/b75c5fa4321a89e9eed2) which confirms my original assumption.
The difference is quite noticeable:
```
macbook:bench luigi$ node index.js 
without copy loop x 50,886 ops/sec ±0.64% (96 runs sampled)
with copy loop x 33,855 ops/sec ±0.38% (95 runs sampled)
Fastest is without copy loop
```

luin:
I ran your test case and got the similar result as yours. I though there may be too few arguments in my previous test case, so I updated [it](https://gist.github.com/luin/d06770c736866e0571d0) but the results between the two forms are still slight. Don't know why.

lpinca:
I don't know either, but it doesn't matter, I doubt that someone will ever use an amount of arguments that will make a difference.

AVVS:
It could be that array wasnt optimized (diff types of values, not all strings for instance) or some other similar case

PR #126, opened 2015-08-11
Previous Subscription State Error

fracmak:
This fixes an issue where we restored the previous state AFTER we ran the offline and previous command queue, which could cause us to miss unsubscribe requests. Unit test added for this scenario.


luin:
Makes sense. Thanks for the pr! 😄

fracmak:
Thank you for merging so quickly, any idea when you'll package it up?

luin:
I'm doing a little more testing and will release a new version later today.

PR #128, opened 2015-08-12
Fix empty pipeline not resolving as empty array

### LINKED ISSUE #127: Empty pipeline doesn't resolve when exec() is called
If you create an empty pipeline, it never resolves after calling `exec()`

Here's a failing test case to demonstrate the issue:

```js
// test/functional/pipeline.js
describe('pipeline', function () {
  // ...

  // This test fails with a timeout error
  it('should resolve on an empty pipeline', function(done) {
    var redis = new Redis();
    redis.pipeline().exec(function(err, results) { // This never gets called
      expect(err).to.eql(null);
      expect(results).to.eql([]);
      done(); // This never gets called
    });
  });

  // ...
});
```

A potential solution is checking if `this._queue` is empty in `Pipeline.prototype.exec` and calling `this.resolve([])`. I'm interested to hear any thoughts on this. In particular, whether this should be considered an error, or what the return value should be (an empty array? null? ...?).

Thanks!

phlip9:
Fixes #127

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -185,6 +185,9 @@ Pipeline.prototype.exec = function (callback) {
     this.nodeifiedPromise = true;
     this.promise.nodeify(callback);
   }
+  if (_.isEmpty(this._queue)) {
+    this.resolve([]);
+  }
   var pipelineSlot;
   // Check whether scripts exists and get a sampleKey.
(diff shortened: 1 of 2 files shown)


luin:
Thanks for the pr 😄

PR #132, opened 2015-08-16
Fix for allNodes array containing nodes not serving the specified slot.

henstock:
The allNodes array was accumulating the nodes for all previously processed entries in the `cluster slots` result. This meant read-only clients sometimes repeatedly targeting an incorrect node, especially in larger clusters.


luin:
Nice catch! Thanks for the pull request!

PR #146, opened 2015-09-12
Remove duplicate node version from .travis.yml

lpinca:
This cleans up `.travis.yml` using a simpler way to specify the node version and removes a duplicated version (`iojs-v2`).


luin:
Is `2` the same as `iojs-v2`?

lpinca:
Yes.
```bash
$ nvm install 2
######################################################################## 100,0%
WARNING: checksums are currently disabled for io.js
Now using io.js v2.5.0 (npm v2.13.2)
```

luin:
That's good to know. :laughing:

PR #154, opened 2015-09-22
Added transformer support to scripting.

paskakaska:


luin:
Awesome! I'll check it out the weekend.

luin:
It might be a little confused that `setArgumentTransformer` and `setReplyTransformer` affect all `Redis` instances, whereas custom commands are defined within a single instance.

As for me, it would be better to support custom commands transforming by adding a property in `defineCommand` method, for example:

```javascript
redis.defineCommand('test', {
  lua: 'return {KEYS[1],KEYS[2],ARGV[1],ARGV[2]}',
  transform: {
    argument: function (obj) {
      return [obj.keys.length].concat(obj.keys).concat(obj.values);
    }
  }
});

redis.test({ keys: ['a'], values: ['b'] });
```

However I doubt whether it's necessary to support this feature, since there's a more straightforward way to do this:

 ```javascript
redis.defineCommand('test', {
  lua: 'return {KEYS[1],KEYS[2],ARGV[1],ARGV[2]}'
});

var oriTest = redis.test;
redis.test = function (obj) {
  oriTest.apply(redis, [obj.keys.length].concat(obj.keys).concat(obj.values));
};
redis.test({ keys: ['a'], values: ['b'] });
```

**Comment by bradennapier** [comment]
I think the transform argument makes more sense.  Custom commands having transformers seems like the #1 need for the transformers to have a public api in the first place.  

Could also simply have an instance Command property

```js
import Redis from 'ioredis';
const redis = new Redis();

redis.Command....
```

PR #155, opened 2015-09-23
Use given `connectTimeout` for sentinel connections

ColmHally:
This change allows sentinel connections to use the user-specified connection timeout instead of the hard-coded value of 2000 milliseconds. In cases where a `connectTimeout` is not specified, the default value in `Redis.defaultOptions` will be used (currently 3000 millis).


luin:
LGTM. Thanks for the pr! :laughing:

ColmHally:
Thanks @luin - I appreciated the quick response!

PR #157, opened 2015-09-24
Update options host & port with resolved Redis server

ColmHally:
When `ioredis` connects to a sentinel, retrieves the host and port of the master server, it stills outputs `localhost:6379` in the debug output, which is usually incorrect/misleading.

Even with debug output disabled, it would be nice to be able to print the address of the connected redis server when the `#ready` event fires, e.g.:
```javascript
redis.on('ready', function () {
  console.log('Connected to Redis @ %s:%s', redis.options.host, redis.options.port);
});
```

Now, maybe `options` is not the best place to store this, but since this is what's used for the debug/status messages (e.g. `ioredis:redis status[192.168.33.10:26379]: connecting -> connect +1ms`), I felt this was the simplest place to put it.

### DIFF
--- a/lib/redis.js
+++ b/lib/redis.js
@@ -213,7 +213,9 @@ Redis.prototype.parseOptions = function () {
  * @private
  */
 Redis.prototype.setStatus = function (status, arg) {
-  debug('status[%s]: %s -> %s', (this.options.path ? this.options.path : (this.options.host + ':' + this.options.port)), this.status || '[empty]', status);
+  var remoteHost = this.stream && this.stream.remoteAddress || this.options.host;
+  var remotePort = this.stream && this.stream.remotePort || this.options.port;
+  debug('status[%s:%s]: %s -> %s', remoteHost, remotePort, this.status || '[empty]', status);
   this.status = status;
   process.nextTick(this.emit.bind(this, status, arg));
 };


luin:
I agree with you that the debug output is a little misleading when using with sentinel. However, changing the options would cause ioredis to reconnect to the wrong node (connect to the master node directly without asking the sentinels first).

I just did some checks and I think it might solve the problem to get the node addresses from the socket object instead of the options:

```javascript
var address = this.stream.address();
debug('status[%s]: %s -> %s', address.address + ':' + address.port, this.status || '[empty]', status);
```

But we should check if the socket object (aka. `this.stream`) existing before fetch its address.

ColmHally:
Yes, that's a better idea, but `this.stream.address()` will not work (see [here](https://nodejs.org/api/net.html#net_socket_address) - `address()` returns the *bound* address).

`this.stream.remoteAddress` and `this.stream.remotePort` would better suit our intentions here.

ColmHally:
@luin made some changes that use `remoteAddress` etc. Would you mind taking a look?

luin:
It seems if conneting via unix socket/pipe (`new Redis({ path: '/tmp/redis.sock' })`), `socket.remoteAddress` would be `undefined`, leading to the wrong debug log.

We may check the `path` option first:

```javascript
Redis.prototype.setStatus = function (status, arg) {
  var address = this.options.path;
  if (!address) {
    var remoteHost = this.stream ? this.stream.remoteAddress : this.options.host;
    var remotePort = this.stream ? this.stream.remotePort : this.options.port;
    address = remoteHost + ':' + remotePort;
  }

  debug('status[%s]: %s -> %s', (address, this.status || '[empty]', status);
  this.status = status;
  process.nextTick(this.emit.bind(this, status, arg));
};
```

PR #162, opened 2015-09-28
Discard slave if flagged with s_down or o_down

mtlima:
Flags s_down and o_down are not being taken in account when connecting.


luin:
Great! Thank you for the patch!

Issue #164, opened 2015-09-29
How do I know if I provided an incorrect password?


naugtur:
`new Redis(options)` with options.password being incorrect or missing, doesn't throw (not surprising, AUTH is async) and it doesn't emit `end` nor `close`, nor `error`.

How do I know if the AUTH command that happens in the background failed?

[edit]
It works if I call `.auth` myself, but otherwise, the error is swallowed. I even found one location which might be responsible for swallowing it: https://github.com/luin/ioredis/blob/master/lib/redis.js#L124

But switching to lazyConnect and connecting on my own would also resolve happily without reporting auth issues, so that's not all there is to it.

[one more edit]
Unfortunately, calling `.auth` on my own is not possible, because a "Ready check" is made, and it fails because auth was not made yet... So no workaround here.

luin:
Yes, we may emit an event when the password is wrong. May be `auth error`.

naugtur:
So for the workaround to work:

`enableReadyCheck: false` in options is required, then:
```js
redisInstance.on("connect", function () {
    redisInstance.auth("pawsword", function (err, res) {
        if (err) {
            console.error(err);
            redisInstance.disconnect()
            process.exit();

        }

    })
});
```

But I'd appreciate a way to catch the initial built-in auth failure. 

```js
redisInstance.on("error",function(authError){})
```
^would be the way to go imho

luin:
`error` event is a special event that would crash the app if there's no listener for it. I would go for another event name like `auth error`. :laughing:

naugtur:
Well, I'm on this camp that prefers when errors have to be handled ;) but ok.
If I can suggest something I'd not put a space in the event name for the sake of people coming from jQuery where it means 2 events ;)

PR #165, opened 2015-09-29
Emit auth error for wrong password

luin:
#164

### DIFF
--- a/lib/redis/event_handler.js
+++ b/lib/redis/event_handler.js
@@ -12,9 +12,9 @@ exports.connectHandler = function (self) {
 
     // AUTH command should be processed before any other commands
     if (self.condition.auth) {
-      self.auth(self.condition.auth, function (err, res) {
-        if (err && err.message.match('no password is set')) {
-          console.warn('`auth` is specified in the client but not in the server.');
+      self.auth(self.condition.auth, function (err) {
+        if (err) {
+          self.emit('auth error', err);
         }
       });
     }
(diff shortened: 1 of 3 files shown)


lib/redis/event_handler.js
```diff
@@ -12,9 +12,9 @@ exports.connectHandler = function (self) {
 
     // AUTH command should be processed before any other commands
     if (self.condition.auth) {
-      self.auth(self.condition.auth, function (err, res) {
-        if (err && err.message.match('no password is set')) {
-          console.warn('`auth` is specified in the client but not in the server.');
+      self.auth(self.condition.auth, function (err) {
+        if (err) {
+          self.emit('auth error', err);
```
naugtur: could it be `auth-error` or `authError` or `authfailed` ?

luin: `authError` looks good to me

naugtur:
That doesn't look like a change that'd make all tests fail...

luin:
The test fails because it expects ioredis would print a warning message if the server doesn't require a password: https://github.com/luin/ioredis/blob/master/test/functional/auth.js#L44. I'm thinking whether we can remove this test.

naugtur:
Now that the user is in control documenting the auth error event should be enough. 
For an app that doesn't care about error handling ignoring this seems like the way to go ;)

luin:
Released with 1.9.1 :rocket:

naugtur:
:+1:

Issue #96, opened 2015-07-10
Flushall in Redis Cluster


skreutzberger:
It may just be a lack of understanding but I could not find the answer in browsing the unit tests of a Redis Cluster (latest Redis version).

It seems that a flushall() command which is sent to a cluster connection does take some time to populate or is not properly handled in terms of callback / waiting.

For example, I have the following function which should flush Redis (it is used insde the AfterEach function of my Mocha unit tests) by using Promises (I love Bluebird Promises!):

```js
function flushDBAsync(cluster) {
    log.debug("flushing Redis ...");
    return cluster.flushall()
        .then(function(val) {
            if (val === "OK") {
                log.debug("flushing finished.");
                log.debug("disconnecting from Redis ...");
                return cluster.disconnect();
            } else {
                var msg = "flushing returned with " + val;
                return Promise.reject(msg);
            }
        })
        .then(function() {
            log.debug("disconnecting finished.");
            return true;
        })
        .catch(function(err) {
            log.error("Error in flushDB: " + err);
            return false;
        });
}
```

The function is working always 100% perfectly if I am using a "normal" Redis connection (not a cluster). But when used with a cluster connection then the flushing is not fully completed (populated?) after the return and there are still some keys existing in the cluster.

For now my hacky solution is to wait 2 seconds after I called the function but that is really slowing down my unit tests: `return flushDBAsync(redisConnection).delay(2000);`

How should I properly do a flushall on a Redis Cluster consisting of 3 masters and 3 slaves (default from Redis website)? Redis Cluster is really awesome and I would love to use it with ioredis.

Thanks for looking into that surely very special case!

Sebastian

luin:
Hi Sebastian, ioredis only sends each command to one specified node. If the command has a key, ioredis would calculate the slot of it and send the command to the node serving the slot. However if the command doesn't has a key(e.g. `info`, 'flushall` etc.), it'll be sent to a **random** node of the cluster.

If you want to flush all data of the cluster, you should send `flushall` to all master nodes, which can be accessed via `cluster.masterNodes`(https://github.com/luin/ioredis/blob/master/test/functional/cluster.js#L839):

```javascript
for (var key in cluster.masterNodes) {
  if (!cluster.hasOwnProperty(key)) continue;

  var node = cluster.masterNodes[key];
  node.flushall();
}
```

However it make me confused that you said waiting 2 seconds would fix this, if you can confirm this, I'll definitely look into it.

skreutzberger:
[...]

PR #166, opened 2015-09-29
extends cluster with .mastersExec, which executes a given command on every already discovered master

AVVS:
* `.mastersExec(command, [callback)` utility
* `.flushCluster([callback])`
* `.flushClusterDb([callback])`

It's often needed to do something like, so why not add a helper?

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -496,6 +496,38 @@ Cluster.prototype.getInfoFromNode = function (redis, callback) {
   }, 1000));
 };
 
+Cluster.prototype.mastersExec = function () {
+  var masterNodes = _.values(this.masterNodes);
+
+  var argumentsLength = arguments.length;
+  var hasCallback = typeof arguments[argumentsLength - 1] === 'function';
+
+  var argsLength = hasCallback ? argumentsLength - 1 : argumentsLength;
+  var args = new Array(argsLength - 1);
+  for (var i = 1; i < argsLength; i++) {
+    args[i - 1] = arguments[i];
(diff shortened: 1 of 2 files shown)


AVVS:
Could be a basis for #96

lib/cluster.js
```diff
@@ -496,6 +496,38 @@ Cluster.prototype.getInfoFromNode = function (redis, callback) {
...
+
+  return promise;
+}
+
+Cluster.prototype.flushCluster = function (callback) {
+  return this.mastersExec('flushall', callback);
+};
+
+Cluster.prototype.flushClusterDb = function (callback) {
+  return this.mastersExec('flushdb', callback);
```
luin: As for me, syntax like `cluster.to('masters').flushall();` seems to be a little better that we don't need to give every Redis command a cluster version method.

luin: Or maybe just give users an array containing the master nodes (the same as `_.values(this.masterNodes)`), so they can send commands to every master node easily:

```javascript
cluster.masters.map(node => node.flushall()).then().catch();
```

AVVS: First idea seems cleaner, second would require a bit more code all the time. However we would want to make a single instance of a command distributor like this and I cant quite come up with the design just yet - do you have any particular implementation of this in mind?

luin: Not yet. Personally I'd like to go for the latter one since although it requires a bit more code, users are easier to figure out how do things work. I'll think about it however.

PR #172, opened 2015-10-08
Fix tls connect event is "secureConnect"

jeffjen:
[See API page](https://nodejs.org/api/tls.html#tls_class_tls_tlssocket_1)

TLS stream upon connection emits "secureConnect" event.

Tested against *stunnel*

### DIFF
--- a/lib/redis.js
+++ b/lib/redis.js
@@ -256,13 +256,14 @@ Redis.prototype.connect = function (callback) {
         reject(err);
         return;
       }
+      var CONNECT_EVENT = _this.options.tls ? 'secureConnect' : 'connect';
 
       _this.stream = stream;
       if (typeof _this.options.keepAlive === 'number') {
         stream.setKeepAlive(true, _this.options.keepAlive);
       }
 
-      stream.once('connect', eventHandler.connectHandler(_this));
+      stream.once(CONNECT_EVENT, eventHandler.connectHandler(_this));
       stream.once('error', eventHandler.errorHandler(_this));
       stream.once('close', eventHandler.closeHandler(_this));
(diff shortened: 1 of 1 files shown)


luin:
Doesn't tls stream also emit `connect` event?

jeffjen:
Well it looks like v4.1.2 had the tls connection emit "connect" as well as "secureConnect"

For the node build from https://deb.nodesource.com/setup, which had version `v0.10.40`, it doesn't emit "connect", only emits "secureConnect".

Consider this fix for backwards compatibility then.

lib/redis.js
```diff
@@ -256,13 +256,17 @@ Redis.prototype.connect = function (callback) {
         reject(err);
         return;
       }
+      var CONNECT_EVENT = 'connect';
+      if (_this.options.tls) {
+          CONNECT_EVENT = 'secureConnect';
+      }
```
luin: It would be better to use `var CONNECT_EVENT = _this.options.tls ? 'secureConnect' : 'connect';` instead

luin:
Good to know:-). According to the [issue](https://github.com/nodejs/node-v0.x-archive/issues/6903), it seems that `secureConnect` is more preferable than `connect`. I just wrote a comment

jeffjen:
Pushed a change to do tertiary assignment to CONNECT_EVENT.

I know its not recommended to connect via TLS to REDIS since the support of it is minimal (need SSL proxy to do it), but hopefully this will payout in the long run.

luin:
Thank you for the patch! Merged :rocket:

luin:
Shipped in 1.10.0

PR #177, opened 2015-10-15
allow the connection to be named even if susbscribed and reconnected

cgiovanacci:
From the redis documentation: "setting names to connections is a good way to debug connection leaks due to bugs in the application using Redis.". The problem is on subscribed connection the name must be set before any subscribe, although the developer can take care of it on the initial connection, the name is lost if ioredis reconnects automatically and resubscribe. A new option "connectionName" can be sued to set the name and leave it to ioredis to set the name consistently across reconnections.

### DIFF
--- a/lib/redis.js
+++ b/lib/redis.js
@@ -153,6 +153,7 @@ Redis.defaultOptions = {
     return Math.min(times * 2, 2000);
   },
   keepAlive: 0,
+  connectionName: null,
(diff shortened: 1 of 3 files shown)


luin:
Awesome! :rocket: Thanks for the pr!

luin:
Shipped in 1.10.0

PR #206, opened 2015-12-06
fix critical typo in handleError

shaharmor:
this is a critical fix as the timeout never gets nulled so it will not work after a single failover


luin:
Fixed in 1.12.2

PR #213, opened 2015-12-14
Made scan streams closeable

jeffjen:
I have a query interface that allows users to query based on key pattern. The user can terminate their request at arbitrary moment, thus I need to have the scan stream "closable", and the receiver knows to drain the remaining data from buffer before closing.

What should happen here is if you do
```
var stream = redis.scanStream({pattern: "*", count: 100});

req.once("close", function() {
    stream.close(); // begin connection draining
});

stream.on("data", function(keys) {
    if (stream.isClosed()) {
        return; // connection draining
    }
    // do work here, or pipe to another stream
});

stream.on("end", function() {
    // close down stream, or bye bye
});

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,8 @@ This file is a manually maintained list of changes for each release. Feel free t
 
 ## Master Branch (Unreleased)
 
+## v.1.13.1 - December 20, 2015
+
 * Fix `mset` transformer not supporting keyPrefix ([#217](https://github.com/luin/ioredis/issues/217)).
 
 ## v.1.13.0 - December 13, 2015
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by luin** [comment]
+1 for the `close` method. However, I doubt whether the `isClosed` is necessary here since it seems not to be a standard method of `Stream.Readable`. You may store the status of the stream on the client side:

```javascript
var streamIsClosed;

req.once("close", function() {
    stream.close(); // begin connection draining
    streamIsClosed = true;
});

stream.on("data", function(keys) {
    if (streamIsClosed) {
        return; // connection draining
    }
    // do work here, or pipe to another stream
});
```

jeffjen:
Oh yes, I will submit a new change without isClosed.  I was thinking whether it needed a way to indicate it is draining.  Not looking at node source code on Readable, I don't think it will end the stream outright when there are data in buffer, right?

PR #236, opened 2016-01-22
feat: revisit of .to(nodeGroup) command

AVVS:
Provides ability to send arbitrary commands too all nodes in a requested group or
just a handy helper to select nodes from such a group

Supports 'all', 'masters' and 'slaves' groups exposing `call` and `callBuffer` command
semantics similar to Commander

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -401,7 +401,7 @@ Cluster.prototype.executeClusterDownCommands = function () {
 
 Cluster.prototype.to = function (name) {
   var fnName = '_select' + name[0].toUpperCase() + name.slice(1);
-  var fn = typeof this[fnName];
+  var fn = this[fnName];
   if (typeof fn !== 'function') {
     // programmatic error, can't happen in prod, so throw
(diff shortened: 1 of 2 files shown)


lib/cluster.js
```diff
@@ -399,6 +399,59 @@ Cluster.prototype.executeClusterDownCommands = function () {
   }
 };
 
+Cluster.prototype.to = function (name) {
+  var fnName = '_select' + name[0].toUpperCase() + name.slice(1);
+  var fn = typeof this[fnName];
```
BridgeAR: The `typeof` is to much. Otherwise this function is always going to throw ;-)

luin:
Awesome! Fix the `typeof` part that @BridgeAR pointed out and the pr will be good to merge :laughing: . Serval test cases for this pr will be cool.

AVVS:
Fixed the typeof duplication. Tests cases would, indeed, be great, but I don't have time to write them for the next week or so. If some1 wants to pitch in - that would be awesome

luin:
I'll write the test cases for it. Merged :rocket:

PR #238, opened 2016-01-27
fix issue where options were not changed if passed freezed object

### LINKED ISSUE #233: Cluster isn't ready and enableOfflineQueue options is false, with defaults
Hey,

We are using `ioredis` for quite some time now and its great.

We noticed that sometimes (Can't reproduce really) we have logs of `Cluster isn't ready and enableOfflineQueue options is false`, even though we are using the default settings of `enableOfflineQueue` which is `true`.

Any ideas?

shaharmor:
fixes #233 

The problem was that the `_.defaults()` function was using the `options` argument that was passed to it as the base object, and when that object was being freezed by `Object.freeze()`, the defaults weren't being used.

### DIFF
--- a/lib/cluster.js
+++ b/lib/cluster.js
@@ -61,7 +61,7 @@ function Cluster(startupNodes, options) {
   this.slots = [];
 
   this.retryAttempts = 0;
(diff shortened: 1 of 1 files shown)


luin:
Nice catch :rocket:

PR #241, opened 2016-01-31
feat(offlineQueue): add option to limit the offline queue size

### LINKED ISSUE #76: Offline queue size limit
As ioredis supports offline queue, is it possible that the queue will grow to infinity if the client can not connect to redis server for long time, which will lead to huge memory growth ?

If so, I think it is nice to have a param to control the max size of the offline queue, and eviction policy, e.g. LRU

luin:
When the size of offline queue reaches the `maxOfflineQueueSize`, the overflow commands will be
flushed with an error.

Closes #76


dirkbonhomme:
Looks ok to me. Simple option and easy to understand.

BridgeAR:
I still recommend using bytes to determine if the queue is full or not. The memory is not exact but it's not difficult to have a approximation that is about right.

edevil:
This PR, along with https://github.com/luin/ioredis/pull/658, seem very important to me in order to prevent an unresponsive Redis server from bringing down an application. Is there a reason for them not having been merged?

edevil:
ping

Issue #225, opened 2016-01-07
Cluster emits 'ready' when not connected.


Salakar:
When using cluster, the `ready` event is emitted even if the connection is invalid or not connected.

The readme says that `ready` fires after `connect`. Is that correct?

Was getting this when my cluster was down. Also happens if I just put in a couple random hosts that don't exist - still get a ready event.

When I try running any commands I get `Connection is closed` errors.

luin:
`ready` fires right after fetching the slot information of the cluster by sending `CLUSTER SLOTS` to one of the nodes in the cluster, so that `ready` means there's at least one node is accessible and ioredis has already known where the other nodes are.

`Connection is closed` is because all nodes are down and `clusterRetryStrategy` is disabled.

However, it's strange to get `ready` event when you put in a couple random hosts that don't exist. Could you please enable the debug mode (`DEBUG=ioredis:* node yourapp.js`) and post the logs here?

Salakar:
When i tried with just one cluster node, which was in a `CLUSTER DOWN` state it still emits ready. But there's no connections. 

i.e:

```javascript
this.client = new Redis.Cluster([
        {host: '127.0.0.1', port: 6379}  // in a CLUSTER DOWN state
      ]);
```

I'm using default settings, so not touched `clusterRetryStrategy` yet.

luin:
That's correct. `CLUSTER SLOTS` command returns successfully even if the cluster in the `CLUSTER DOWN` state. In that case, when emits `ready`, ioredis has already connected to a node and fetch the information of the cluster successfully.

Salakar:
Should it not check that the cluster can receive commands?

The readme says `ready` is emitted when the server is ready to receive commands, doesn't `CLUSTER DOWN` count as not ready to receive commands (as in userland ones, as apposed to info commands ioredis uses to check state)?

Salakar:
```javascript
import Redis from 'ioredis';

const client = new Redis.Cluster([
  {host: '127.0.0.1', port: 7777}
]);

client.on('error', function (error) {
  console.dir(error);
});

client.on('ready', function () {
  console.log('****** READY ******'); // getting spammed

  client.get('ABCD', function (err) {
    // never calls back - stuck in offline queue i guess?
    console.error(err);
  });
});
```

It also spams ready loads when cluster is down.

![image](https://cloud.githubusercontent.com/assets/5347038/12197876/d2cf7dea-b602-11e5-9afd-f44970014882.png)

[...]

PR #246, opened 2016-02-04
ioredis 2.0

luin:
- [x] upgrade Bluebird -> 3.x
- [x] upgrade Lodash -> 4.x
- [x] use redis-parser module
- [x] use redis-commands module
- [x] [cluster] support `TRYAGAIN` error
- [x] [cluster] read-write splitting
- [x] [cluster] fix `ready` event #225 
- [x] multi support hgetall
- [x] add examples
- [x] `Cluster#setStartupNodes()` or something similar to update startup nodes dynamically
- [x] emit `authError` when redis requires a password #258

```javascript
// Read-write splitting:
var cluster = new Cluster([/* nodes */], {
  splitReads: 'masters' // 'slaves', 'all'
});

// Get slaves:
cluster.nodes('slave');

// Get masters:
cluster.nodes('master');

// Send command to all masters:
Promise.all(cluster.nodes('master').map(function (node) {
  return node.flushdb();
}));

// Get all:
cluster.nodes(); // or cluster.nodes('all');
```


lib/redis/parser.js
```diff
@@ -15,18 +17,19 @@ var debug = require('debug')('ioredis:reply');
 exports.initParser = function () {
   var self = this;
```
shaharmor: Maybe use _this for consistency?

shaharmor:
There is a memory leak somewhere here.
Running this vs latest stable clearly shows it.

I'll try to debug it and find where exactly

shaharmor:
Ok i think it only happens when using a string for the host instead of an IP

luin:
@shaharmor ioredis uses key `${ip}:${port}` to differentiate between the different connections. So using a  host string instead of an IP will create two connections for each node. That will definitely use more memory (twice maybe) but should not cause memory leak.

Seems we can resolve the host to the IP manually to solve the problem.

shaharmor:
@luin i know it shouldn't, but it is :smile: 
Not sure why exactly, but it causes the memory to increase indefinitely, until the server reaches its max memory and crashes.

shaharmor:
Its also disconnecting those servers (the ones using hostname instead of IP)

shaharmor:
I know for a fact that it isn't happening in older version because we are using hostnames all over the place

shaharmor:
Now i'm not sure its related... using 127.0.0.1 and still happens

luin:
All connections not included in the result of `CLUSTER SLOTS` will be disconnected. So it's normal that the connections using hostnames are lost since `CLUSTER SLOTS` only returns ips.

Could you give me a reproducible example?

shaharmor:
Try once with cluster & once without:
https://gist.github.com/shaharmor/7fbf9ef4cf606b06c214

luin:
I'll do some tests these days.

ramonsnir:
@shaharmor What if you use `dns.lookup` applicatively?

shaharmor:
@ramonsnir i though it was related to the DNS issue at first, but i managed to replicate it without it.

Even with 127.0.0.1, the RSS keeps rising forever (After about 1-2 minutes you are already at 300MB)
Without cluster, it stays steady at around 75MB

ramonsnir:
Weird. I have it stabilizing at 65MB after 1-2 minutes. Tested both with mainline ioredis 1.x and with this 2.x branch. With a 3-masters-10-slaves cluster (was testing something, so had it ready).

What version of Redis are you using? I'm running 3.0.7 locally.

ramonsnir:
Also, I'm oddly still running Node 0.10.40. Did you move forwards?

shaharmor:
Locally 3.0.6, in production 3.0.7. It happens on both.

shaharmor:
Yeah i'm on Node v4.3.0...

AVVS:
btw, any chance we get to use [Map](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Map) for handling nodes? This seems to be more suited towards it and is well supported in 4.x and 5.x, but this would mean 0.10.x will need to include polyfill for it

ramonsnir:
@luin Really strange. I installed with nvm Node 0.12.10, Node 4.3.0 and Node 5.6.0: all three are leaking. 0.10.40 isn't.

@AVVS +1 for Map.

AVVS:
Do you have memory dump for this? I mean if its leaking as fast as you are saying - there should be some clear indication in it

Or code to replicate it would be just as nice

ramonsnir:
Correction for 0.12: it stabilized, eventually, around 350MB.

@AVVS I'll upload a heapdump from Node 4, but Shahar gave a link above for code to reproduce: https://gist.github.com/shaharmor/7fbf9ef4cf606b06c214

luin:
How does `Map` benefit us despite of semantic?
Is there any chance V8 gc not getting involved. What about calling gc explicitly before printing memory usage (https://simonmcmanus.wordpress.com/2013/01/03/forcing-garbage-collection-with-node-js-and-v8/)?

ramonsnir:
Explicitly calling the gc seems to resolve this. Seems that someone the gc's behavior changed during Node's iojs phase of life...

shaharmor:
But it doesn't happen on stable version (From my testing), so its not related to node really..

ramonsnir:
http://expirebox.com/download/a9ad167c59af84903ad48b58b2cab0c3.html - dump from Node 4.6, without calling gc explicitly (although, calling heapdump **does** some gc before it saves, but not a full gc).

ramonsnir:
Looks like the node list of each slot is allocated in memory A LOT of times.

AVVS:
@luin the way objects are treated in node is basically the following:

1. object created, uses optimized mode
2. when property changes - it recreates underlaying structure of object
3. it does 1 and 2 until there were too many mutations on this object or until `delete` is called at least once, then it switches to hash mode and becomes many times slower

With map it doesn't try to do these things and is more suited towards dynamic allocation of keys. I would be mainly concerned about obj[key] vs Map.get(key) performance, because this is potentially a hot place in the code. Either way it's not likely to yield anything more than a few nanoseconds, but in the example posted above it should work faster.

@shaharmor @ramonsnir - in the posted example there are tons of strings being initialized and then they all must be disposed of, since it's utf-8 and it eats a lot of memory. so it's possible GC just doesn't kick in fast enough to dispose of all the garbage

shaharmor:
@AVVS so why doesn't it happen on non-cluster / stable cluster?

shaharmor:
I updated the gist: https://gist.github.com/shaharmor/7fbf9ef4cf606b06c214

Can you try now?

ramonsnir:
@shaharmor If all you did was reuse the "key" string, then it's still leaking on Node.js 4+.

shaharmor:
Yeah i though it will :smile: 

I still think its somewhere in the library it self because it doesn't happen in older version.

AVVS:
key is a `String` - not a reference like object, therefore, copied all the time anyway. Likely it's not the case anyway

luin:
The heapdump seems not telling much information at first sight. Will look into it this weekend.

luin:
I ran @shaharmor's gist on my computer and the memory usage report is very similar to v1.x. I didn't notice any sign of memory leak.
Here's my result: https://gist.github.com/luin/674a876791da9d35def3

shaharmor:
Which node version?
Did you run it against Redis Cluster? What configuration?

luin:
My node version is 4.2.1.

```
ioredis git:2.x ❯ redis-cli -p 6380                                                      ⏎ ✭
127.0.0.1:6380> cluster nodes
0a9570f08b951255ea1d753eb13552eaef34b1a5 127.0.0.1:6383 slave 1b781975c7bb4a5cdee0879587fe28ed04ca3a15 0 1455956006170 4 connected
311d5c617cd24d48719238071cca4a0aba62df69 127.0.0.1:6381 master - 0 1455956005125 2 connected 5461-10922
e25c7a0336d08750f13390929c986f55cefe3fc2 127.0.0.1:6384 slave 311d5c617cd24d48719238071cca4a0aba62df69 0 1455956008237 5 connected
8a82bad90c4c0de3e93bf6a5cbce31b635cd2d57 127.0.0.1:6385 slave 34d65132841ed3fda45d6fd79c7ba4f43292b741 0 1455956007206 6 connected
34d65132841ed3fda45d6fd79c7ba4f43292b741 127.0.0.1:6382 master - 0 1455956002014 3 connected 10923-16383
1b781975c7bb4a5cdee0879587fe28ed04ca3a15 127.0.0.1:6380 myself,master - 0 0 1 connected 0-5460
```

shaharmor:
Don't really know what to say.. I can reproduce it each time. @ramonsnir any suggestions?

luin:
@shaharmor Could you enable the debug mode `DEBUG=ioredis:* yourapp.js` to see whether the commands are added into the `offlineQueue` or `clusterDownQueue`?

ramonsnir:
```
  ioredis:redis queue command[0] -> set(quay,993) +0ms                                                                                                                                                                                                                  [534/2097]
  ioredis:redis queue command[0] -> set(quay,994) +0ms
  ioredis:redis queue command[0] -> set(quay,995) +0ms
  ioredis:redis queue command[0] -> set(quay,996) +0ms
  ioredis:redis queue command[0] -> set(quay,997) +0ms
  ioredis:redis queue command[0] -> set(quay,998) +0ms
  ioredis:redis queue command[0] -> set(quay,999) +0ms
  ioredis:redis status[172.17.0.3:7000]: connecting -> connect +2ms
  ioredis:redis write command[0] -> info() +0ms
  ioredis:redis status[172.17.0.3:7000]: connect -> ready +0ms
  ioredis:connection send 1000 commands in offline queue +0ms
  ioredis:redis write command[0] -> set(quay,0) +0ms
  ioredis:redis write command[0] -> set(quay,1) +0ms
  ioredis:redis write command[0] -> set(quay,2) +0ms
  ioredis:redis write command[0] -> set(quay,3) +0ms
  ioredis:redis write command[0] -> set(quay,4) +0ms
  ioredis:redis write command[0] -> set(quay,5) +0ms
  ioredis:redis write command[0] -> set(quay,6) +0ms
  ioredis:redis write command[0] -> set(quay,7) +0ms
  ioredis:redis write command[0] -> set(quay,8) +1ms
  ioredis:redis write command[0] -> set(quay,9) +0ms
  ioredis:redis write command[0] -> set(quay,10) +0ms
```

@luin can you try to run the following docker-compose? It leaks locally. If it doesn't link to you, then I'd want to ask what Linux Kernel version you have (I'm on 4.1.13).

https://gist.github.com/ramonsnir/bb7158931db43cb7185e

luin:
I saw the leaking when running the docker in an Ubuntu (kernel 4.4.0). Digging into the problem...

shaharmor:
@luin anything?

luin:
@shaharmor Still not have any substantive progress. If I call `global.gc()` manually every 500ms, the leaking would not be obvious (which is similar to the master branch). However, if not calling `global.gc()`, 2.x branch costs a lot more memory than master branch. Very strange.

shaharmor:
Maybe we can go commit by commit and see which one was the one causing the leak.
Its possible that its one of the libraries like bluebird or lodash too.

ramonsnir:
My bisect stopped on https://github.com/luin/ioredis/commit/23851692b2e0f8e7d39d42061ab586e6071ac175 @luin

luin:
@ramonsnir @shaharmor @AVVS Finally found the "leaking": https://github.com/luin/ioredis/commit/410af511b7796ee3f2c656c28a8a1842c6b0f23d. :beers:

AVVS:
Great catch. Makes sence that the function is referenced and when it's never called - it appears to still be in the scope.

shaharmor:
@luin good! i just found it on the flight back home too :)

shaharmor:
So now that the bug is fixed, any plans to finish this PR soon?

luin:
@shaharmor Just released v2.0.0-alpha2. I'll finish the other todos asap.

ScheerMT:
Two questions!

1. Curious on how that was causing a leak...Was it because it was in an enclosing scope and the function inside of it was referencing it?
2. There was talk about this project merging with another redis package at one point i believe... Did that fall through and you are proceeding with a v2 release seperate of that?


Awesome work!

Salakar:
Been testing `v2.0.0-alpha2` clustered for a couple days now, not had any issues so far, memory issue seems to have been resolved.

The new scaleReads options seems to be working fine also, much more flexible than the old `readOnly` option.

luin:
@ScheerMT 
> Curious on how that was causing a leak...Was it because it was in an enclosing scope and the function inside of it was referencing it?

Not sure about that. It seems to be a problem that `_.partial` creates many closures that never used.

> There was talk about this project merging with another redis package at one point i believe... Did that fall through and you are proceeding with a v2 release seperate of that?

It's not easy to merge node_redis and ioredis into a single project. However we are trying to create some common modules which can be used by both them. For example, redis-parser and redis-commands are the two modules that integrated into ioredis in this pull request.

BridgeAR:
@ScheerMT one reason why this can't be done right away is also that the clients support different things at the moment and both need to have round about the same features to make such a thing possible.

So reusing the same code is a step into that direction.
Two of the goals in this milestone are actually to use those abstractions.

luin:
All todos are finished. I merged `2.x` branch into `master` and the old master branch are renamed to `1.x`. v2.0.0-alpha3 is just released.

Issue #250, opened 2016-02-09
ioredis 2: custom filter predicates for "scaleReads" 


ramonsnir:
Great job, btw, both on ioredis 1 and ioredis 2!

It would be really great if it would be possible to give the "scaleReads" option a custom function that filters which nodes are applicable to be read from, in case not all nodes are born equal (e.g. some are closer to the application than others, network-wise).

Thinking about it, even a bit more:

1. Giving 2 functions: one for readonly commands and one for readwrite commands.
2. Supporting not only a yes/no boolean, but a 0-9 priority (so ioredis would prefer those with 9 over those with 8, etc. etc., but never those with 0). This is for more complicated Redis Cluster topologies.

If you agree that this is at least an acceptable feature, I could open a PR for you into the 2.x branch.

P.S.: shouldn't there be a way to mark eval/evalsha as readonly - so reading lua scripts won't be redirected to a master?

luin:
Sounds interesting. Could you give me an example about the custom function case?

P.S. It seems that redis always consider `EVAL` and `EVALSHA` as writes, so the slaves will not handle them (https://github.com/antirez/redis/blob/unstable/src/cluster.c#L5178-L5184).

ramonsnir:
The one I'm personally having at the moment is that my cluster spans across multiple network subnets, and the latency between them is relatively large. So my function would take the node's IP, check if it's in the "local subnet IP range", and return a lower priority if it isn't. This would lower significantly the read times, as otherwise reads would go to a random slave which may or may not be nearby (i.e. 50ms latency instead of 1ms latency).

P.S.: That is too bad :disappointed: Yet another regression from classic Redis.

luin:
That makes sense. What about a function with two parameters `slaves` and `command`? `slaves` is an array of the slaves serving the slot. Return an index of the array to send the command to the specified slave, or return a `false` value to send to master.

```javascript
new Redis.Cluster([/* nodes */], {
  scaleReads: function (slaves, command) {
    if (command.name === 'get') {
      return 0;
    }
  }
});
```

ramonsnir:
That could work, although I thought that you'd prefer to keep as much logic as possible inside the library, just letting the custom function filter out low-prioritized nodes.

Your suggestion 100% solves the problem.

luin:
Closing this issue in favour of #252.

PR #252, opened 2016-02-10
feat(cluster): add the option for a custom node selector in scaleReads

ramonsnir:
#250

### DIFF
--- a/lib/cluster/index.js
+++ b/lib/cluster/index.js
@@ -40,9 +40,10 @@ function Cluster(startupNodes, options) {
   this.options = _.defaults(this.options, options, Cluster.defaultOptions);
 
   // validate options
-  if (['all', 'master', 'slave'].indexOf(this.options.scaleReads) === -1) {
+  if (typeof this.options.scaleReads !== 'function' &&
+      ['all', 'master', 'slave'].indexOf(this.options.scaleReads) === -1) {
     throw new Error('Invalid option scaleReads "' + this.options.scaleReads +
-      '". Expected "all", "master" or "slave"');
+      '". Expected "all", "master", "slave" or a custom function');
   }
 
   if (!Array.isArray(startupNodes) || startupNodes.length === 0) {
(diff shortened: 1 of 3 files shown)


luin:
Awesome! Thank you for the pull request!

It seems better to change `function(slaves, command) {}` to `function(nodes, command) {}` where the first  element of `nodes` is the master node and the reset are slaves so that it's possible to send the queries to masters.

lib/cluster/index.js
```diff
@@ -453,15 +454,28 @@ Cluster.prototype.sendCommand = function (command, stream, node) {
...
-              key = utils.sample(nodeKeys, 1);
+            if (typeof to === 'function') {
+              var nodes =
+                  nodeKeys
+                  .slice(1)
+                  .map(function(key) {
+                    return _this.connectionPool.nodes.all[key];
+                  });
+              redis = to(nodes, command);
+              if (redis.constructor === Array) {
```
luin: `if (Array.isArray(redis))` may be better IMO, and we should check whether `redis` is null/undefined and default to the master node.

ramonsnir: Yes, you are correct. I'm sadly used too much to client-side JS, and Array.isArray is unavailable in some MSIE versions...

luin:
LGTM. Any thoughts about changing `function(slaves, command) {}` to `function(nodes, command) {}` so that we don't need to call `slice(1)` everytime.

ramonsnir:
It bothered me too... I'll do that, and make sure the README states it clearly.

luin:
Cool! Merged. :+1:

ramonsnir:
Thanks!

lib/cluster/index.js
BridgeAR: As a hint: there's a easy [polyfill](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Polyfill) for this and Microsoft itself does [not maintain IE 8](https://support.microsoft.com/en-us/lifecycle#gp/Microsoft-Internet-Explorer) or older, not even with security fixes on any regular Windows version. Most libraries drop support for it or have already dropped the support and the faster all developers switch to not supporting it anymore, the faster no one is going to visit any website with it anymore.

PR #264, opened 2016-03-07
feat(cluster): redirect on TRYAGAIN error

luin:


test/functional/cluster.js
```diff
@@ -453,6 +453,37 @@ describe('cluster', function () {
     });
   });
 
+  describe('TRYAGAIN', function () {
```
shaharmor: I think this test is not well defined, as its possible that the 2nd `foo` will come from the line 482 and not from the retry itself.

this will be better tested like this:
```javascript
describe('TRYAGAIN', function () {
    it('should retry the command', function (done) {
      var times = 0;
      var slotTable = [
        [0, 16383, ['127.0.0.1', 30001]]
      ];
      var server = new MockServer(30001, function (argv) {
        if (argv[0] === 'cluster' && argv[1] === 'slots') {
          return slotTable;
        }
        if (argv[0] === 'get' && argv[1] === 'foo') {
          if (times++ === 1) {
            return;
          } else {
            return new Error('TRYAGAIN Multiple keys request during rehashing of slot');
          }
        }
        if (argv[0] === 'get' && argv[1] === 'boo' && times === 2) {
          process.nextTick(function () {
            cluster.disconnect();
            disconnect([server], done);
          });
        }
      });

      var cluster = new Redis.Cluster([
        { host: '127.0.0.1', port: '30001' }
      ], { retryDelayOnTryAgain: 1 });
      cluster.get('foo', 1, function () {
        cluster.get('boo');
      });
    });
  });
```

luin: Nice catch. It seems to fix the test by just changing

```javascript
cluster.get('foo', function () {
  cluster.get('foo');
});
```

to

```javascript
cluster.get('foo');
```

Issue #277, opened 2016-04-05
[question] Disconnect from cluster


devnetf:
Hello,
         I have the following test code: 
        `cluster.get('status').then(function(res){
     console.log(res);
     console.log(cluster.status);
     cluster.disconnect(); 
});

setInterval(function(){
    console.log(cluster.status);
}, 1000);`

After cluster.disconnect(); is called, cluster.status still give me "ready". How should I disconnect from the cluster? 

Thanks!

devnetf:
Anyone help please?

luin:
Sorry I didn't notice this issue. Could you enable the debug mode (`DEBUG=ioredis:* node yourapp.js`) to see the logs?

devnetf:
![ioredis_log](https://cloud.githubusercontent.com/assets/4071282/14357643/8447c0e6-fcb7-11e5-8902-8fac8f49aab6.png)

Here is the debug log. I am using the newest version: "2.0.0-rc1"

Thanks!

luin:
Thank you for the details. It seems that hosts `uco-dev***` are resolved to internal ips `10.149.*.*`. Did you create the cluster with the hosts rather than the ips? What about using IP address instead?

devnetf:
Thanks for the quick reply. I am not sure about the cluster setup because it is our system admins who set it up. However, when I change the version to "1.15.1". The exact same code works. Could it be a bug in v2.0.0?

luin:
Yes, it might be a bug if the code works with v1.15.1. I'll look into it this weekend.

luin:
@devnetf This issue should be fixed in v2.0.0-rc2. Feel free to reopen it if the issue still exists. Thank you for the feedback!

PR #281, opened 2016-04-09
fix(CLUSTER): fix cluster not disconnected when called disconnect method

luin:
Before this change, when disconnecting from a cluster, the nodes will be disconnected one by one. However, when the subscriber is disconnected, a new subscriber will be connected to in `-node` event, leading the cluster keeping connected.

Related issue: #277

### DIFF
--- a/lib/cluster/index.js
+++ b/lib/cluster/index.js
@@ -57,7 +57,7 @@ function Cluster(startupNodes, options) {
 
   var _this = this;
   this.connectionPool.on('-node', function (redis) {
-    if (_this.subscriber === redis) {
+    if (_this.status === 'ready' && _this.subscriber === redis) {
       _this.selectSubscriber();
     }
     _this.emit('-node', redis);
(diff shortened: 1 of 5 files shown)


lib/cluster/index.js
```diff
@@ -57,7 +57,7 @@ function Cluster(startupNodes, options) {
 
   var _this = this;
   this.connectionPool.on('-node', function (redis) {
-    if (_this.subscriber === redis) {
+    if (_this.status === 'ready' && _this.subscriber === redis) {
```
shaharmor: We should export possible statuses to an enum to avoid having typos.
something like:

`_this.status = Redis.Statuses.READY`

luin: Enum may not be able to avoid typos IMO. For example, I may wrongly write `Redis.READY` as `Redis.REEDY`, and the latter will be an `undefined`, no exception would be thrown.

shaharmor: Yeah, but your IDE will probably detect it.

shaharmor: Not critical though, just saying :)

PR #293, opened 2016-05-03
feat: add dropBufferSupport option to improve the performance

luin:
https://github.com/luin/ioredis/wiki/Improve-Performance


test/functional/drop_buffer_support.js
```diff
@@ -0,0 +1,76 @@
...
+    });
+    redis.geteval(function (err, res) {
+      expect(err).to.eql(null);
+      expect(res).to.eql('string');
+      redis.disconnect();
+      done();
+    });
+  });
+
+  it('should return strings correctly', function (done) {
```
shaharmor: duplicate from two tests above, did you mean to use `dropBufferSupport: false`

luin: This test makes sure non-binary methods (`get`, `set`) works when `dropBufferSupport` is enabled.

shaharmor: Yeah but there is another test that is exactly the same down below (Look at the last test)

luin: Aha...you're right.
shame-LfFC2erYa5qN2

PR #297, opened 2016-05-09
Use given 'dropBufferSupport' for sentinel connections

TeeAaTeeUu:
Basically same as https://github.com/luin/ioredis/pull/155 , but this time passing the `dropBufferSupport` option also through the sentinel-connector.

Would it be better to just pass the whole `this.options`, and only override or exclude properties like `host`, `port` and `enableReadyCheck`?

### DIFF
--- a/lib/connectors/sentinel_connector.js
+++ b/lib/connectors/sentinel_connector.js
@@ -124,7 +124,8 @@ SentinelConnector.prototype.resolve = function (endpoint, callback) {
     host: endpoint.host,
     retryStrategy: null,
     enableReadyCheck: false,
-    connectTimeout: this.options.connectTimeout
+    connectTimeout: this.options.connectTimeout,
+    dropBufferSupport: true
   });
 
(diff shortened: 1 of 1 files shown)


luin:
Nice catch. However, maybe we could just set `dropBufferSupport` to true since we don't need any binary command.

TeeAaTeeUu:
So defaulting it to be true even for the non-sentinel connections, or removing them totally? I don't know about that, but for our use case they are useless especially with the https://github.com/redis/hiredis-node

luin:
Just change

```javascript
connectTimeout: this.options.connectTimeout,
dropBufferSupport: this.options.dropBufferSupport
```

to

```javascript
connectTimeout: this.options.connectTimeout,
dropBufferSupport: true
```

will be OK since the client created in sentinel_connector.js is used temporarily and will be destroyed when sentinels are connected.

TeeAaTeeUu:
Thanks for the guidance @luin!

PR #299, opened 2016-05-10
Add example of changing Redis server config

spiffytech:
Existing docs have no examples of CONFIG SET, and no mention of multi-word commands.


luin:
LGTM. Could you please change the double quotes to single quotes?

spiffytech:
Done.

PR #308, opened 2016-05-19
Include missing data in monitor events

9point6:
The monitor events generated in ioredis don't currently include all of the data that redis provides (The redis database and source are currently missing).

This adds the extra parameters to the end of the emitted event.

~~It seems like my text editor has cleared whitespace from the end of the lines of your docs too, but I can remove those changes if it's preferred to keep the pull request simpler.~~

### DIFF
--- a/API.md
+++ b/API.md
@@ -23,8 +23,8 @@
 <a name="Redis"></a>
 
 ## Redis ⇐ <code>[EventEmitter](http://nodejs.org/api/events.html#events_class_events_eventemitter)</code>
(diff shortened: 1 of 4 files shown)


luin:
Nice work! Could you please revert the changes of API.md since it's generated by scripts `npm run generate-docs `.

9point6:
Ah right! no worries, I'll do that now.

9point6:
@luin done!

luin:
🚀

PR #309, opened 2016-05-20
Update sentinels after getting master.

### LINKED ISSUE #304: Get list of sentinels from sentinel
After the client connects to a sentinel, it should get the list of all sentinels to add to the list passed in through options.

ddunkin:
Based on http://redis.io/topics/sentinel-clients#sentinels-list-automatic-refresh. Closes #304.


luin:
Just skipped through the code and it LGTM. I'll think about it carefully when I have time before merging this though.

PR #323, opened 2016-06-09
perf(cluster): improve the performance of calculating slots

luin:
By using the https://www.npmjs.com/package/cluster-key-slot package.


shaharmor:
LGTM

Salakar:
Just one small change that needs doing, the pipeline can use `var calculateSlotMulti = require('cluster-key-slot').generateMulti;` to calculate the slots for the array of keys there, then instead of checking for 'undefined' you just check if slot < 0 - then throw the 'keys must be of same slot' error. 

The multi on the calc repo is a little bit more performant. https://github.com/Salakar/cluster-key-slot#usage

Salakar:
Hmm build test failures seem a little random and unrelated?

luin:
@Salakar Just fixed it. If one of the test failed, the connection to the server will not be disconnected, so other tests will also fail.

Salakar:
Looks good, LGTM =]

PR #324, opened 2016-06-09
refactor: remove unused Command#getKeyPart

luin:
`getKeyPart` methods is not used anymore as the package redis-commands is used. Remove it won't break BC since it's not a public API (not listed on https://github.com/luin/ioredis/blob/master/API.md).


shaharmor:
LGTM

PR #339, opened 2016-06-26
feat(cluster): add Cluster#quit() to quit cluster gracefully. Close #315

### LINKED ISSUE #315: Disconnecting from Cluster while waiting for all pending commands
Hi,

Is it possible today to disconnect (or quit) a Redis Cluster while waiting for all pending commands to finish?

I know that i can do `redis.quit()` on a single instance, but what about Cluster?
Should i just do `cluster.nodes()` and run `redis.quit()` on each?

I tried that but the cluster keeps reconnecting. Should it somehow detect that we called `.quit()` and not try to reconnect?

luin:


lib/cluster/index.js
```diff
@@ -209,6 +209,29 @@ Cluster.prototype.disconnect = function (reconnect) {
...
+  this.manuallyClosing = true;
+
+  if (this.reconnectTimeout) {
+    clearTimeout(this.reconnectTimeout);
+    this.reconnectTimeout = null;
+  }
+  return Promise.all(this.nodes().map(function (node) {
+    return node.quit();
+  })).then(function (res) {
+    return res.length;
```
luin: Should we return the number of the node been disconnected or just `"OK"` to keep in line with the result of native quit.

shaharmor: I think just `"OK"` is good

PR #341, opened 2016-06-28
Ensure node exists before being redirected via an ASK

jpallen:
If a master node in the Redis cluster does not have any slots, it is not returned by the `cluster slots` command, and so does not have a client in the connectionPool. If slots are migrated to this new master node, the old master node returns an `ASK` error with the new node details. At the moment ioredis looks up the node in the connectionPool but it doesn't exist, so the following error is returned:

```
TypeError: Cannot call method 'asking' of undefined
  at tryConnection (/var/www/document-updater/releases/5543/node_modules/ioredis/lib/cluster/index.js:464:19)
  at Object._this.handleError.ask (/var/www/document-updater/releases/5543/node_modules/ioredis/lib/cluster/index.js:404:11)
  at Cluster.handleError (/var/www/document-updater/releases/5543/node_modules/ioredis/lib/cluster/index.js:503:52)
  at Command.command.reject (/var/www/document-updater/releases/5543/node_modules/ioredis/lib/cluster/index.js:389:13)
  at Redis.exports.returnError (/var/www/document-updater/releases/5543/node_modules/ioredis/lib/redis/parser.js:75:18)
```

It looks like a similar issues with a `MOVED` reply was fixed in https://github.com/luin/ioredis/commit/0dcb768e2e418b9dd3fa7feaf01ad85cb01521ef. This patch applies the same fix to the `ASK` error.

### DIFF
--- a/lib/cluster/index.js
+++ b/lib/cluster/index.js
@@ -401,6 +401,8 @@ Cluster.prototype.sendCommand = function (command, stream, node) {
         },
         ask: function (slot, key) {
           debug('command %s is required to ask %s:%s', command.name, key);
+          var splitKey = key.split(':');
+          _this.connectionPool.findOrCreate({ host: splitKey[0], port: Number(splitKey[1]) });
           tryConnection(false, key);
         },
         tryagain: partialTry,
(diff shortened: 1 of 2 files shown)


luin:
Released in v2.2.0. Thank you for the patch! 🍻

jpallen:
Wow, that was super quick! I see you've already done another release too, thanks so much :)

PR #345, opened 2016-07-08
Use redis.get only in callback

migg24:
This is a toxic example that will only work on your local machine. Add network latency and clustering and you will not find ```foo``` all the time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,10 @@ $ npm install ioredis
 var Redis = require('ioredis');
 var redis = new Redis();
 
-redis.set('foo', 'bar');
-redis.get('foo', function (err, result) {
-  console.log(result);
+redis.set('foo', 'bar', function() {
+  redis.get('foo', function (err, result) {
+    console.log(result);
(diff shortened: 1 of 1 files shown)


luin:
Network latency will not make this example fails since the commands are processed one by one.

migg24:
Are they executed in a blocking way?

AVVS:
@migg24 they aren't, but they are sent through the same tcp socket sequentially

migg24:
Ok then maybe latency is not a problem and the example is fine. Then the problem is our clustering of the app. If you make a request to a route that executes the ```set``` and instantly another request to another route (on a different server) that executes the ```get``` then the set may not be finished and you get an "id not found" error as we were sporadically getting.

luin:
@migg24 yes, that may happen since the two commands are sent with two tcp connection so you're not sure which one is handled by Redis first.

Closing this pr.

migg24:
THX for enlighting :)

PR #352, opened 2016-07-21
Cluster lazy connect

### LINKED ISSUE #350: Add config option to not connect in the constructor
Hi,

Can we add a config option to the constructor so that it will **not** try to connect until manually calling `.connect()`?

shaharmor:
fixes #350 
@luin not sure if i missed something


luin:
Nice work! `Cluster#disconnect()` and `Cluster#quit()` require special handling since when not connected to the cluster, connectionPool will not emit `drain` event when calling `reset([])`.

shaharmor:
thnx @luin , didn't have any time for this

PR #354, opened 2016-08-03
Reject with general error in Redis#connect

Gerhut:
Now it rejects with an `undefined`, which will cause warnings.


luin:
Nice catch! Could you change the error message to `utils. utils.CONNECTION_CLOSED_ERROR_MSG`?

Gerhut:
Changed

PR #370, opened 2016-09-24
Sentinel preferredSlaves option

### LINKED ISSUE #38: rfc - a preferred slave list in a sentinel setup
We want to run a sentinel setup: 
- each webserver has it's own redis slave (will NOT become master - using `slave-priority = 0`)
- a set of sentinels running to protect us from partioning and handle the quorum
- a master and a few redis slaves for failover of master

In the current implementation , the slaves can be selected using the `role slave` but it picks a rather arbitrary slave. We would propose to have a way to indicate a list of preferred slaves for a client.

This is useful is in multi-AZ aws setups, where we want to keep the latency to app<->slave low .
A new syntax could look like: 

```
var redis = new Redis({
  preferred_slaves: [ { prio: 10 , host: 'localhost', port: 6381 }]
  name: 'mymaster'
});
````

doublesharp:
Option to specify a preferred slave (or slaves by priority) when connecting to a Sentinel group.

Fixes #38 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -608,10 +608,46 @@ The arguments passed to the constructor are different from the ones you use to c
 
 * `name` identifies a group of Redis instances composed of a master and one or more slaves (`mymaster` in the example);
 * `sentinels` are a list of sentinels to connect to. The list does not need to enumerate all your sentinel instances, but a few so that if one is down the client will try the next one.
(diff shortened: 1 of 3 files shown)


luin:
Released with 2.4.0. Thank you for the pull request (and the documentation for the changes) :-)

Issue #405, opened 2016-12-02
Perf: after creating connection enable tcp_nodelay


AVVS:
As in put stream.setNoDelay() on the socket
Will decrease latency in most cases

luin:
Thank you for addressing that! I've added you as the contributor, please review & merge the pull request above.

AVVS:
Thanks, will do

AVVS:
Closed by #406

AVVS:
We also might want to add that if somebody does bulk export/import and its important to increase _throughput_, whereas latency doesn't matter as much - `noDelay` should be set to `false`, because while this decreases latency, it reduces maximum throughput

AVVS:
@luin can you release the latest version?

luin:
v2.4.1 released. 😆

PR #406, opened 2016-12-02
perf: by default calling setNoDelay on the stream

luin:
#405

### DIFF
--- a/API.md
+++ b/API.md
@@ -56,6 +56,7 @@ Creates a Redis instance
 | [options.family] | <code>string</code> | <code>4</code> | Version of IP stack. Defaults to 4. |
 | [options.path] | <code>string</code> | <code>null</code> | Local domain socket path. If set the `port`, `host` and `family` will be ignored. |
 | [options.keepAlive] | <code>number</code> | <code>0</code> | TCP KeepAlive on the socket with a X ms delay before start. Set to a non-number value to disable keepAlive. |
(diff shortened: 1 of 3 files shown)


AVVS:
@luin dont have write access, but probably would do the following to ensure tls support and make sure it doesn't crash here. Afaik there were some changes to TLS sockets, but this actually could make it implementation agnostic for the future, as in unix pipes might have different interface, etc:

P.S. don't have write access to push it to the branch, nor to merge when it all passes tests

```js
      if (_this.options.noDelay) {
        if ('setNoDelay' in stream) {
          debug('disabling Naggle on raw tcp socket');
          stream.setNoDelay(true);
        }

        // NOTE: can also check based on _this.options.tls,
        // but this seems good enough
        if ('socket' in stream && 'setNoDelay' in stream.socket) {
          debug('disabling Naggle on tls socket');
          stream.socket.setNoDelay(true);
        }
      }
```

luin:
I've sent a invite link to you so that you should be able to write to this repo. You can check out https://github.com/luin/ioredis/invitations if you don't see it.

It seems `tls.TLSSocket` is the subclass of `net.Socket`, so I think it should have the `setNoDelay` method (ref: https://github.com/nodejs/node-v0.x-archive/issues/6194).

AVVS:
@luin you seem to be right, at least for the latest versions, methods are proxied:
https://github.com/nodejs/node/blob/master/lib/_tls_wrap.js#L290-L305

In that case this should be good to go

PR #410, opened 2016-12-14
Quit immediately when reconnecting

ruimarinho:
I'm not sure if this is the right approach but the idea is to be able to quit immediately when `redis.quit()` is called and the client is in the reconnecting state. In this case, `quit()` should not have to wait for the connection to be ready.

What do you think?

### DIFF
--- a/lib/redis.js
+++ b/lib/redis.js
@@ -579,6 +579,12 @@ Redis.prototype.sendCommand = function (command, stream) {
     return command.promise;
   }
 
+  if (!writable && command.name === 'quit' && this.offlineQueue.length === 0) {
+    this.disconnect();
+    command.resolve(new Buffer('OK'));
+    return command.promise;
+  }
+
   if (writable) {
     debug('write command[%d] -> %s(%s)', this.condition.select, command.name, command.args);
     (stream || this.stream).write(command.toWritable());
--- a/test/functional/connection.js
+++ b/test/functional/connection.js
@@ -146,6 +146,38 @@ describe('connection', function () {
         }
       });
     });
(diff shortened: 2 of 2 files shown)


ruimarinho:
@luin do you have time for some feedback here?

luin:
Sorry for the late response. There's a problem that the `QUIT` command returns an `OK` so the `command.resolve()` should reflect it.

ruimarinho:
No problem, I appreciate your time! Is that a simple `OK` or `[null, OK]`? From the Redis specification it looks like it's a Simple String Response. I've updated the PR to reply with an `OK` for now.

luin:
I think `command.resolve('OK')` should be `command.resolve(new Buffer('OK'))` to handle the case that users call `redis.quitBuffer()`.

We'd better also include a test for it:

```
redis.quitBuffer().then(function (result) {
    expect(result).to.be.instanceof(Buffer);
    expect(result.toString()).to.eql('OK');
})
```

ruimarinho:
SGTM! Do you think this should be a separate test?

luin:
Yes I think it should be a separate test since `quit` & `quitBuffer` are two different commands.

ruimarinho:
Let me know if you like the style of the new test.

luin:
LGTM! 👍

ruimarinho:
@luin would you be up to a patch/minor release which includes this PR? Not urgent, only if you have spare time and agree with it. Thanks!

luin:
Sure! Just released v2.5.0. Thank you for the pull request.

Issue #387, opened 2016-10-25
Read from slaves with replication


Alvaroctal:
Hi, 

I'm planning to use ioredis with redis-connect for a distributed express application with consistent sessions.

My doubt, after reading the docs, I've assumed the following:
- Clustering is not needed for session storage, the data won't be enough big.
- The majority of the operations will be reads in order to transform cookies into a req.session.

having the following nodes: 
- Node `A` master (with a known ip).
- Node `B` slave (replica) of `A` (localhost).
- Node `C` slave (replica) of `A` (other node, unknown ip).

I can not perform write operations to `B`, even if I set `slave-read-only` to `no`, the updates won't propagate, I need to write to `A`, so...

``` node
var Redis = require('ioredis');
var redis = new Redis('a-ip', port);

redis.set('foo', 'bar');
redis.get('foo', function (err, result) {
  console.log(result);
});
```

The ideal behavior would be, I am node `B`, I open a session, so I go `A` and set it, then the client comes in again, and I have to identify the cookie, so I read from `B` (localhost). Is this possible?

luin:
What you want is read-write splitting, which is already supported by Redis Cluster (`new Redis.Cluster()`). However, given your setup, I think you even don't need it since the dataset is relatively small (that the whole dataset can be held by one node) and the performance of Redis is very good when all your commands are simply `GET` & `SET`.

Nepoxx:
> which is already supported by Redis Cluster

Isn't a Redis cluster only for sharded data?

I too, have this use case of non-sharded data that's replicated where I want writes to the master and reads to the closest instance. I'm not quite sure how to do that (mongodb does this natively and that's what I'm looking into doing).

luin:
Read-write splitting is only implemented in Cluster mode currently. The support of it for standalone Redis or Sentinel is not here yet but pr is welcome.

It shouldn't be hard to do the same thing for Sentinel. You can refer to https://github.com/luin/ioredis/blob/master/lib/cluster/index.js#L427-L433 for how to detecting whether a command is readonly (that can be sent to slaves) or not. Then override `Redis#sendCommand` or just write a wrapper for it to send every command to different Redis instance.

Somebi:
There is no pipeline implemented in cluster setup in many redis clients. We decided to use sentinels and replications with sharding and scaling reads with this setup would be really awesome.

nodo:
[...]

PR #438, opened 2017-03-14
Add support for non-clustered replication without Sentinel

mvaker:
Hi!

I have made a few modifications to Cluster class to enable the support for replication cluster (which is intended to solve: https://github.com/luin/ioredis/issues/387).
I tried it in a project which I have a Redis cluster setup which doesn't use sharding but replication (master-slave full replicas) and it seems to work properly.

I would very much appreciate feedback about this feature and my resolution and I hope this branch gets merged some day so that we can all benefit from it and gets maintained.

### DIFF
--- a/lib/cluster/index.js
+++ b/lib/cluster/index.js
@@ -439,7 +439,10 @@ Cluster.prototype.sendCommand = function (command, stream, node) {
     }
   }
 
-  var targetSlot = node ? node.slot : command.getSlot();
+  var targetSlot = false;
+  if (!this.options.replication) {
+    targetSlot = node ? node.slot : command.getSlot();
+  }
   var ttl = {};
   var _this = this;
   if (!node && !command.__is_reject_overwritten) {
--- a/lib/cluster/index.js
+++ b/lib/cluster/index.js
@@ -168,12 +168,16 @@ Cluster.prototype.connect = function () {
     this.once('close', closeListener);
     this.once('close', this._handleCloseEvent.bind(this));
 
-    this.refreshSlotsCache(function (err) {
-      if (err && err.message === 'Failed to refresh slots cache.') {
-        Redis.prototype.silentEmit.call(this, 'error', err);
-        this.connectionPool.reset([]);
-      }
-    }.bind(this));
+    if (!this.options.replication) {
+      this.refreshSlotsCache(function (err) {
+        if (err && err.message === 'Failed to refresh slots cache.') {
+          Redis.prototype.silentEmit.call(this, 'error', err);
+          this.connectionPool.reset([]);
+        }
+      }.bind(this));
+    } else {
+      this.emit('refresh');
+    }
     this.selectSubscriber();
   }.bind(this));
 };
--- a/lib/cluster/index.js
+++ b/lib/cluster/index.js
@@ -490,7 +490,7 @@ Cluster.prototype.sendCommand = function (command, stream, node) {
       return;
     }
     var redis;
-    if (_this.status === 'ready' || (command.name === 'cluster')) {
+    if (_this.status === 'ready' || (command.name === 'cluster') || _this.options.replication) {
       if (node && node.redis) {
         redis = node.redis;
       } else if (Command.checkFlag('ENTER_SUBSCRIBER_MODE', command.name) ||
--- a/lib/cluster/index.js
+++ b/lib/cluster/index.js
@@ -624,31 +624,41 @@ Cluster.prototype.getInfoFromNode = function (redis, callback) {
  * @private
  */
 Cluster.prototype._readyCheck = function (callback) {
-  this.cluster('info', function (err, res) {
-    if (err) {
-      return callback(err);
-    }
-    if (typeof res !== 'string') {
-      return callback();
-    }
-
-    var state;
-    var lines = res.split('\r\n');
-    for (var i = 0; i < lines.length; ++i) {
-      var parts = lines[i].split(':');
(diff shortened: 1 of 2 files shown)


luin:
Thank you for the pull request! Could you give us more details of which problem this pr is intended to solve? I revisited #387 but read-write splitting is already supported in the Cluster mode.

mvaker:
The problem is with read-write splitting in Cluster mode with replication (not with sharded clusters, nor sentinel). When in cluster mode, it always tries to search for a slot to which a command has to be sent, and if the cluster is a replication one (not sharding), it fails because all slots are in all nodes.

AVVS:
> The problem is with read-write splitting in Cluster mode with replication (not with sharded clusters, nor sentinel). When in cluster mode, it always tries to search for a slot to which a command has to be sent, and if the cluster is a replication one (not sharding), it fails because all slots are in all nodes.

Its not a cluster - its a replicated single machine setup. Master - Slave, which should use sentinel for monitoring and automatic failover. It seems there is confusion over a cluster and sentinel.

@mvaker what you are referring to is a sentinel based setup (should be)

mvaker:
Hello! Thanks for your early comments.
It is not a Sentinel based setup, we are using this setup (cluster mode dissbled): 
http://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/Replication.Redis-RedisCluster.html

http://docs.aws.amazon.com/AmazonElastiCache/latest/UserGuide/AutoFailover.html

It does pretty much the same as sentinel but amazon provides it all itself (failover, monitor, etc). So it's a redis cluster without sharding (or with all shards in the same node), but since it's elasticache it's not sentinel based.
I have not tried the sentinel client here, but I understand it won't work either because no sentinel commands are available.
I think that the best approach is to use the cluster mode to keep track of existing nodes and connection pool, while sending reads to random reader (slave) instances.
Thank you!

mvaker:
BTW my resolution in this PR is a little bit "dirty" and I know this can probably be done better (maybe a separate class or client..). It was just a quick approach to see if it worked and hear what are your thoughts about it.

mvaker:
Hi, any thoughts on this? Have you checked my last comments? I would appreciate very much any comments. Thank you.

christoph-puppe:
Would love to see that. Don't need sentinel and no cluster. Just read-only slaves for 90% of the load but need to get the 10% writes to the master @luin 

My usecase: frontend nodes are too far away for cluster, so latency is high. Solution: have a slave per frontend-instance and speed up the whole thing for reads.

fenichelar:
This is a much needed feature! The AWS non-clustered approach is much better for write heavy workflows than clustering because AWS only supports on-the-fly resizing for non-clustered Redis.

fenichelar:
@mvaker You should probably rename this PR to "Add AWS non-clustered replication support" for clarity. This is not clustering.

mvaker:
hi @luin, I would like to discuss again this PR, so clearly this is a missing feature in this library and other people have proved it being useful for them. Thanks a lot!

luin:
Sorry for the late response. I've been thinking about whether this feature should be in a separate module since ioredis is focused on the official Redis features instead of the platform-specified ones.

fenichelar:
@luin The problem is the assumption that either Sentinel or Cluster mode are used if replication is used. Redis supports replication without Sentinel or Cluster mode but `ioredis` does not and therefore does not support read/write splitting for the following 3 use cases:
- Vanilla Redis without Sentinel but with master/slave replication (which is perfectly valid for read heavy workloads)
- Redis Enterprise
- AWS ElastiCache in non-cluster mode
While I haven't looked into the contents of the PR, I think the feature is well within the scope of scenarios `ioredis` should support natively.

STRML:
This is particularly useful if you just have heavy read or pub/sub flow on a local machine and need more cores to process the subscriptions.

fenichelar:
@mvaker I said you should probably rename this PR to "Add AWS non-clustered replication support" but I think that is misleading, my fault. Instead, "Add support for non-clustered replication without Sentinel" is probably more accurate. While AWS is a use case, there are other use cases where this PR is quite useful. @STRML just brought up yet another use case.

@luin I use Redis Enterprise and `ioredis` (hosted on AWS lol). At some point, I am going to have more read traffic on a single key than a single shard in Redis Enterprise can handle. So adding another shard to the Redis Enterprise cluster will not help. I will have to create a read replica of the cluster so the same key is stored on different shards. The replication is managed by Redis Enterprise, not Sentinel. Having `ioredis` support this natively would be very nice :)

caffeineshock:
Hey @luin, what's your current feeling about this pull request? Do you think it will get merged into the source eventually or are you rather inclined towards the separate modules solution?

We are currently stuck with a cluster that is not managed by sentinel and need to use read/write splitting quite soon. If this pull request might get merged soon, we would like to use the introduced feature. Otherwise we will have to split reading and writing manually by using different clients.

christoph-puppe:
Any news?

STRML:
FWIW, we worked on a fork of this PR for a while, and abandoned it in favor of just putting HAProxy in front of a bunch of read slaves, opening port 6378, and having all the pub/sub clients connect to that. That way, the adapter doesn't have to know anything (this logic is brittle, particularly in regards to reconnects).

azman0101:
Is there any chance that these use-case described in this issue will be one day supported ?

This redis chart https://github.com/helm/charts/tree/master/stable/redis also support replication cluster.

https://redis.io/topics/replication

PR #441, opened 2017-03-17
Add a default port for sentinel connections.

reconbot:
Also add a few helpful notes to the README. This came from some frustrations I had the first time I setup a project to talk to sentinels.

### DIFF
--- a/README.md
+++ b/README.md
@@ -590,7 +590,7 @@ var redis = new Redis({
 
 ## Sentinel
 ioredis supports Sentinel out of the box. It works transparently as all features that work when
-you connect to a single node also work when you connect to a sentinel group. Make sure to run Redis >= 2.8.12 if you want to use this feature.
+you connect to a single node also work when you connect to a sentinel group. Make sure to run Redis >= 2.8.12 if you want to use this feature. Sentinels have a default port of 26379.
 
 To connect using Sentinel, use:
 
--- a/lib/connectors/sentinel_connector.js
+++ b/lib/connectors/sentinel_connector.js
@@ -210,7 +210,7 @@ SentinelConnector.prototype.resolve = function (endpoint, callback) {
     Redis = require('../redis');
   }
   var client = new Redis({
-    port: endpoint.port,
+    port: endpoint.port || 26379,
     host: endpoint.host,
     retryStrategy: null,
     enableReadyCheck: false,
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -590,14 +590,15 @@ var redis = new Redis({
...
+you connect to a single node also work when you connect to a sentinel group. Make sure to run Redis >= 2.8.12 if you want to use this feature. Sentinels have a default port of 26379.
 
 To connect using Sentinel, use:
 
 ```javascript
 var redis = new Redis({
-  name: 'mymaster'
+  name: 'mymaster',
+  db: 0
```
luin: `db` defaults to 0 so why we add it here?

lib/connectors/sentinel_connector.js
```diff
@@ -36,7 +36,10 @@ SentinelConnector.prototype.connect = function (callback) {
     this.currentPoint = -1;
   }
   if (!Array.isArray(this.sentinels)) {
-    this.sentinels = this.options.sentinels;
+    this.sentinels = this.options.sentinels.map(function (endpoint) {
+      endpoint.port = endpoint.port || 26379;
+      return endpoint;
+    });
```
luin: Updating properties of the option introducing side effects. For example:

```javascript
const sentinels: [{host: '192.168.1.3'}]
const redis = new Redis({sentinels})

console.log(sentinels) // values changed unexpectedly
```

What about give the default port when creating a new instance by `port: endpoint.port || 26379`: https://github.com/luin/ioredis/blob/master/lib/connectors/sentinel_connector.js#L213.

There's an `isSentinelEql` function should be updated accordingly.

reconbot: We could copy the sentinel objects, but I get your point. I'll patch.

README.md
reconbot: As a new user to sentinels and not finding any docs for it in the api.md having this clue would have saved me a bit of headache.

luin: `db` option is not sentinel-specified and it's documented here: https://github.com/luin/ioredis/blob/master/API.md#new-redisport-host-options. What problem did you encounter?

reconbot: I didn't understand that fact. I thought sentinels overrode it. I was coming from a project that used urls.

luin:
Thank you for the pull request! All look good to me except the `db: 0` part. I'm afraid setting it explicitly will lead people into thinking `db` is required even if they just want to connect to the first database.

luin:
Aha! Merged! 🍻

Issue #445, opened 2017-03-31
Failed sentinel connection doesn't report Error


SamBergeron:
Hi, if I start a new connection with an array of sentinels where none are valid, no error is reported in the error listener. This is not the same behavior as with a single server connection.

Here is what I see with an invalid redis connection (or a down server) with debug enabled:
```
  ioredis:redis status[192.168.56.1:6379]: connecting -> close +1ms
  ioredis:connection reconnect in 400ms +0ms
  ioredis:redis status[192.168.56.1:6379]: close -> reconnecting +0ms
  ioredis:redis status[192.168.56.1:6379]: reconnecting -> connecting +400ms
  ioredis:connection error: Error: connect ECONNREFUSED 192.168.56.1:6379 +2ms
2017-03-31T15:49:03.012Z - error: Redis client error: Error: connect ECONNREFUSED 192.168.56.1:6379
```

The last log message being the one printed by `redis.on('error', ...`

Here is what I see when I have invalid (or down) sentinels 
```
  ioredis:redis status[192.168.56.1:26379]: [empty] -> connecting +219ms
  ioredis:redis queue command[0] -> sentinel(get-master-addr-by-name,nteredis) +0ms
  ioredis:connection error: Error: connect ECONNREFUSED 192.168.56.1:26379 +2ms
  ioredis:redis status[192.168.56.1:26379]: connecting -> close +0ms
  ioredis:redis status[192.168.56.1:26379]: close -> end +0ms
  ioredis:SentinelConnector failed to connect to sentinel 192.168.56.1:26379 because Error: Connection is closed. +1ms
  ioredis:redis status[192.168.56.1:26378]: [empty] -> connecting +0ms
  ioredis:redis queue command[0] -> sentinel(get-master-addr-by-name,nteredis) +0ms
  ioredis:connection error: Error: connect ECONNREFUSED 192.168.56.1:26378 +2ms
  ioredis:redis status[192.168.56.1:26378]: connecting -> close +0ms
  ioredis:redis status[192.168.56.1:26378]: close -> end +0ms
  ioredis:SentinelConnector failed to connect to sentinel 192.168.56.1:26378 because Error: Connection is closed. +0ms
  ioredis:SentinelConnector All sentinels are unreachable. Retrying from scratch after 230 +0ms
  ioredis:redis status[localhost:6379]: end -> end +1ms
```

As you can see there is no log message caught in  `redis.on('error', ...`.
[...]

PR #446, opened 2017-03-31
fix: report error on Sentinel connection refused (#445)

SamBergeron:
This is a quick fix to report on a sentinel cluster not being able to connect. 
This will also report retries in `redis.on('end', ...` for subsequent retries of the Sentinel cluster, this behavior is similar to what happens on a failed connection with a single redis instance.
This is the quickest fix I could think of.

### DIFF
--- a/lib/connectors/sentinel_connector.js
+++ b/lib/connectors/sentinel_connector.js
@@ -74,7 +74,6 @@ SentinelConnector.prototype.connect = function (callback) {
     _this.resolve(endpoint, function (err, resolved) {
       if (!_this.connecting) {
         callback(new Error(utils.CONNECTION_CLOSED_ERROR_MSG));
-        callback(new Error(utils.CONNECTION_CLOSED_ERROR_MSG));
         return;
       }
       if (resolved) {
--- a/lib/connectors/sentinel_connector.js
+++ b/lib/connectors/sentinel_connector.js
@@ -28,7 +28,7 @@ SentinelConnector.prototype.check = function (info) {
   return true;
 };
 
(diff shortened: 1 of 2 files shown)


lib/connectors/sentinel_connector.js
```diff
@@ -48,21 +48,26 @@ SentinelConnector.prototype.connect = function (callback) {
...
+        setTimeout(connectToNext, retryDelay);
       }
-      debug('All sentinels are unreachable. Retrying from scratch after %d', retryDelay);
-      setTimeout(connectToNext, retryDelay);
-      return;
+
+      if (lastError) {
+        error += ' Last error: ' + lastError.message;
+      }
+      return callback(new Error(error));
```
luin: We shouldn't `callback()` with an error here when we still keep retrying since a callback must not be called more than once.

Instead, what about just emit an event (like "sentinelError") when all sentinels are unreachable just like when a connection to a standalone Redis server is down.

SamBergeron:
Any idea why `1) should raise error when all sentinel are unreachable and retry is disabled` is failing?
I added a second callback to call the `silentEmit` function from redis.js. Mainly because with "normal" connections you callback an error everytime and loop inside redis.js, but with sentinels it's a recursive method. What do you think of this? Thanks.

SamBergeron:
Added a callback to `silentEmit` so that if none of the sentinels can connect we report an error, also added `sentinelError` reports to know which of the individual sentinels aren't connection if you're listening for that.

SamBergeron:
Just following up on this? Thoughts @luin ?

luin:
Sorry for the delay. Merged! 🍻

PR #453, opened 2017-04-28
command: handle errors to avoid `Unhandled rejection Errors`

achselschweisz:
I added the commit to get rid of `Unhandled rejection Errors` which can occur when issuing a `quit` call to the client, for instance.

I don't normally use promises, hence I cannot vouch for correctness here.

PS: I am currently using v2.5.0 of ioredis

### DIFF
--- a/lib/command.js
+++ b/lib/command.js
@@ -79,6 +79,12 @@ Command.prototype.initPromise = function () {
       _this.reject = reject;
     }
   }).nodeify(this.callback);
+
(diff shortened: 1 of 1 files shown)


luin:
The callback here may be called twice here

achselschweisz:
I added the PR to show you that the rejections need to be handled, I don't really have the time to dig through the depths of this - Cheers!
I'll remove the PR and convert it to a proper issue instead

PR #478, opened 2017-05-31
Require redis-parser@^3.0.0

christianpv:
Please read my comment in redis-parser for details: https://github.com/NodeRedis/node-redis-parser/pull/36#issuecomment-305188317

Update: I had to remove node versions < 4 from .travis.yml because `redis-parser@^3.0.0` is using ES6 features. Also, those versions are not being maintained: https://github.com/nodejs/LTS#lts-schedule1


luin:
Hi @christianpv! Thank you for the pull request.

I checked out your comments in redis-parser, however, I'm wondering what error did you encounter. Although `/lib/hiredis.js` requires hiredis, `lib/parser.js` wraps `require('./hiredis')` in a try-catch block, so no error would occur if 'hiredis' isn't specified as the parser name, and even it is, an error message will be printed to ask you to remove the `name` option.

christianpv:
This is error that I get while compiling my code:

ERROR in ./~/redis-parser/lib/hiredis.js
    Module not found: Error: Can't resolve 'hiredis' in '/Users/christianpv/homer/gapi/node_modules/redis-parser/lib'
     @ ./~/redis-parser/lib/hiredis.js 3:14-32
     @ ./~/redis-parser/lib/parser.js
     @ ./~/redis-parser/index.js
     @ ./~/ioredis/lib/redis/parser.js
     @ ./~/ioredis/lib/redis.js
     @ ./~/ioredis/index.js
     @ ./src/server/index.ts

luin:
I created a repo trying to reproduce to error: https://github.com/luin/test-ioredis-parser.

When I ran `node index.js`, the output are:

```
Hiredis parser is abandoned in ioredis v3.0, and JavaScript parser will be used
```

And no error occured. Is it your case?

BridgeAR:
@christianpv with what are you compiling your code? I guess you use a bundler and it resolves all dependencies up front and therefore does not detect, that the module may actually fail to load.

@luin in the parser v.3 hiredis does not include hiredis at all anymore.

christianpv:
@luin I'm using TypeScript, and Webpack. I will create a repo reproducing the error.

luin:
@christianpv Just as @BridgeAR said, you bundled all js files into one file. It's pretty uncommon in the backend world. Ignore the node_modules folder should solve the problem.

christianpv:
I understand what you are trying to said but its common to use Webpack for both.
https://stackoverflow.com/a/37920356
https://github.com/DxCx/webpack-graphql-server
https://github.com/kriasoft/react-starter-kit/blob/master/tools/webpack.config.js#L280
https://github.com/zeit/next.js#customizing-webpack-config

That is why Webpack provides the ability to specify the build output `node`.
I have four options. 

1. Continue to use my fork, and keep it to date until this PR gets merge.
2. Add `hiredis` to my dependencies.
3. Try using ignore plugin but I don't know if that will work: https://webpack.js.org/plugins/ignore-plugin/
4. This PR gets merge, and I update to latest version.

I hope this gets merge because I wouldn't like anyone else to get this issue. I'm going with option 2. Thanks a lot for your time!

luin:
@christianpv It makes sense to bundle your code into a single file, but it's not common that the code includes third-party modules as far as I know. For example, it's impossible to include native modules written in C/C++ into the output file (I'm a little curious about how you did since hiredis itself is a native module).

https://github.com/DxCx/webpack-graphql-server uses webpack-node-externals to ignore node_modules.

Dropping supports for older Node.js versions is a breaking change, so redis-parser v3 will be included in the next major version of ioredis. I will merge this pull request when we're prepared for that.

Anyway, I would recommend you to ignore files under node_modules folder currently.

christianpv:
@luin I'm not using the `hiredis` parser, I'm just adding it to my dependencies so that Webpack does not give me an error. Webpack includes all the modules that are require by you, and your dependencies in the final bundle.

shaylevi2:
Pleaseeeee merge this, the performance of the current parser is just horrible.

shaylevi2:
+1

PR #480, opened 2017-06-05
stop mutating the arguments when calling multi

reconbot:
This copies the arguments for a multi command instead of mutating the passed in arrays. This allows users to log the commands that gave an error if the multi command errors.


luin:
Thank you for the pull request! There are two missing trailing commas. Could you add them?

PR #482, opened 2017-06-09
ArgumentTransformer for zadd

### LINKED ISSUE #468: ArgumentTransformer for zadd
Doc for zadd: https://redis.io/commands/zadd

Hello,
The zadd command can accept endless `score member [score member ...]`. May I have a default ArgumentTransformer for zadd? I can create a PR for that.

My code:
```js
function convertFilpObjectToArray (obj) {
  var result = []
  for (var key in obj) {
    result.push(obj[key], key)
  }
  return result
}

Redis.Command.setArgumentTransformer('zadd', args => {
  if (args.length > 1) {
    let last = args.pop()
    if (typeof last === 'object' && last !== null) {
      args.concat(convertFilpObjectToArray(last))
    } else args.push(last)
  }
  return args
})
```

### LINKED ISSUE #481: [suggestion] Standard coding style
How about apply the Standard coding style to this project?
There are auto fix tool to use.

https://github.com/feross/standard

# The Rules

* 2 spaces – for indentation
* Single quotes for strings – except to avoid escaping
* No unused variables – this one catches tons of bugs!
* No semicolons – It's fine. Really!
* Never start a line with (, [, or `
  * This is the only gotcha with omitting semicolons – automatically checked for you!
  * [More details](https://github.com/feross/standard/blob/master/RULES.md#semicolons)
* Space after keywords if (condition) { ... }
* Space after function name function name (arg) { ... }
* Always use === instead of == – but obj == null is allowed to check * null || undefined.
* Always handle the node.js err function parameter
* Always prefix browser globals with window – except document and navigator are okay
  * Prevents accidental use of poorly-named browser globals like open, length, event, and name.

taichunmin:
Close #468 .


taichunmin:
I have no idea why zadd return undefined.
[Document](https://redis.io/commands/zadd) says return value is `int` or `string`.

PR #494, opened 2017-06-29
Switch from single lodash library to functions required

chris-olszewski:
No functional change, just slims down the install size.


AVVS:
usually people bundle lodash with their app and then it's deduped. That being said the change will actually increase the final bundle size given that scenario

chris-olszewski:
I would argue that if someone is already pulling in all of lodash instead of the functions they need, they don't care about the very minor bloat of 16 functions. 

Also, a quick experiment on a fresh project:
ioredis `3.1.1`
```
yarn why ioredis
yarn why v0.27.5
[1/4] Why do we have the module "ioredis"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
info Has been hoisted to "ioredis"
info This module exists because it's specified in "dependencies".
info Disk size without dependencies: "240kB"
info Disk size with unique dependencies: "6.16MB"
info Disk size with transitive dependencies: "6.18MB"
info Number of shared dependencies: 8
```
ioredis `git://github.com/chris-olszewski/ioredis.git#get_rid_of_monolithic_lodash`
```
yarn why ioredis
yarn why v0.27.5
[1/4] Why do we have the module "ioredis"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
info Has been hoisted to "ioredis"
info This module exists because it's specified in "dependencies".
info Disk size without dependencies: "580kB"
info Disk size with unique dependencies: "2.18MB"
info Disk size with transitive dependencies: "2.2MB"
info Number of shared dependencies: 23
```

Siilwyn:
Yes please, I was already wondering why `ioredis` is so much bigger than `node_redis`, `lodash` takes up 5MB.

luin:
Released in v3.1.3

PR #497, opened 2017-07-11
Handle stunnel options with sentinels

bluenodesolutions:
It would be great if ioredis could handle stunnel configuration when using sentinels.
If you pass in stunnel configuration data which is used on the resolve host/port after the sentinel has returned the Redis master IP address. You can not connect directly back to the Redis master as you need to go back through stunnel again. This allows you to pass your stunnel configuration and the correct stunnel port will be selected on your localhost. The example below assumes you are running three Redis servers.

new ioredis(
{
  name: 'mymaster',
  stunnel: [ 
   { redisHost: 'ip-redis1', redisPort: 6379, stunnelHost: 'localhost', stunnelPort: 6381 },
   { redisHost: 'ip-redis2', redisPort: 6379, stunnelHost: 'localhost', stunnelPort: 6382 },
   { redisHost: 'ip-redis3', redisPort: 6379, stunnelHost: 'localhost', stunnelPort: 6383 },
  ]
});

### DIFF
--- a/lib/connectors/sentinel_connector.js
+++ b/lib/connectors/sentinel_connector.js
@@ -80,6 +80,8 @@ SentinelConnector.prototype.connect = function (callback, eventEmitter) {
       }
       if (resolved) {
         debug('resolved: %s:%s', resolved.host, resolved.port);
+        resolved = _this.stunnel(resolved);
+        debug('resolved: %s:%s', resolved.host, resolved.port);
         _this.stream = net.createConnection(resolved);
         callback(null, _this.stream);
       } else {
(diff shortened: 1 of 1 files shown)


luin:
Thank you for the pull request!

It's a little confusing that the `stunnel` option has nothing to do with stunnel [itself](https://www.stunnel.org/index.html) (Although it solves the problem that the hosts resolved by the Redis Cluster are not the correct addresses that the clients should connect to, which is caused by stunnel).

As for me, a general option name would make more sense (e.g. `nodeAddressMap`). Anyway, I'm curious whether there's a better way to setup a redis cluster behind an SSL proxy which is transparent to the clients.

zag2art:
I'm for the PR

mblum14:
This would be great! we were forced to talk to redis over ssh and I would love for this to be accepted! I will donate my firstborn if this makes it in.

Issue #503, opened 2017-07-24
multi modifies input parameter


ccs018:
Using ioredis 3.1.1 / redis 4.0.

When using multi/exec, I build an array of commands and arguments to pass to the multi constructor:

```
    let pipeline = redis_controller.multi(redisCmds);
    pipeline.exec(function (err, results) { ... }
```
(I two-step this to more easily unit test and stub out with sinon.)

When the call to multi() returns, redisCmds has been modified.  Specifically, the first entry in each array of redisCmds (the "command") is removed.

Here's an example of what is passed into multi():

```
[
  [
    "set",
    "key1",
    "data1"
  ],
  [
    "zadd",
    "key2",
    1,
    "data2"
  ],
  [
    "zadd",
    "key3",
    1,
    "data3"
  ],
  [
    "zadd",
    "key4",
    1,
    "data4"
  ]
]
```

Here's what redisCmds looks like after the return from multi():

```
[
  [
    "key1",
    "data1"
  ],
  [
    "key2",
    1,
    "data2"
  ],
  [
    "key3",
    1,
    "data3"
  ],
  [
    "key4",
    1,
    "data4"
  ]
]
```

My code later examines the results from exec and tries to use the content of redisCmds to log metrics and details of which commands failed (if any).  However, since the 'commands' have been removed from redisCmds, I'm unable to provide the correct logging information.

I'll have to work around this for now by cloning the array, but it does not seem appropriate for multi() to be modifying the caller's data.

dirkbonhomme:
Nice find! As a general rule I think we should aim to have all input variables be immutable. (as in: ioredis should not modify them in any way)

luin:
Could you provide a test case for this? The following code:

```javascript
const Redis = require('ioredis')
const redis = new Redis()

const redisCmds = [
  [
    "set",
    "key1",
    "data1"
  ],
  [
    "zadd",
    "key2",
    1,
    "data2"
  ],
  [
    "zadd",
    "key3",
    1,
    "data3"
  ],
  [
    "zadd",
    "key4",
    1,
    "data4"
  ]
]
const pipeline = redis.multi(redisCmds)
pipeline.exec(function () {})

console.log(redisCmds)
```

Logs:

```javascript
[ [ 'set', 'key1', 'data1' ],
  [ 'zadd', 'key2', 1, 'data2' ],
  [ 'zadd', 'key3', 1, 'data3' ],
  [ 'zadd', 'key4', 1, 'data4' ] ]
```

[...]

PR #504, opened 2017-07-25
Do not modify input parameters

ccs018:
This should resolve issue #503.  The test app from #503 works.

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -182,7 +182,7 @@ Pipeline.prototype.sendCommand = function (command) {
 Pipeline.prototype.addBatch = function (commands) {
   var command, commandName, args;
   for (var i = 0; i < commands.length; ++i) {
-    command = commands[i];
+    command = commands[i].slice();
     commandName = command[0];
     args = command.slice(1);
     this[commandName].apply(this, args);


AVVS:
Change as it is will work, but, at the same time, there will be a performance penalty associated with this, as we'd have to create shallow copies of args for each command

@luin what do you think?

ccs018:
It appears that this PR isn't needed.  There is another commit on (June 8th) for #480 which also appears to resolve the issue and would probably explain why @luin wasn't able to reproduce it:  I'm using the released 3.1.1 and @luin was likely running off of the latest code.  3.1.1 does not include the commit for #480.

If that's the case, then this PR can be rejected.

3.1.1 code:
```
Pipeline.prototype.addBatch = function (commands) {
  for (var i = 0; i < commands.length; ++i) {
    var command = commands[i];
    var commandName = command.shift();
    this[commandName].apply(this, command);
  }

  return this;
};
```

Latest version:
```
Pipeline.prototype.addBatch = function (commands) {
  var command, commandName, args;
  for (var i = 0; i < commands.length; ++i) {
    command = commands[i];
    commandName = command[0];
    args = command.slice(1);
    this[commandName].apply(this, args);
  }

  return this;
};
```

luin:
@ccs018 Thank you for pointing this out! Just released v3.1.2. Sorry for not released this version when the fixed had been ma

ccs018:
Thanks for the updated release!

PR #507, opened 2017-08-10
Allow convertObjectToArray to handle objects with no prototype

southpolesteve:
Currently this method expects `hasOwnProperty` to exist on the passed object. By using the method on `Object.prototype` instead, objects created with `Object.create(null)` can also be properly handled.

Ran into this bug trying to pass graphQL input objects (heavily uses `Object.create(null)`) directly to hmset commands

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -170,10 +170,10 @@ exports.timeout = function (callback, timeout) {
  */
 exports.convertObjectToArray = function (obj) {
   var result = [];
-  for (var key in obj) {
-    if (obj.hasOwnProperty(key)) {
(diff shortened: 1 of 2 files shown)


luin:
Nice catch! What about using `Object.keys` to get the keys? See "Object.keys for loop" in https://jsperf.com/object-keys-vs-for-in-with-closure/3

southpolesteve:
I was curious how `reduce` would perform but it looks worse: https://jsperf.com/object-keys-reduce-vs-for-loop/1

Updated to use Object.keys + a for loop. Also added a test for an object without a prototype

Issue #521, opened 2017-09-14
Debug output unusable when commands args contain large strings


iamjochem:
hi there,

I love the fact that you use the `debug` module, it makes tracing what's going on in our own app as well as it's dependencies (in this case `ioredis`) a breeze.

I have stumbled upon a situation where I want to see the debug output from ioredis but due to the nature of what I happen to be storing the output is so large that it's making the console output unusable (actually my terminal regularly locks up in this particular situation).

### What am I doing?   

I'm storing "large" strings (JSON data) for caching purposes.

### What am I seeing?

I'm see debug output along the lines of:

```zsh
ioredis:redis write command[0] -> set("cache_data_1234",{"here_be":"lots of JSON",
.... hunderds/thousands of lines of output in my console ....
}) +337ms
```

### What would I like to see?

Something along the lines of:
```zsh
ioredis:redis write command[0] -> set("cache_data_1234",<large_arg_redacted>) +337ms
```

### Discuss?

Would you be willing to accept a PR that manipulates the `command.args` that are included in the `debug` output so that "large" strings are redacted? 

If yes, what do you feel regarding:

1. what the redaction string should be? (in my example I used `<large_arg_redacted>`)
2. what should be the length threshold for string threshold?
3. should the threshold be configurable?
4. should there be somekind of config option and/or ENV var that can be used to turn off the redaction completely?

thanks for taking the time to read, look forward to your reply!

luin:
PR is welcome! To your questions:

> what the redaction string should be? (in my example I used <large_arg_redacted>)

I'd recommend printing the heading several chars of the large argument to provide more details for developers to debug. For example: `foobarfoobarfoobar<...8292 more chars>`. (I'm not a native speaker so I'm open to any suggestions.

> what should be the length threshold for string threshold?

50~200 should be enough as for me.

> should the threshold be configurable?

I don't think an option is needed for that.

> should there be somekind of config option and/or ENV var that can be used to turn off the redaction completely?

The same answer as above. Let's make it default.

iamjochem:
hi @luin,

as promised here is a PR: https://github.com/luin/ioredis/pull/523

PR #523, opened 2017-09-25
truncate large/long debug output arguments

iamjochem:
wrap the debug node_module so that large/long debug output args are always truncated (avoids potentially MBs of data being outout to console).

this PR adds a module that offers a transparent wrap for the 3rd-party `debug` module, 
you just need to require `./utils/debug` in place of `debug` itself.

the idea for this PR was originally discussed here: https://github.com/luin/ioredis/issues/521

there are a couple of unit tests included.


luin:
Nice patch! However, it seems that these changes may have a impact on the performance even when the debug mode is off (e.g. truncating strings).

I just found out that the debug module supports custom formatters. What about define a custom formatter for the arguments?

For example:

```
const createDebug = require('debug');
createDebug.formatters.a = (v) => {
  return genRedactedString(v, len);
};

module.exports = createDebug;

iamjochem:
you are correct about the custom formatters - this is what I actually currently use to solve this issue in my own project. 

the reason I chose not to use custom formatters here is because they are applied "module" wide ... i.e. you may very well be overriding existing custom formatters or applying them (without asking) to all instances of "debug" used in a project's own code and all it's other dependencies (assuming they use the same version) ... I felt that this was something to be avoided (it is unfortunate that `debug` does not support custom formatters per 'instance')

the fix for the potential performance hit is easy enough - I pushed a commit that make the debug call a no-op (basically performing the same check as is done in the `debug` module itself) ... what do think?

luin:
Awesome! Makes sense to me. I'll do some tests this weekend, and merge then.

luin:
Released as v3.2.0. 🍺

iamjochem:
@luin - thanks for merging, it's a goog feeling to start the week with a successful PR :-)

PR #531, opened 2017-10-13
Update redis parser to latest version

shaylevi2:
Updating to latest version of redis parser


shaylevi2:
@luin is there a way to push this forward? 
redis-parser 3 has far better performance.

BridgeAR:
@shaylevi2 I am not sure why you think v3 is much faster than v2. ioredis supports older Node.js versions than the redis-parser v3. The main performance difference is only for errors and this should not have a big influence on any application. I am closing this as I recommend against changing this right now.

PR #594, opened 2018-03-06
Allow Monitor clients to be named

mgetz:
We are setting up some tools to monitor the health and activity of Redis and could not name a Monitoring client. This fixes that issue and lets the monitoring client be named just like rest of the client threads.

There is still some awkwardness in the naming process but at least it can be done after this change.

### DIFF
--- a/lib/redis/event_handler.js
+++ b/lib/redis/event_handler.js
@@ -120,6 +120,11 @@ exports.readyHandler = function (self) {
     self.setStatus('ready');
     self.retryAttempts = 0;
 
+    if (self.options.connectionName) {
+      debug('set the connection name [%s]', self.options.connectionName);
+      self.client('setname', self.options.connectionName);
+    }
+
     if (self.options.monitor) {
       self.call('monitor');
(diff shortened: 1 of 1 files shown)


shaharmor:
What awkwardness?

mgetz:
Thanks, Shahar!
The `monitor` command does not take an options parameter. So the only way I saw to set the name (after this submitted code change) is to change the connectionName on `this.redis.options` and then change it back after the instance has been duplicated. I also change the retryStrategy on the monitoring instances so that it does not come back to life after I kill it. 

Ideally, the monitor command would take an options parameter. 

Another way to do it would be to use the duplicate command and pass in options or create another instance with the monitoring settings of course.

Is there a better way to do it?
Thanks!
Mark

shaharmor:
Hey @mgetz,
Sorry for not being responsive for a few days.

Can you please add a test that checks the name is set for `monitor` clients here: https://github.com/luin/ioredis/blob/master/test/functional/connection.js

PR #595, opened 2018-03-07
fix(redis): attach connection handlers to stream events

szeist:
Hi,

Currently the connect() promise doesn't resolves after the connect event. Attaching the event handlers to the socket will solves this issue.

All the best,
Istvan

### DIFF
--- a/lib/redis.js
+++ b/lib/redis.js
@@ -309,8 +309,8 @@ Redis.prototype.connect = function (callback) {
         _this.removeListener(CONNECT_EVENT, connectionConnectHandler);
         reject(new Error(utils.CONNECTION_CLOSED_ERROR_MSG));
       };
-      stream.once(CONNECT_EVENT, connectionConnectHandler);
-      stream.once('close', connectionCloseHandler);
+      _this.once('connect', connectionConnectHandler);
+      _this.once('close', connectionCloseHandler);
     }, function (type, err) {
       _this.silentEmit(type, err);
     });
--- a/lib/redis.js
+++ b/lib/redis.js
@@ -309,8 +309,8 @@ Redis.prototype.connect = function (callback) {
         _this.removeListener(CONNECT_EVENT, connectionConnectHandler);
         reject(new Error(utils.CONNECTION_CLOSED_ERROR_MSG));
       };
-      _this.once(CONNECT_EVENT, connectionConnectHandler);
-      _this.once('close', connectionCloseHandler);
+      stream.once(CONNECT_EVENT, connectionConnectHandler);
+      stream.once('close', connectionCloseHandler);
     }, function (type, err) {
       _this.silentEmit(type, err);
     });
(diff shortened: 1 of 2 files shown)


shaharmor:
@szeist can you please add a test that checks it?

shaharmor:
Isn't this test verifies it? https://github.com/luin/ioredis/blob/master/test/functional/connection.js#L77-L95

szeist:
I don't think so. Because of the lazy connect we need to do some redis operations to trigger the connect.
I'll write a test that ensures this.

szeist:
I created test for the plain and for the TLS connection. Can you check them out?

luin:
`stream.once(CONNECT_EVENT, eventHandler.connectHandler(_this))` will trigger `_this.setState('connect')`, which will emit `connect` event so that `_this.once(CONNECT_EVENT, connectionConnectHandler)` can catch the event and resolve the promise.

We can just update `Redis#setStatus()`. When a 'connect' status is received, checking whether we are connecting via tls or not, and emit `secureConnect` or `connect` correspondingly.

szeist:
How about listening to the `connect` local event. It fired even on `secureConnect` and I agree with you that it's better to use here the local events instead of the underlying stream events.

shaharmor:
@szeist I still don't understand the difference between this:
https://github.com/szeist/ioredis/blob/b4402b12f23b90515b1641270b38b0b77e6ed160/test/unit/redis.js#L154-L165

and this:
https://github.com/luin/ioredis/blob/master/test/functional/connection.js#L78-L95

shaharmor:
@szeist I still don't understand the difference between this:
https://github.com/szeist/ioredis/blob/b4402b12f23b90515b1641270b38b0b77e6ed160/test/unit/redis.js#L154-L165

and this:
https://github.com/luin/ioredis/blob/master/test/functional/connection.js#L78-L95

szeist:
Hi,

The unit test uses the stubbed socket and it ensures the returned promise resolved on the `connect` event on the socket. The functional test connects to a redis server and tries to connect an invalid port the ensure the promise resolution and rejection. (For plain connection it worked fine w/o this pull request.

It created in pair with the unit test for the  `secureConnect` socket event. The reason for I created this test to make the promise resolution unit test more complete.

All the best,
Istvan

szeist:
What do you think, these modification are OK or I need to change something?

shaharmor:
They are fine, but I think @luin is not available. Haven't seen him in quite a while

szeist:
Hi,

Can we merge this in the near future?

lib/redis.js
```diff
@@ -310,7 +310,7 @@ Redis.prototype.connect = function (callback) {
         _this.removeListener(CONNECT_EVENT, connectionConnectHandler);
```
monkbroc: This should also be changed back to `'connect'` to avoid leaking a listener.

monkbroc:
Thanks for making the change @szeist 
I hope this gets merged soon :)

monkbroc:
Thanks @luin! I'm looking forward to the final release of 4.0.0.

PR #620, opened 2018-04-14
docs(README): add example for hgetall AoA reply transformer

silverwind:
When dealing with binary hash keys, the default reply transformer for `hgetall` is performing a unwanted conversion of hash keys to string.

I added a new reply transformer example to the README which shows how the reply format can be changed to array of arrays, it might be useful for other people.

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,6 +434,23 @@ redis.get('k3', function (err, result) {
 });
 ```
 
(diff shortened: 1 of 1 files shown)


silverwind:
The CI failure (something about CERT_UNTRUSTED) looks unrelated.

PR #636, opened 2018-06-07
Add NAT support for redis cluster

heri16:
Add support for redis clusters than are behind Network Address Translation (NAT).
This is useful for scenarios like this: https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/accessing-elasticache.html#access-from-outside-aws

### DIFF
--- a/README.md
+++ b/README.md
@@ -610,6 +610,19 @@ var redis = new Redis({
 });
 ```
 
+```javascript
+var cluster = new Redis.Cluster(nodesList, {
+  redisOptions: {
+    tls: {
+      // Refer to `tls.connect()` section in
+      // https://nodejs.org/api/tls.html
+      // for all supported options
+      ca: fs.readFileSync('cert.pem')
+    }
+  }
+});
+```
+
 <hr>
 
 ## Sentinel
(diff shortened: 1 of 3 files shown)


ccs018:
This only works if natRemap contains the entire set of nodes in the cluster.  In general, you don't need to specify all of the nodes in the nodeList argument when you do a `new Redis.Cluster()` and `ioredis` will dynamically learn the rest of the nodes in the cluster.  This solution essentially requires the client to be fully configured with all of the nodes.

Further, this solution does not allow for the cluster to dynamically change.

1) The size of the cluster can not increase unless natRemap is originally set to include all possible nodes that may exist in the future.

2) The NAT mapping is static.  If a node is ever restarted with a different IP address, the natRemap will need to be updated and the client restarted.

For rather static deployments, this appears to work, but the above limitations should be highlighted as this won't work in all deployments.

heri16:
@ccs018 Yes, I am well aware of these limitations. ElastiCache ensures that both the DNS name and the IP address of the cache node remain the same when cache nodes are recovered in case of failure. 

Adding shards while online would require a RegEx-based implementation that maps all matching auto-generated internal-DNS hostnames to the Natted ip address.

heri16:
Just in case others are wondering why-NAT, this is to mitigate cold-start-times on AWS Lambda, in particular the slow performance when initializing new VPC-connection. This means an extra long time for ioredis to successfully connect to redis cluster because the VPC network interface requires some time to initialize and be ready.

[See article](https://www.robertvojta.com/aws-journey-api-gateway-lambda-vpc-performance/)

heri16:
Will look into it. Thanks!

Issue #639, opened 2018-06-15
support node.js native Promise?


sophister:
hi , can `ioredis` work with native Promise of node.js

luin:
Released in v4.0.0-0. We using native Promise by default in this version.

Issue #637, opened 2018-06-07
Buffer keys support for Redis.Cluster


heri16:
Redis Class seems to support Both String and Buffer keys, but Redis.Cluster fails miserably if key provided is a Buffer.

Issue #408, opened 2016-12-02
dropping node < 4


AVVS:
@luin what do you think? Can this make it into v3?
I'm primarily concerned about dependencies and being able to use some of the recent es6 features

luin:
I definitely agree that we can drop the support for node < 4 so we don't need to worry about the BC for the older version, though I'm a little worried about the performance of es2015 with V8 (http://v8project.blogspot.kr/2016/12/v8-release-56.html?m=1).

abritinthebay:
This would also open up dropping the pointless (given what it is used for here) Bluebird library and using Native Promises. As far as I can tell Bluebird is only used for functionality that Native Promises provides out of the box.

AVVS:
bluebird is a) faster than native promises and b) has useful utils, so when you say pointless, thats highly debatable

abritinthebay:
The only parts it is faster on is because it doesn't follow the spec. Neither do those utils. Bluebird is NOT promises. It doesn't follow the spec. It *mostly* follows the spec. 

Following broken psuedo-implementations is an anti-pattern. Best to nip it in the bud as early as possible.

AVVS:
I hear you, but these are not arguments, this is a personal opinion. Explain your position, tell what broken, whats working not as expected? 
As far as I’m personally aware BB follows A+ spec and adds extra methods on top, which makes promises actually useful and pleasant to work with.

For example, basic promises are limited to .then, .catch and .all, but it leaves out all the actual usage patterns behind. Before promises control flow libs were highly popular due to hardly avoidable callback hell,
now they have their niche, but so do promises (when you can actually use them for control flow, like with BB). Of course, there is async/await, but its not released yet either, and it uses try/catch semantics, and that is debatable, too

So once again, please facts & arguments, not tunnel-visioned statements, such as BB is bad and native is good

abritinthebay:
>  it leaves out all the actual usage patterns behind

Great. Add those. Don't call it Promises, because it's not. Don't reimplement a (not-quite up to spec) version of Promises to add it.

> As far as I’m personally aware BB follows A+ spec and adds extra methods on top

Yeah, it doesn't. For example `new Promise` is an anti-pattern in Bluebird. Why? Because speed-wise it drops back down to native levels. Their benchmarks cheat a lot in this way. They also don't allocate arrays for promises' handlers. Meaning Promise branching is limited to incomplete.

[...]

Issue #581, opened 2018-02-02
Cluster: Failed to refresh slots cache. when options.redisOptions.enableOfflineQueue = false


ccs018:
If I pass `options` to new Cluster with `options.redisOptions.enableOfflineQueue = false`, ioredis (3.2.2) fails to connect to the cluster.  If I remove setting that specific option, then everything is good (went through them one-by-one).

Below is sample code with the output when `options.redisOptions.enableOfflineQueue = false`.

Note that I also have set `options.enableOfflineQueue = false` - it is not entirely clear if both need to be set.  In my usage, I do NOT want commands to be queued when the cluster or the intended node is not available.

```
'use strict';

let Redis = require('ioredis');

function _retryStrategy(times) {
    let retry;

    if (times === 1) {
        retry = 1;  // immediate retry
    } else {
        retry = Math.min((times - 1) * 2000, 20000);
    }
    return retry;
}

const options = {
    // Cluster-level options
    // Override ioredis defaults.
    clusterRetryStrategy   : _retryStrategy,
    enableOfflineQueue     : false,                     // ioredis default is true
    scaleReads             : 'all',                     // ioredis default is 'master'

    // ioredis defaults - explicitly set for clarity.
    enableReadyCheck       : true,
    maxRedirections        : 16,
    retryDelayOnFailover   : 100,
    retryDelayOnClusterDown: 100,
    retryDelayOnTryAgain   : 100,

    // Per Server options
    redisOptions: {
        // Override ioredis defaults.
        enableOfflineQueue           : false,           // ioredis default is true   !!! THIS IS THE PROBLEM !!!
        connectTimeout               : 2000,            // ioredis default is 10000
        autoResendUnfulfilledCommands: false,           // ioredis default is true
        retryStrategy                : _retryStrategy,
        readOnly                     : true,            // ioredis default is false

        // ioredis defaults - explicitly set for clarity.
        family           : 4,
        path             : null,
        keepAlive        : 0,
        noDelay          : true,
        db               : 0,
        dropBufferSupport: false,
        enableReadyCheck : true,
        autoResubscribe  : true,
        lazyConnect      : false,
        keyPrefix        : '',
        stringNumbers    : false
     }
};

let startupNodes = [];
for (let i = 0; i < 3; i++) {
    startupNodes[i] = {
        host: '127.0.0.1',
        port: 6400 + (10*i)
    };
    console.log('Using startupNode %s:%d.', startupNodes[i].host, startupNodes[i].port);
}

[...]

PR #642, opened 2018-06-23
WIP: v4

luin:
In this major version, we would drop the support for node 4 (#408) in order to use the built-in Promise instead of bluebird (#639 ).

Other plans are:
- [x] Support Per-Request retry limitations.
- [ ] Upgrade redis-parser to v3
- [x] Easier way to iterate the key space via `SCAN` and ES6 Iterators.
- [ ] Use TypeScript to rewrite this module for cleaner code and official type decroations

Cluster related:
- [x] Refresh slots automatically #641 
- [ ] Support for buffer keys #637 

Any other ideas? @shaharmor @Salakar @AVVS


shaharmor:
Cluster:
- Refresh slots upon a disconnection from one of the cluster nodes
- Restructure the cluster initialization code so that the initial [slots refresh call](https://github.com/luin/ioredis/blob/master/lib/cluster/index.js#L176) will be done only after connecting to the first node. This will help with issues like #581 where the user is setting `enableOfflineQueue: false` which causes the slots refresh call to fail immediately and disconnect from the node.
- Consider setting `enableOfflineQueue: true` [explicitly](https://github.com/luin/ioredis/blob/master/lib/cluster/connection_pool.js#L62) on the internal cluster nodes - @luin correct me if I'm wrong, but the offline queue of the internal node shouldn't be used in cluster mode because when ever the socket will disconnect the node will automatically be removed.

Other:
- Remove callback support? Not sure why we have to support callbacks if we support promises. If anyone really needs callback support, they can "callbackify" the promises themselves. Why do we need the entire library to support it?

luin:
> Refresh slots upon a disconnection from one of the cluster nodes

A disconnection won't lead the update of slots immediatially. A cluster will consider a master to be failing only after **NODE_TIMEOUT** has elapsed.

So most of time we only need refetch the slots in two different situations (https://redis.io/topics/cluster-spec#clients-first-connection-and-handling-of-redirections):
* At startup in order to populate the initial slots configuration.
* When a MOVED redirection is received.

However, given a failing of a slave won't trigger a `MOVED` error, so refresh slots intervally help us solve the problem.

> Remove callback support? Not sure why we have to support callbacks if we support promises. If anyone really needs callback support, they can "callbackify" the promises themselves. Why do we need the entire library to support it?

Good idea. I've seriously considered removing callback support. However, what makes me tend to keep callback support in this major version is that many built-in functions (`fs.readFile`) still follow callback-style, and `util.promisify` was absent until v8.0.0. The major frameworks (e.g. express, mocha) continue to be friendly to the callback style, so there should be many legacy codes prevent developers from migrating to the ioredis v4 if we drop callback support at this moment.

We may do this on the next major version, which only supports node >= v8 that support async functions.

luin:
Just release a beta version v4.0.0-0. The remaining issues will be resolved in the next beta version.

PR #647, opened 2018-06-28
wait for ready state before resolving cluster.connect()

shaharmor:


luin:
LGTM! We should also make `Redis#connect()` resolved only when the status is ready to keep the behavior same. I'll post a pull request for this.

luin:
I think the `reject()` should be called on the `end` event instead of `close` since when `close`, a reconnection will be made and the connection may eventually become `ready`.

luin:
Committed to https://github.com/luin/ioredis/pull/642 with cherry-pick. See also for the changes on `Redis`: https://github.com/luin/ioredis/pull/648

PR #648, opened 2018-06-29
feat: `Redis#connect()` will be resolved when status is ready

luin:
BREAKING CHANGE:
`Redis#connect()` will be resolved when status is ready
instead of `connect`:

```
const redis = new Redis({ lazyConnect: true })
redis.connect().then(() => {
  assert(redis.status === 'ready')
})
```

### DIFF
--- a/lib/redis.js
+++ b/lib/redis.js
@@ -306,15 +306,15 @@ Redis.prototype.connect = function (callback) {
         stream.setNoDelay(true);
       }
 
-      var connectionConnectHandler = function () {
+      var connectionReadyHandler = function () {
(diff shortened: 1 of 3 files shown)


lib/redis.js
```diff
@@ -305,10 +311,10 @@ Redis.prototype.connect = function (callback) {
         resolve();
       };
       var connectionCloseHandler = function () {
-        _this.removeListener(CONNECT_EVENT, connectionConnectHandler);
+        _this.removeListener('ready', connectionConnectHandler);
```
shaharmor: rename `connectionConnectHandler` to `connectionReadyHandler`

Issue #635, opened 2018-05-31
sentinel connect unhandled error


luckyscript:
```
Unhandled error event: Error: All sentinels are unreachable. Retrying from scratch after 10ms.
```
but I can reach all sentinels server,and all redis operator is ok.

so, I dont understand why this error shown frequently.

wish solutions, sincerely.

zluo001:
I‘m get the same problem, but the detail error message is
 Unhandled error event: Error: connect ETIMEDOUT
    at Socket.<anonymous> (/node_modules/ioredis/lib/redis.js:289:21) 
Also, I can access all sentinel servers.

luin:
Sorry for the late response. Could you enable the debug log (DEBUG=ioredis:* node yourapp.js) and post the logs here?

luckyscript:
``` 
ioredis:redis status[10.27.18.227:26379]: ready -> close +1ms
  ioredis:connection skip reconnecting since the connection is manually closed. +1ms
  ioredis:redis status[10.27.18.227:26379]: close -> end +0ms
  ioredis:redis status[10.37.65.92:6379]: connect -> ready +0ms
  ioredis:redis status[10.37.65.92:6379]: ready -> close +4s
  ioredis:connection reconnect in 50ms +6ms
  ioredis:redis status[10.37.65.92:6379]: close -> reconnecting +1ms
  ioredis:redis status[10.37.65.92:6379]: reconnecting -> connecting +52ms
  ioredis:SentinelConnector All sentinels are unreachable. Retrying from scratch after 10ms +0ms
[2018-07-02 09:02:14.104] [ERROR] console - [ioredis] Unhandled error event: Error: All sentinels are unreachable. Retrying from scratch after 10ms
    at connectToNext (E:\*\node_modules\ioredis\lib\connectors\sentinel_connector.js:64:19)
    at SentinelConnector.connect (E:\*\node_modules\ioredis\lib\connectors\sentinel_connector.js:44:3)
    at Redis.<anonymous> (E:\*\node_modules\ioredis\lib\redis.js:264:20)
    at Promise._execute (E:\*\node_modules\bluebird\js\release\debuggability.js:300:9)
    at Promise._resolveFromExecutor (E:\*\node_modules\bluebird\js\release\promise.js:483:18)
    at new Promise (E:\*\node_modules\bluebird\js\release\promise.js:79:10)
    at Redis.connect (E:*\node_modules\ioredis\lib\redis.js:250:10)
[...]

PR #652, opened 2018-07-02
fix sentinel connector bugs.

luckyscript:
fix https://github.com/luin/ioredis/issues/635
only `lastError` exist and currentPoint is the last , then throw a Error.

### DIFF
--- a/lib/connectors/sentinel_connector.js
+++ b/lib/connectors/sentinel_connector.js
@@ -47,27 +47,26 @@ SentinelConnector.prototype.connect = function (callback, eventEmitter) {
     _this.currentPoint += 1;
     if (_this.currentPoint === _this.sentinels.length) {
       _this.currentPoint = -1;
-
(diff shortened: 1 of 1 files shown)


lib/connectors/sentinel_connector.js
```diff
@@ -47,27 +47,26 @@ SentinelConnector.prototype.connect = function (callback, eventEmitter) {
     _this.currentPoint += 1;
     if (_this.currentPoint === _this.sentinels.length) {
       _this.currentPoint = -1;
+      if (lastError) {
```
luin: If there's no `lastError`, shouldn't we just connect to the first sentinel again instead of just returning?

luin: We may set `lastError` to `null` when resolved successfully.

luckyscript: Thanks for the check and your nice help.

Issue #587, opened 2018-02-16
Cluster: Fail to reconnect to node


ccs018:
I'm running a simple test with 3 masters + 1 slave per master.  If I take down one of the slaves and restart it, ioredis fails to reconnect to that node.

```
  ioredis:redis status[127.0.0.1:6421]: ready -> close +7s
  ioredis:connection skip reconnecting because `retryStrategy` is not a function +0ms
  ioredis:redis status[127.0.0.1:6421]: close -> end +0ms
Received -node: "127.0.0.1:6421"
```

**Note, however, that retryStrategy is a function.**  I've repeated this failure with 1) no options supplied new Cluster, 2) with options, but not setting options.redisOptions.retryFunction and 3) with option, but setting options.redisOptions.retryFunction to function() { return 3000; }.

Even long after the node has been restarted and rejoined the cluster (verified with redis-cli cluster slots and cluster nodes), ioredis continues to report the cluster.nodes().length = 5.

So, the only way to recover from this very basic failure mode is for the application to monitor the '-node' event and destroy and re-create the client.  This is a bad defect.

I would have hoped that ioredis was periodically monitoring the cluster configuration by periodically performing a cluster slots command, it is clearly impossible for ioredis to detect if new slaves are ever added to the cluster in an expansion scenario.

ccs018:
Keep alive.  This issue needs to be resolved.

ccs018:
Keep alive. This issue still needs to be resolved.

ccs018:
freshbot

shaharmor:
Hey @ccs018, as far as I know, ioredis only refreshes the cluster nodes list when there is a slot change.

Can you try to change one of the slots and see if ioredis reconnects to that slave?

You are correct that we should add a refresh interval for the cluster nodes list, I'll see if I can hook something up

ccs018:
@shaharmor , thanks for the response.  I could try a slot change, but I believe this also is not immediately detected, but only if an operation results in a MOVED response.

Note that I've raised a couple other related issues regarding how ioredis manages cluster connections.  In this particular issue, there is no real change in the cluster topology.  Rather one of the nodes failed and ioredis decides to never attempt to reconnect to that node.  If you query any node in cluster, that temporarily failed node was never removed from the cluster, yet ioredis decides that it is gone forever.  In this scenario, ioredis should not unilaterally remove the node from its view of the cluster, but should continue to retry re-connecting to that node.  As I noted above error may be related to the following debug output:
[...]

PR #658, opened 2018-07-06
feat: add "timeoutPerCommand" option to detect dead connection

luin:
Usage:

```
const redis = new Redis({
  timeoutPerCommand: 10000
})
```


lib/redis/event_handler.js
```diff
@@ -39,6 +39,25 @@ exports.connectHandler = function (self) {
...
+      self.timeoutPerCommand = setInterval(() => {
+        var commandQueueLength = self.commandQueue.length;
+        if (
+          commandQueueLength > 0 &&
+          Date.now() - self.lastWriteTime > self.options.timeoutPerCommand
+        ) {
+          debug('Command timed out. Pending commands: %s', commandQueueLength);
+          var err = new Error('Command timed out');
+          self.flushQueue(err, { offlineQueue: false });
+          self.silentEmit('error', err);
```
ccs018: Is emitting an `error` event sufficient?  In my usage, I'm always supplying a callback in the form of:
```
redis.get(key, function (err, reply) {
    if (err) {
        // handle error
    } else {
        // handle data
    }
});
```
While I do also handle the `error` event, that handling is somewhat generic.  It would be better if the callback was called with the error so that specific error handling could be invoked.

lib/redis/event_handler.js
```diff
@@ -39,6 +39,25 @@ exports.connectHandler = function (self) {
...
+          self.disconnect(true);
```
ccs018: I assume that this will trigger an automatic reconnect to that server, correct?

ccs018: Good, but see my follow-up to #587.  If this node was part of a cluster, `retryStrategy` is overridden and set to `null` so there would in fact be no automatic reconnect to that node.

vweevers:
Is `socket.setKeepAlive(true)` not an option? TCP keep-alive probes are very cheap and work well.

lib/redis/event_handler.js
```diff
@@ -39,6 +39,25 @@ exports.connectHandler = function (self) {
...
+        }
+      }, 500);
```
ccs018: It appears that the minimum practical value for `timeoutPerCommand` is 500ms.  Even if I specify `timeoutPerCommand = 100`, the check for a timed out command occurs every 500ms and so the command time out error won't trigger as quickly as I would expect.

lib/redis/event_handler.js
```diff
@@ -39,6 +39,25 @@ exports.connectHandler = function (self) {
...
 
+    self.lastWriteTime = null;
+    if (typeof self.options.timeoutPerCommand === 'number') {
+      var timeoutPerCommand = self.options.timeoutPerCommand;
+      debug('Per-command timeout set to %s', self.options.timeoutPerCommand);
+      self.timeoutPerCommand = setInterval(() => {
+        var commandQueueLength = self.commandQueue.length;
+        if (
+          commandQueueLength > 0 &&
+          Date.now() - self.lastWriteTime > self.options.timeoutPerCommand
```
ccs018: Since `self.lastWriteTime` is updated every time a command is added to the `commandQueue`, if the client attempts commands fast enough (more than one every `timeoutPerCommand` ms), then `self.lastWriteTime` continues to get updated such that `Date.now() - self.lastWriteTime` will continue to be less than `timeoutPerCommand` and thus the original command will never time out.

Perhaps in `Redis.prototype.sendCommand` `lastWriteTime` is only updated if `commandQueue.length === 1`.

ccs018:
@vweevers While a little old, see https://github.com/nodejs/node-v0.x-archive/issues/6194.  The main take away is that node.js doesn't provide sufficient configurability of the TCP keepalive functionality.  Thus we'd have to wait up 10-11 minutes before detecting that the connection has closed.  Searching through the node 10.x documentation, I didn't find anything that allows further configuring TCP keep alives to detect a broken connection any faster.

vweevers:
@ccs018 Oh I see, the issue is about detecting dead connections *faster*. That's cool. I just hope the added functionality will remain opt-in (because for my needs, waiting 10 minutes is perfectly fine) and not add side effects to an already complicated module.

ccs018:
It's not sure much about detecting dead connections fast.  It's I issue a redis command (e.g., get), but if the connection dies before the response is received nothing is reported back to the client - the callback is never invoked.

luin:
@vweevers This option will be disabled by default and it won't add any side effects when disabled.

@ccs018 If you issue a command, The `timeoutPerCommand` interval will check if the response has been received every 500ms, and if not after the specified time (`timeoutPerCommand` option), the callback will be invoked with a timeout error.

vweevers:
@ccs018 This must mean TCP keep-alive is disabled by default (client side).

@luin Does ioredis expose the raw socket somewhere, so that I can call `setKeepAlive` on it?

luin:
@vweevers Keep-alive is enabled by default. Refer to https://github.com/luin/ioredis/blob/master/API.md for details on the options for that.

lib/redis/event_handler.js
```diff
@@ -39,6 +39,25 @@ exports.connectHandler = function (self) {
 
     self.initParser();
 
+    self.lastWriteTime = null;
+    if (typeof self.options.timeoutPerCommand === 'number') {
+      var timeoutPerCommand = self.options.timeoutPerCommand;
+      debug('Per-command timeout set to %s', self.options.timeoutPerCommand);
+      self.timeoutPerCommand = setInterval(() => {
```
soulrebel: how about using this: https://nodejs.org/api/net.html#net_socket_settimeout_timeout_callback to generate a timeout event from the socket if it becomes inactive while the queue is being processed?

ccs018: This timeout guards against a socket which has no activity.  The value to use probably greatly depends on the use cases.  In a production environment, there's probably enough traffic to keep such a timeout from triggering.  But in a dev environment, there might not be any traffic for relatively long stretches of time making this an idle connection which then times out.  Setting a per command timeout seems better.

Also, if this socket timeout occurs, it doesn't enable specific error handling / recovery for the specific commands which timed out.

Or were you suggesting this as an additional check and not an alternative solution?

lib/redis/event_handler.js
ccs018: Per #587, there will not be any automatic reconnect in a cluster environment.

lib/redis/event_handler.js
soulrebel: I am suggesting to set this to something reasonable while we are processing commands and to disable it while the socket is not used.
It just seems that it can be easier to implement than setting and cancelling timeouts in the lib. And more powerful as well.

silverwind:
I have a feeling that this should be a per-command option. I could certainly see using different timeouts for different commands.

I'm currently using a wrapper function that includes a `Promise` constructor and a `setTimeout`, but seeing that ioredis does not expose a way to cancel a pending command, there is a risk that it would accumulate lots of pending commands over time and eventually exhaust the system's  resources.

silverwind:
I guess we could do without this new option if commands were cancellable. A common way to do this on promise interfaces is to expose a `.cancel` method on the promise. There's also [p-cancellable](https://github.com/sindresorhus/p-cancelable) which adds a bit more.

edevil:
I still think that a configurable global command timeout is useful when dealing with unresponsive servers because there are some modules that take a Redis proxy object and issue the commands themselves. All these modules would need to be updated to correctly guard for timeouts.

edevil:
ping

andrewaustin:
@luin Any chance of reviving this one?

BryanDonovan:
In case anyone stumbles upon this issue, I think it's closed by https://github.com/luin/ioredis/pull/1320

seminarian:
Cancelling the promise after `commandTimeout` was hit would be nice..

PR #662, opened 2018-07-09
Update connection_pool.js

ccs018:
Do not force `retryStrategy` to be `null`.  If the node had simply restarted, then ioredis may never reconnect to this node.  Waiting for a `MOVED` error to appear to trigger a reconnect is a false assumption as a `MOVED` error may never appear - especially in simple node restart scenarios.


shaharmor:
@ccs018 we now refresh the slots map every 5s which will find that missing node and re-join it.

ccs018:
Good point. Was that change merged in?  I didn't see it mentioned in the release notes for the 4.0 RCs.

shaharmor:
Yes it was merge, and you are correct it wasn't mentioned.

shaharmor:
@ccs018 any objections to close this PR now?

ccs018:
No, no objections.  I'll need to test out the RC.

PR #667, opened 2018-07-20
Export ScanStream at package level

elbuo8:
Purpose: If you want to handle your own ScanStream (particularly when using a Cluster), having this exported makes it a bit more cleaner.


luin:
Thanks for the pull request!

PR #681, opened 2018-08-06
Mention duplicate keys in scan stream example

philraj:
I've used a Set to handle the deduplication in the example, but I can switch it to a plain old object as I wrote in the comment if that feels less out of place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -488,16 +488,18 @@ provides a streaming interface for the `SCAN` command to make things much easier
 var redis = new Redis();
 // Create a readable stream (object mode)
 var stream = redis.scanStream();
-var keys = [];
+var keys = new Set();
 stream.on('data', function (resultKeys) {
   // `resultKeys` is an array of strings representing key names.
-  // Note that resultKeys may contain 0 keys.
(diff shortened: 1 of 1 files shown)


vweevers:
Using a Set to record visited keys is one solution, but having to keep all keys in memory can be a huge drawback. Another solution is to make your operations on the keys idempotent, so that it doesn't matter if you visit a key 1 time or 10 times.

What the best solution is depends on the use case, and I don't think ioredis docs should focus on that. Instead briefly explain the problem and leave it up to consumers to solve it.

philraj:
@vweevers That's fair, but the original example was pushing them into an array, which would use even more memory (since duplicates are kept). I'm good with removing the storage of keys from the example entirely, simply logging from the `.on('data')` handler, and leaving the detail about duplicates as a simple comment.

vweevers:
> I'm good with removing the storage of keys from the example entirely, simply logging from the .on('data') handler

Sounds good! That's also closer to the intention of streams

luin:
Awesome! And thanks @vweevers for the advice!

PR #691, opened 2018-08-29
when password contain ':'

zsea:
when password contain ':',password set error.


zsea:
```javascript
var url = require("url").URL;
var options= new url("redis://:%3a123456789127.0.0.1/1");
```
it can read password and username

Issue #621, opened 2018-04-17
Adding new builtin commands and transformers


ksmithut:
I am currently playing around with [redis streams](https://gist.github.com/antirez/4e7049ce4fce4aa61bf0cfbc3672e64d) (and [here](https://gist.github.com/antirez/68e67f3251d10f026861be2d0fe0d2f4) which add some new commands. I don't think these commands are available for production use, but I assume they'll be added to the list of commands currently in the `redis-commands` module when they're ready.

Until then, I'm currently looking to add these commands to my ioredis instance. What is the best way to do this that will support things like pipelines and transactions? The way I got it to work seems hacky (using and overriding prototypes), but this is what I've got so far:

```js
const { string, buffer } = Redis.prototype.createBuiltinCommand('xadd')
Redis.prototype.xadd = string
Redis.prototype.xaddBuffer = buffer
Redis.Command.setArgumentTransformer('xadd', args => /* TODO */ args)
Redis.Command.setReplyTransformer('xadd', reply => /* TODO */ reply)
```

ksmithut:
It appears that this doesn't add it to the multi pipelining. It would be great to have an answer for this, even if the answer is "we don't support adding custom commands".

ksmithut:
```js

luin:
Sorry for the late response! Yes, adding commands to pipeline directly should do the trick. XADD command will be added as soon as it's added into the redis-commands package.

ksmithut:
Awesome, thanks :)

PR #692, opened 2018-08-29
deps: upgrade redis-commands package

flintinatux:
This should open up access to Redis 5.0 streams as requested by #621, among others.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,9 +22,9 @@
       "integrity": "sha1-DNkKVhCT810KmSVsIrcGlDP60Rc=",
       "dev": true,
       "requires": {
-        "kind-of": "3.2.2",
-        "longest": "1.0.1",
-        "repeat-string": "1.6.1"
+        "kind-of": "^3.0.2",
+        "longest": "^1.0.1",
+        "repeat-string": "^1.5.2"
       }
     },
     "amdefine": {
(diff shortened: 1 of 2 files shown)


luin:
🍻

lakano:
Hello !
We need to use the Redis 5.0 streams commands. Currently on stable ioredis version 4.0.0, I can't uses the Redis 5.0 commands ( tested XADD in a redis.multi([ XADD ... ]).exec() )
Your commit sent one month ago seems to fix the problem, so I would like to known when a new version of ioredis will be submitted to permit us to uses it please?

lakano:
Hello!
Thanks for the version 4.0.1 published 3 hours ago.
But, I don't understood, I can't uses « bzpopmin » command ( added in redis 5.0 ), and it's inside the package redis-commands 1.3.5 referenced in your package.json ( https://github.com/NodeRedis/redis-commands/blob/651d1592b7103cf045e0b0be0ca0532a19e6bfe9/commands.json ) 
Is it normal or do I need to report a bug on ioredis ( or may be in redis-commands ) ?
Thanks for your help!

luin:
@lakano What the version of redis-commands installed in your project (you can see it via `npm ls | grep redis-commands`)?

lakano:
The last one, 1.3.5, it's why it's really strange.. :-/

lakano:
Nervermind, it's my Docker cache, sorry for the waste of time and thanks you for the upgrade! :)

lakano:
Finally, I think there is a bug, I've recreated a simple example without Docker and from scratch, the bug is still present. Can you take a look on #716 please?

PR #697, opened 2018-09-14
fix(Cluster): don't fetch node info from the subscribe

### LINKED ISSUE #696: Refreshing nodes on interval timer breaks socket.io pubsub adapter
Using socket.io-adapter causes the redis connection to enter subscriber only mode, any other command causes an error, and a reconnect.

> Error: Connection in subscriber mode, only subscriber commands may be used

Therefore the ioredis interval timer which refreshes cluster nodes every 5 seconds causes an error, and continual reconnections due to the fact that is runs `CLUSTER NODES` which is not allowed.

Not sure how to resolve this, given the purpose of the interval – but I thought you should be aware.

luin:
This fix will skip the subscriber who can't execute any command.

Close #696


jeremytm:
@luin I might be misunderstanding the code, but I'm pretty sure this doesn't skip the subscriber node, and only creates a debug message.

shaharmor:
@jeremytm you are correct

hhgyu:
@luin
https://github.com/luin/ioredis/pull/697#discussion-diff-218673494R417
The problem occurs when querying the list of clusters, the attempt is from ioredis, is not an error, and makes it difficult to identify it when the actual error occurs.

stockholmux:
Seeing similar things... what's the status on this?

luin:
Have fully tested with socket.io-redis. Merged and released in v4.0.1. Feel free to report any issues in this. version

PR #710, opened 2018-09-27
fix: endless cluster reconnect - race condition

AVVS:
I do have concerns about circuit breaker, in case all nodes are stuck in disconnected state. But I think it should be fine, because nodes should get cleaned up and the error would be different once there are no nodes or error is not caused due to .disconnect being called prior to the command

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,6 +2,9 @@
   "env": {
     "node": true
   },
+  "parserOptions": {
+    "ecmaVersion": 6
+  },
   "rules": {
     "no-const-assign": 2,
     "strict": [2, "global"],
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 node_modules
 *.cpuprofile
 /test.js
+/coverage
 built
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,6 @@ node_modules
 *.cpuprofile
 /test.js
 /coverage
+/.nyc_output
+/.npmrc
 built
--- a/.nycrc
+++ b/.nycrc
@@ -1,0 +1,18 @@
+{
+  "extension": [
+    ".js",
+    ".ts",
+    ".tsx"
+  ],
+  "include": [
+    "lib/**"
+  ],
+  "exclude": [
+    "**/*.d.ts"
+  ],
+  "reporter": [
+    "lcov",
+    "text-summary"
+  ],
+  "all": true
+}
--- a/lib/cluster/ConnectionPool.ts
+++ b/lib/cluster/ConnectionPool.ts
@@ -2,9 +2,12 @@ import {parseURL} from '../utils'
 import {EventEmitter} from 'events'
 import {noop, defaults} from '../utils/lodash'
 import {NODE_STATUS} from '../utils'
+import {multi} from 'event-to-promise'
 
 const Redis = require('../redis')
 const debug = require('../utils/debug')('ioredis:cluster:connectionPool')
+const maxInt = 2147483647
+const resolveStatus = [NODE_STATUS.WAIT, NODE_STATUS.CONNECT, NODE_STATUS.READY, NODE_STATUS.END]
 
 type NODE_TYPE = 'all' | 'master' | 'slave'
 
@@ -25,6 +28,8 @@ export default class ConnectionPool extends EventEmitter {
   }
 
   private specifiedOptions: {[key: string]: any} = {}
+  private ready: boolean = false
+  private iteration: number = 0
 
   constructor (private redisOptions) {
     super()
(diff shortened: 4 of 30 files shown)


AVVS:
@luin what do you think about this fix? the flow is generally the following:

* cluster connects with hostnames:port
* starts ready check
* starts connecting to ipaddress:port
* disconnects hostname based connection
* ready check returns error due to .disconnect()
* cluster.disconnect() is called forcing connection shut down across all nodes

endlessly repeats this sequence, unless ready check has not started or already finished before .disconnect on the hostname:port

Another fix that might be a bit more reliable is to resolve hostname to ipaddress before connecting to redis cluster, most likely forming a correct node key right away. But it would still be a problem in case that ipaddress is different from what the redis cluster returns

luin:
This change makes sense to me for fixing the issue. However, I'm wondering there may be a case that there is only one node in the cluster (or only one node that we can reach from the given node), `this._readyCheck` will be triggered endless.

AVVS:
I think in case of only one node that node should be removed from the pool completely after the failure? Or is it not how it works - I admit I don't follow the complete logic for maintaining pool of connections yet so maybe there is something that I'm missing

AVVS:
BTW, in the tests 1 case seem to be particularly flaky, any idea why is that?

```
  1) cluster:pipeline
       should retry when redis is down:
      Uncaught AssertionError: expected [ null, 'OK' ] to deeply equal [ null, 'bar' ]
      + expected - actual
       [
         [null]
      -  "OK"
      +  "bar"
       ]
      
      at /home/travis/build/luin/ioredis/test/functional/cluster/pipeline.js:207:28
      at tryCatcher (node_modules/standard-as-callback/lib/utils.js:10:19)
      at /home/travis/build/luin/ioredis/node_modules/standard-as-callback/index.js:20:31
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)
```

luin:
The test above failed because the new code skips the reconnection. However, the test will make node2 alive again when reconnect (see the stub function of `refreshSlotsCache`). Since the node2 doesn't become alive, ioredis will pick a node random. So node1 will be picked, which will return 'OK' for both `get foo` & `set foo bar` since we don't specify the reply handler for these commands.

AVVS:
One other possible solution I see is resolving hostname -> ip address before connecting, since we know that redis cluster would be returning ip addresses either way, what do you think?

@luin that way there should not be a readyCheck triggered on a node that we already know is going to be disconnected from the list, however, if it does get disconnected for some other reason - it could still be the case of endless retries with that race condition. 

Another stab at this could possibly be delaying readyCheck until all of the nodes in the pool have "connected" state

AVVS:
@luin reworked proposed fix a bit, I think it should handle the problem a bit better

* chore: all literals that I've found were replaced with string constants, but probably a few more should be replaced, where there are .on/once listeners
* once stream is up (sentinel / redis) the node key is updated if it is different, thus removing the need for the node to be disconnected

.eslintrc
```diff
@@ -2,6 +2,9 @@
   "env": {
     "node": true
   },
+  "parserOptions": {
+    "ecmaVersion": 6
```
AVVS: did that because my eslint was going nuts :) since we do support only node 6+ - that shouldnt be a problem at all, and I assume once codebase has been completely migrated to ts eslint would go completely

AVVS:
alright, this seem to be working rather well now, except that I still can see some issues in somewhat random places - one of the tests for monitor & max retries would bug out at times with a timeout error. However, this is also the case with non-modified code

Noteworthy - updated TS to 3.x, added sourcemaps, nothing seems to be breaking, sinon to 6 (5 major versions jump), istanbul of 0.4.x to nyc with most recent versions of it. Locally tests seem to be passing well, waiting for CI

AVVS:
not the best piece of my code, but I think the last solution thats been implemented should be the best. Basically it waits before all nodes are in some sort of final "pre-cluster connected" state and only then tries to do readyCheck

I haven't implemented cancelling promises since its not in the spec, but it shouldn't leak since it must be always resolving with any status - either connect or failure

AVVS:
did some more testing and that specific solution still doesn't eliminate the problem completely, guess need to make some sort of combination between them or automatic redirect of node-agnostic commands on the "manual" disconnect, such as the cluster status one

luin:
Do you think resolving the hostnames to IPs before connecting could solve the problem?

AVVS:
it will work for some cases, but I dont think its fail proof. In my opinion it would be great to somehow rework the ready check, specifically to make sure it doesn't trigger full reconnect when one node is manually disconnected due to configuration mismatch

AVVS:
> As for the rework of ready check, in which case this pull request doesn't eliminate the problem?

I was thinking that when some node is faulty and wont let us connect for some reason its possible for command to be dispatched to it and then go in circles

PR #717, opened 2018-10-09
refactor: migrate cluster-related code to TypeScript

luin:


lib/cluster/index.ts
```diff
@@ -0,0 +1,663 @@
...
+var asCallback = require('standard-as-callback');
+var ClusterSubscriber = require('./ClusterSubscriber').default;
+var DelayQueue = require('./DelayQueue').default;
+var PromiseContainer = require('../promiseContainer');
+var {AbortError} = require('redis-errors');
+
+/**
+ * Creates a Redis Cluster instance
+ *
+ * @constructor
```
SimonSchick: You can probably omit most of those `@constructor` and other class related hints.

lib/cluster/index.ts
```diff
@@ -0,0 +1,663 @@
+import ClusterAllFailedError from '../errors/ClusterAllFailedError'
+import {defaults, noop} from '../utils/lodash'
+import ConnectionPool from './ConnectionPool'
+import {NodeKey, IRedisOptions, normalizeNodeOptions} from './util'
+
+var Deque = require('denque')
```
SimonSchick: Some of those requires can probably be moved to import (namely node and some of your own).
Everything should probably be `const`

lib/cluster/index.ts
```diff
@@ -0,0 +1,663 @@
...
+      this.subscriber.start();
+    });
+  };
+
+  /**
+   * Called when closed to check whether a reconnection should be made
+   *
+   * @private
+   */
+  _handleCloseEvent () {
```
SimonSchick: You can get rid of `@public` and `@private` and use the keywords instead.

lib/cluster/index.ts
```diff
@@ -0,0 +1,646 @@
...
+      setImmediate(function () {
+        this.setStatus('close')
+        this.handleCloseEvent()
+      }.bind(this))
+
+      return ret
+    }
+    return asCallback(
+      Promise.all(this.nodes().map(function (node) {
+        return node.quit()
```
AVVS: this sometimes would throw if node that we issue quit to is already in the disconnected state (often in the tests where redis instance is down before the quit is processed)

lib/cluster/index.ts
```diff
@@ -0,0 +1,646 @@
...
+   * Change cluster instance's status
+   *
+   * @private
+   * @param {ClusterStatus} status
+   * @memberof Cluster
+   */
+  private setStatus(status: ClusterStatus): void {
+    debug('status: %s -> %s', this.status || '[empty]', status)
+    this.status = status
+    process.nextTick(this.emit.bind(this, status))
```
AVVS: this.emit.bind could be saved as a reference, ie boundEmit
and then we can `process.nextTick(this.boundEmit, status)` - that should be a bit faster

lib/cluster/index.ts
luin: Good catch! I'll fix this in a separate commit.

PR #721, opened 2018-10-14
perf(cluster): make disconnecting from cluster faster

luin:
This commit will prevent status changes from 'disconnecting'
to 'connecting' and will stop connecting to a nodes
when disconnecting, which will make disconnecting faster.

### DIFF
--- a/lib/cluster/index.ts
+++ b/lib/cluster/index.ts
@@ -348,8 +348,12 @@ class Cluster extends EventEmitter {
       const node = nodes[index]
       debug('getting slot cache from %s:%s', node.options.host, node.options.port)
       _this.getInfoFromNode(node, function (err) {
-        if (_this.status === 'end') {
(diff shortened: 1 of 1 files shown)


lib/cluster/index.ts
```diff
@@ -348,8 +348,12 @@ class Cluster extends EventEmitter {
       const node = nodes[index]
       debug('getting slot cache from %s:%s', node.options.host, node.options.port)
       _this.getInfoFromNode(node, function (err) {
-        if (_this.status === 'end') {
-          return wrapper(new Error('Cluster is disconnected.'))
+        switch (_this.status) {
+          case 'close':
+          case 'end':
```
AVVS: Same error returned for 3 statuses, was this intentional?

luin: "disconnecting" returns a different error.

AVVS: ah, didn't notice the difference at the end

PR #722, opened 2018-10-15
chore: automatic deploy with semantic-release

AVVS:
Tried to remain release cycle the same as it was, added deploy section to .travis.yml (not sure it works - please review this part especially)

Idea is that every new release will get a tag on github and will be released under @next on the npm so its installable right away

Once you are sure its stable - just merge the same code to release branch and it will promote latest from next tag. I think this should be good enough


.travis.yml
```diff
@@ -29,3 +29,19 @@ addons:
...
 after_success:
 - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
+
+deploy:
+  # runs semantic-release on each
+  - provider: script
+    script: npm ci && npm run semantic-release
+    on:
+      tags: false
+      all_branches: true
```
luin: Shouldn't branch limit to "master"?

AVVS: It will do a dry run if branch is not master (marked in releaserc)

luin: Why we need a dry run?

AVVS: to check that the setup is not broken (ie in case we update some of packages to incompatible versions)

luin: Will the dry run works when GITHUB_TOKEN & NPM_TOKEN is absent?

luin:
> I'd advise to make these env variables only available on the master/release branches so that people wont be able to steal them with malicious PR

It seems Travis-CI doesn't support setting the scope of env variables.

AVVS:
https://github.com/semantic-release/condition-travis - there is more info here on how to set things up and I do think it has the travis specific setup with encryption and stuff

AVVS:
https://github.com/semantic-release/semantic-release/blob/caribou/.travis.yml example of how they set it up on their own repo

https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions - this also states that only people who send PRs with write access to the repo are "trustworthy" and the secrets are only exposed to PRs like that

luin:
Merged!

AVVS:
Checked the deploy - a few things - it did seem to only detect v4.0.0 as the last release, which is not correct. Can we specifically tell it to push the tag from git?

Add

`"getLastRelease": "@semantic-release/git"` to the releaserc for that

PR #723, opened 2018-10-15
feat: support customize dns lookup function

luin:
Since Redis cluster doesn't support hostname at all (antirez/redis#2410),
it's reasonable to resolve the hostnames to IPs before connecting.


lib/cluster/ClusterOptions.ts
```diff
@@ -108,5 +128,6 @@ export const DEFAULT_CLUSTER_OPTIONS: IClusterOptions = {
   retryDelayOnClusterDown: 100,
   retryDelayOnTryAgain: 100,
   slotsRefreshTimeout: 1000,
-  slotsRefreshInterval: 5000
+  slotsRefreshInterval: 5000,
+  dnsLookup: lookup
```
AVVS: maybe wrap it as a promise here via `util.promisify` (I dont remember when it became available) or just as a

```ts
dnslookup(hostname: string): Promise<string> => {
    lookup(hostname, (err, address) => {
          if (err) {
            debug('failed to resolve hostname %s to IP: %s', hostname, err.message)
            reject(err)
          } else {
            debug('resolved hostname %s to IP %s', hostname, address)
            resolve(address)
          }
        })
  });
},
```

AVVS: and then the code would be cleaner in the actual resolve function

luin:
4.2.0 is released with the "latest" tag instead of "next". Besides that, everything works like a charm. @AVVS

AVVS:
need to make sure that .npmrc is present and I've configured it correctly (ie maybe dist-tag=next) is a wrong way to set it :) but yeah, pretty cool 👍

lib/cluster/index.ts
```diff
@@ -117,59 +129,68 @@ class Cluster extends EventEmitter {
         reject(new Error('Redis is already connecting/connected'))
         return
       }
+      const epoch = ++this.connectionEpoch
```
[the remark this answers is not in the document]
glensc: permalink at the time of the posting:
- https://github.com/fastify/fastify/blob/main@%7B2018-10-15T18:21:52Z%7D/lib/logger.js#L32-L38

PR #749, opened 2018-11-14
Fix MOVED slot redirection handler

vflopes:
The moved handler must update the targetSlot before tryConnection call to correctly choose the node that owns the slot.


shaharmor:
Is this a new bug?

Issue #365, opened 2016-09-08
ioredis client should not attempt to connect to IP addresses from cluster configuration


bsnote:
I configured Redis cluster consisting of 2 nodes running on the same machine:

**redis.conf for 1st node:**

```
bind 0.0.0.0
port 6379
cluster-enabled yes
cluster-config-file nodes.conf
cluster-node-timeout 5000
```

**redis.conf for 2nd node:**

```
bind 0.0.0.0
port 6380
cluster-enabled yes
cluster-config-file nodes.conf
cluster-node-timeout 5000
```

```
redis-cli -p 6379 cluster addslots $(seq 0 8191)
redis-cli -p 6380 cluster addslots $(seq 8192 16383)
redis-cli -p 6380 cluster meet 127.0.0.1 6379
```

`redis-cli -p 6379 cluster slots` returns the following result:

```
1) 1) (integer) 8192
   2) (integer) 16383
   3) 1) "127.0.0.1"
      2) (integer) 6380
      3) "c8e9a312411cefdd8ee4f74c923e5b7099db5edd"
2) 1) (integer) 0
   2) (integer) 8191
   3) 1) "127.0.0.1"
      2) (integer) 6379
      3) "e44691fe2b1adb2772fd5114d577dab105482f32"
```

Now running `DEBUG=ioredis:* node 1` on a different machine where

``` javascript
// 1.js
var Redis = require('ioredis');

var cluster = new Redis.Cluster([{
  port: 6379,
  host: '10.2.19.39'
}]);
```

shows the following:

```
  ioredis:cluster status: [empty] -> connecting +0ms
  ioredis:redis status[10.2.19.39:6379]: [empty] -> wait +6ms
  ioredis:cluster getting slot cache from 10.2.19.39:6379 +3ms
  ioredis:redis status[10.2.19.39:6379]: wait -> connecting +3ms
  ioredis:redis queue command[0] -> cluster(slots) +1ms
  ioredis:redis status[10.2.19.39:6379]: connecting -> connect +7ms
  ioredis:redis write command[0] -> info() +1ms
  ioredis:redis status[10.2.19.39:6379]: connect -> ready +7ms
  ioredis:connection send 1 commands in offline queue +0ms
  ioredis:redis write command[0] -> cluster(slots) +1ms
  ioredis:redis status[127.0.0.1:6380]: [empty] -> wait +4ms
  ioredis:redis status[127.0.0.1:6379]: [empty] -> wait +0ms
[...]

PR #758, opened 2018-12-08
feat(cluster): add NAT support

### LINKED ISSUE #693: Can't connect to ElastiCache redis cluster via NAT instance
We're using a single node cluster to test our workflow, and cannot connect via a NAT instance.

Having done some debugging, we get an error immediately after the node is connected:

> Disconnect ec2-XX-X-XX-XX.compute-1.amazonaws.com:6379 because the node does not hold any slot

Looking at the code this happens because the actual cluster node has an IP address from inside our VPC, while the NAT instance which we use to establish connectivity has a different public hostname. Because ioredis relies on the ip address of each node to create keys (or slots in the context of this error message), it thinks the node is missing and immediately disconnects it.

Is this something you guys would ever fix? I'm thinking some sort of map specified in connection options which maps the NAT hostname and port to the real VPC IP and ports could be a solution?

**For anyone asking why we want to run local code with a remote redis implementation:**

Usually our code runs remotely in AWS lambda. However our fastest dev workflow is where we test locally for convenience. Usually a local redis server is enough, but we regularly need to connect to a remote redis instance to debug with live data – and uploading our full repo to lambda on every iteration is way too slow.

Up until now we've used public redis instances in redislabs. However, we're moving to a more scalable solution in ElastiCache with redis cluster (which means running inside a VPC) and would like to be able to achieve the same fast workflow.

luin:
Closes #693, #365

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ used in the world's biggest online commerce company [Alibaba](http://www.alibaba
 0. Support for ES6 types, such as `Map` and `Set`.
 0. Support for GEO commands (Redis 3.2 Unstable).
 0. Sophisticated error handling strategy.
+0. Support for NAT mapping.
 
 # Links
 * [API Documentation](API.md)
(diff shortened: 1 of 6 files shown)


heri16:
@luin I'll be testing to make sure this PR does not break end-to-end TLS connection over NAT.

Issue #740, opened 2018-10-26
error: Error: connect ETIMEDOUT


kasarlashushruth:
This the sample code i am using
node version v9.2.0
ioredis version 4.2.0
express 4.16.3

var ssl = {
    key: privatekey,
    // secureProtocol: 'TLSv1_method',
    // strictSSL: false,
    passphrase: pass,
    cert: certificate,
    ca
};
var sentinel_list = [
    { host: 'host', port: port_number }
];
bd_name = 'some name';
var client = new Redis({
    sentinels: sentinel_list,
    password: "password",
    // socket_keepalive: true,
    name: bd_name,
    // showFriendlyErrorStack: true,
    connectTimeout: 10000,
    tls: ssl
});
client.on('error', function(err){
    console.log('Something went wrong ', err);
});
client.on('connect', function() {
    console.log('Connected to Redis');
});
set and get methods bellow

When ever i try to connect it gives "error: Error: connect ETIMEDOUT"
Trials i have done
1) Increased connectTimeout from 3 sec to 1 hour same error
2) The certs are working if connect with java jedis
3) Tried lot combinations of sending certs and key with or without encryption and chained etc one time just key and ca one time with all key cert and ca , chained once but same error

The ioredis Debug logs are bellow

> DEBUG=ioredis:* node app.js

  ioredis:redis status[localhost:6379]: [empty] -> connecting +0ms
  ioredis:redis status[host:port]: [empty] -> connecting +2ms
  ioredis:redis queue command[host:port]: 0 -> sentinel([ 'get-master-addr-by-name', 'name' ]) +1ms
  ioredis:redis queue command[localhost:6379]: 0 -> set([ 'foo1', 'shushruth' ]) +1ms
  ioredis:redis queue command[localhost:6379]: 0 -> get([ 'foo1' ]) +0ms
  ioredis:redis status[ip]: connecting -> connect +249ms
  ioredis:redis status[ip]: connect -> ready +1ms
  ioredis:connection send 1 commands in offline queue +0ms
  ioredis:redis write command[ip]: 0 -> sentinel([ 'get-master-addr-by-name', 'name' ]) +0ms
  ioredis:redis write command[ip]: 0 -> sentinel([ 'sentinels', 'name' ]) +39ms
  ioredis:SentinelConnector Updated internal sentinels: [{"host":"host","port":port}] @1 +0ms
  ioredis:SentinelConnector resolved: ip +1ms
  ioredis:redis status[ip]: ready -> close +187ms
[...]

PR #772, opened 2019-01-06
Allow TLS when using Sentinel

fenichelar:
Fixes https://github.com/luin/ioredis/issues/740

### DIFF
--- a/lib/connectors/SentinelConnector/index.ts
+++ b/lib/connectors/SentinelConnector/index.ts
@@ -28,6 +28,8 @@ interface ISentinelConnectionOptions extends ITcpConnectionOptions {
   sentinelRetryStrategy?: (retryAttempts: number) => number
   preferredSlaves?: PreferredSlaves
   connectTimeout?: number
+  enableTLSForSentinelMode: false
+  sentinelTLS: undefined
 }
 
 export default class SentinelConnector extends AbstractConnector {
@@ -104,7 +106,7 @@ export default class SentinelConnector extends AbstractConnector {
         }
         if (resolved) {
           debug('resolved: %s:%s', resolved.host, resolved.port)
-          if (_this.options.tls) {
+          if (_this.options.enableTLSForSentinelMode && _this.options.tls) {
             Object.assign(resolved, _this.options.tls)
             _this.stream = createTLSConnection(resolved)
           } else {
@@ -198,7 +200,7 @@ export default class SentinelConnector extends AbstractConnector {
       port: endpoint.port || 26379,
       host: endpoint.host,
       family: endpoint.family || (isIIpcConnectionOptions(this.options) ? undefined : this.options.family),
-      tls: this.options.tls,
+      tls: this.options.sentinelTLS,
       retryStrategy: null,
       enableReadyCheck: false,
       connectTimeout: this.options.connectTimeout,
--- a/lib/connectors/SentinelConnector/index.ts
+++ b/lib/connectors/SentinelConnector/index.ts
@@ -198,6 +198,7 @@ export default class SentinelConnector extends AbstractConnector {
       port: endpoint.port || 26379,
       host: endpoint.host,
       family: endpoint.family || (isIIpcConnectionOptions(this.options) ? undefined : this.options.family),
+      tls: this.options.tls,
       retryStrategy: null,
       enableReadyCheck: false,
       connectTimeout: this.options.connectTimeout,
--- a/lib/connectors/SentinelConnector/index.ts
+++ b/lib/connectors/SentinelConnector/index.ts
@@ -1,6 +1,6 @@
 import {createConnection, Socket} from 'net'
 import {CONNECTION_CLOSED_ERROR_MSG, packObject, sample} from '../../utils'
-import {TLSSocket} from 'tls'
+import {connect as createTLSConnection, TLSSocket} from 'tls'
 import {ITcpConnectionOptions, isIIpcConnectionOptions} from '../StandaloneConnector'
 import SentinelIterator from './SentinelIterator'
 import {ISentinelAddress} from './types';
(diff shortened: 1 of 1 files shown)


fenichelar:
Tested with Redis Enterprise and is working as expected. Redis Enterprise requires client certificates. Code used for testing:

```js
const fs = require('fs');
const ioredis = require('ioredis');

const redisClient = new ioredis({
  sentinels: [
    {
      host: '10.0.1.1',
      port: 8001,
    },
    {
      host: '10.0.1.2',
      port: 8001,
    },
    {
      host: '10.0.1.3',
      port: 8001,
    },
  ],
  name: 'test',
  tls: {
    key: fs.readFileSync('key.pem').toString(),
    cert: fs.readFileSync('cert.pem').toString(),
    ca: fs.readFileSync('ca.pem').toString(),
    secureProtocol: 'TLSv1_2_method',
    checkServerIdentity: () => {
      return undefined;
    },
  },
});

redisClient.on('error', err => {
  console.error(err);
});

redisClient.on('ready', () => {
  console.log('ready');
  process.exit(0);
});
```

luin:
Thanks for the pull request! A question here, does connecting to sentinel servers need tls support?

fenichelar:
@luin I don't think so. In Redis Enterprise, you enable TLS per database. However, only a single Sentinel runs per node. If you have 2 database on a single node, one can require TLS and the other can not (which is what I was using for testing actually). I don't think there is an option to require TLS for the connection to Sentinel, but I'll look at the documentation again. That would be nice though because a MITM attack against Sentinel could be quite bad.

fenichelar:
Also, I should clarify, for Redis Enterprise it is not actually Sentinel, it is Discovery Service. The Discovery Service API is intentionally compliant with Sentinel but it isn't actually the same thing. See https://docs.redislabs.com/latest/rs/concepts/data-access/discovery-service/.

fenichelar:
It worked! While it doesn't appear to be possible to require TLS connections to the Discovery Service, it appears to be possible to use TLS to connect.

Should I modify this PR to encrypt the connection to Sentinel / Discovery Service if TLS options are supplied or make a seperate PR?

fenichelar:
It was so easy to implement I just added it here.

Please let me know if there is anything else.

lib/connectors/SentinelConnector/index.ts
```diff
@@ -104,7 +104,12 @@ export default class SentinelConnector extends AbstractConnector {
         }
         if (resolved) {
           debug('resolved: %s:%s', resolved.host, resolved.port)
-          _this.stream = createConnection(resolved)
+          if (_this.options.tls) {
```
luin: if (_this.options.enableTLSForSentinel && _this.options.tls) {

lib/connectors/SentinelConnector/index.ts
```diff
@@ -193,6 +198,7 @@ export default class SentinelConnector extends AbstractConnector {
       port: endpoint.port || 26379,
       host: endpoint.host,
       family: endpoint.family || (isIIpcConnectionOptions(this.options) ? undefined : this.options.family),
+      tls: this.options.tls,
```
luin: tls: this.options.enableTLSForSentinel && this.options.tls

fenichelar:
This makes sense. However, what if someone wanted just the connection to Redis to be encrypted, not Sentinel? Or the other way around?

What if we add `enableTLSForSentinel` and `enableTLSForRedis`? Both will default to `false` for this version. In v5, switch them both to default to `true`. Thoughts?

luin:
We can provide two options in this version:
1. `enableTLSForSentinelMode` (default value is `false`). If true, the `tls` option will be used to connect to the redis server that resolved from sentinels.
2. `sentinelTLS` (default value is `undefined`). Similar to the `tls` option but used for connecting to sentinels.

In v5, `enableTLSForSentinelMode` will be removed and `sentinelTLS` will remain intact. I've changed the comments I left below the code accordingly. What do you think?

fenichelar:
Makes sense. How's that look?

lib/connectors/SentinelConnector/index.ts
```diff
@@ -104,7 +106,12 @@ export default class SentinelConnector extends AbstractConnector {
         }
         if (resolved) {
           debug('resolved: %s:%s', resolved.host, resolved.port)
-          _this.stream = createConnection(resolved)
+          if (_this.options.enableTLSForSentinelMode && _this.options.tls) {
+            Object.assign(resolved, _this.options.tls)
```
luin: Shouldn't `Object.assign(resolved, _this.options.tls)` be `Object.assign(resolved, {tls: _this.options.tls})`?

fenichelar: Hahaha, that's what I thought too and I spent so much time trying to figure out why it wasn't working!!

https://nodejs.org/api/tls.html#tls_tls_connect_options_callback

> secureContext: TLS context object created with tls.createSecureContext(). If a secureContext is not provided, one will be created by passing the entire options object to tls.createSecureContext().


The entire options object is passed to `createSecureContext()`.

fenichelar: The `StandardConnector` works the same: https://github.com/luin/ioredis/blob/master/lib/connectors/StandaloneConnector.ts#L45-L47

fenichelar:
@luin Awesome. I'm trying to setup a new Redis cluster this week with TLS. When do you think a new version of ioredis could be released with this change? Thanks.

luin:
The new version has been released automatically: v4.5.0. The version is released under the next tag (can be installed via npm i `ioredis@next` or `npm i ioredis@4.5.0`) and will be released to the latest tag in a month.

fenichelar:
Oh, awesome!! Thanks :)

PR #777, opened 2019-01-09
fix sentinel connect logic

luckyscript:
I think when `_this.sentinelIterator.hasNext()` is false, just reset the cursor to zero and don't need to emitter an error.
only when `LastError` is defined, we emitter an error.

when connector resolved, we should set LastError to null.


luin:
有遇到什么问题吗？`LastError` 为空但是还是遍历下一个 sentinel 一般是因为配置出错导致 sentinel 返回了空结果。这个时候感觉调用 `sentinelRetryStrategy` 让开发者介入会比较好？

Issue #778, opened 2019-01-10
defineCommand  loads lua scripts several times


akaNightmare:
Hello, 
while executing some custom lua scripts several times in pipeline can see that ioredis:
1. runs `SCRIPT EXISTS` command with the multiple same sha1 hashes
2. runs `SCRIPT LOAD` for each of custom command (in the example below -> `.test()`) in pipeline 

### So questions are:
1. Is it expected behaviour?
2. Is it possible to run `SCRIPT EXISTS` command with uniq sha1 hashes?
3. Is it possible to run `SCRIPT LOAD` only once for same sha1 hashes?

### Small repro
```js
const Redis = require('ioredis');

const redis = new Redis('redis://127.0.0.1:6379');
redis.defineCommand('test', {
    lua: 'return {KEYS[1],KEYS[2],unpack(ARGV)}'
});

(async () => {
    const pipe = redis.pipeline();
    pipe.test(2, 'k1', 'k2', 'a1', 'a2');
    pipe.test(1, 'k1', 'a1');
    pipe.test(3, 'k1', 'k2', 'k3', 'a1', 'a2', 'a3');
    await pipe.exec();
    process.exit(0);
})();

```

### Redis Logs
```
  ioredis:redis status[127.0.0.1:6379]: [empty] -> connecting +0ms
  ioredis:redis queue command[127.0.0.1:6379]: 0 -> script([ 'exists', '86efd0ce7b213e7c68b64afe58b81b21feafeae8', '86efd0ce7b213e7c68b64afe58b81b21feafeae8', '86efd0ce7b213e7c68b64afe58b81b21feafeae8' ]) +4ms
  ioredis:redis status[127.0.0.1:6379]: connecting -> connect +8ms
  ioredis:redis write command[127.0.0.1:6379]: 0 -> info([]) +0ms
  ioredis:redis status[127.0.0.1:6379]: connect -> ready +3ms
  ioredis:connection send 1 commands in offline queue +0ms
  ioredis:redis write command[127.0.0.1:6379]: 0 -> script([ 'exists', '86efd0ce7b213e7c68b64afe58b81b21feafeae8', '86efd0ce7b213e7c68b64afe58b81b21feafeae8', '86efd0ce7b213e7c68b64afe58b81b21feafeae8' ]) +1ms
  ioredis:redis write command[127.0.0.1:6379]: 0 -> script([ 'load', 'return {KEYS[1],KEYS[2],unpack(ARGV)}' ]) +0ms
[...]

PR #781, opened 2019-01-11
(#778) perf: add checking and loading scripts uniqueness in pipeline

akaNightmare:
https://github.com/luin/ioredis/issues/778

### DIFF
--- a/lib/pipeline.js
+++ b/lib/pipeline.js
@@ -266,19 +266,19 @@ Pipeline.prototype.exec = function (callback) {
     return execPipeline();
   }
 
-  return this.redis.script('exists', scripts.map(function (item) {
-    return item.sha;
-  })).then(function (results) {
-    var pending = [];
-    for (var i = 0; i < results.length; ++i) {
-      if (!results[i]) {
-        pending.push(scripts[i]);
+  return this.redis
+    .script('exists', Array.from(new Set(scripts.map(({ sha }) => sha))))
+    .then(function (results) {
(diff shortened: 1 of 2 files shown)


lib/pipeline.js
```diff
@@ -268,6 +268,8 @@ Pipeline.prototype.exec = function (callback) {
 
   return this.redis.script('exists', scripts.map(function (item) {
     return item.sha;
+  }).filter(function (sha, index, shaList) {
+    return shaList.indexOf(sha) === index;
   })).then(function (results) {
```
luin:
```javascript
scripts.map(function (item) {
    return item.sha;
  }).filter(function (sha, index, shaList) {
    return shaList.indexOf(sha) === index;
  })
```

Would be better to change to

```javascript
Array.from(new Set(scripts.map(({sha}) => sha)))
```

test/functional/pipeline.js
```diff
@@ -12,7 +12,7 @@ describe('pipeline', function () {
         [null, 3],
         [null, '3']
       ]);
-      redis.disconnect()
+      redis.disconnect();
```
luin: Don't add `;` here. Code style changes should belong to a separate pull request.

lib/pipeline.js
akaNightmare: resolved

test/functional/pipeline.js
akaNightmare: resolved

PR #784, opened 2019-01-21
feat: add retryTime option when redis server not ready

### LINKED ISSUE #783: Max redis server still loading
When connecting to Redis, if the redis server is still loading / starting up and not ready to handle connections, then ioredis will use the `info.loading_eta_seconds` property from the `info` command to determine the amount of time to wait before retrying.

The problem is, I'm seeing a very large `info.loading_eta_seconds` returned from redis  - e.g. 1349000 / 22 mins.

```
2019-01-20T23:30:46.207Z ioredis:redis Redis server still loading, trying again in 1349000ms
```

Could we add an option to configure a maxRedisServerLoading, or provide a fn to customize the retry duration for the serve loading scenario?

tuananh:
attempt to fix #783


luin:
What about giving a default value for this option (like 10s)? So people won't be blocked to long if they don't notice this option.

PR #789, opened 2019-01-29
fix: more verbose log when emit node error

### LINKED ISSUE #774: Improve event "node error" to know which node (host+port) of cluster fired it
When "node error" event happens on cluster node there is no info in its argument about which node error issued.

```
redis.on('node error', (error) => {
    log('Redis cluster node error:', error.message);
});
// Redis cluster node error: Connection is closed.
// Redis cluster node error: timeout
```
I tried to log entire "error" object but there is no info about node. It is important to quickly know host and port of node which has some problems.

tuananh:
attempt to fix #774


luin:
Thanks for the pull request! Created another pull request for this issue. The new approach works when the debug mode is disabled.

PR #799, opened 2019-02-19
nat support for sentinel connector

shults:


lib/connectors/SentinelConnector/index.ts
```diff
@@ -46,6 +48,10 @@ export default class SentinelConnector extends AbstractConnector {
       throw new Error('Requires the name of master.')
     }
 
+    if (this.options.natMap && Object.keys(this.options.natMap).length === 0) {
+      throw new Error('Empty natMap is not allowed.')
```
luin: We should allow empty `natMap` (`{}`) to keep the same behavior as one in cluster: https://github.com/luin/ioredis/blob/master/lib/cluster/index.ts#L428.

shults: Thanks for reply. Fixed.

PR #816, opened 2019-03-17
support Sentinel instances with authentication.

ynomura-md:
According to the [document](https://redis.io/topics/sentinel), Sentinel instances with authentication has been available since redis version 5.0.1.

I would like to support it.

Would you confirm this PR.

### DIFF
--- a/lib/connectors/SentinelConnector/index.ts
+++ b/lib/connectors/SentinelConnector/index.ts
@@ -25,6 +25,7 @@ type PreferredSlaves =
 interface ISentinelConnectionOptions extends ITcpConnectionOptions {
   role: 'master' | 'slave'
   name: 'string'
+  sentinelPasseword?: 'string'
   sentinels: Array<ISentinelAddress>
   sentinelRetryStrategy?: (retryAttempts: number) => number
   preferredSlaves?: PreferredSlaves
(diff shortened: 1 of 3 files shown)


lib/connectors/SentinelConnector/index.ts
```diff
@@ -25,6 +25,7 @@ type PreferredSlaves =
 interface ISentinelConnectionOptions extends ITcpConnectionOptions {
   role: 'master' | 'slave'
   name: 'string'
+  sentinelPasseword?: 'string'
```
luin: Typo here. `sentinelPasseword` should be `sentinelPassword` I guess?

ynomura-md: I'm sorry. It is typo. fixed.

test/functional/sentinel.js
```diff
@@ -186,7 +185,25 @@ describe('sentinel', function () {
         name: 'master'
       });
     });
+    it('should connect to sentinel with authentication successfully', function (done) {
+      var sentinel = new MockServer(27379, function (argv) {
+        if (argv[0] === 'auth' && argv[1] === 'pass') {
+          authed = true;
+        }
```
luin: Missing assertion here. Should define `authed` as `false` in the outer block, and add an assertion of `expect(authed).to.eql(true)` when `argv[0] === 'sentinel' && argv[1] === 'get-master-addr-by-name'` to make sure ioredis sends `AUTH` before asking for master address.

Close the connections here instead of on `sentinel.once('connect')`.

luin: Simply make sure the test will fail when there's a typo in option name 😄

ynomura-md:
I will cancel this PR once.
I write the test and put it out again.

Issue #824, opened 2019-03-25
4.9.1 requires git to install


oznu:
Hello,

Commit 93ecd70e0e454541bd4934f9b09f746b3ad7c151 in release 4.9.1 adds `git` as a system dependency to successfully install the package.

Without git installed users can expect this error:

```
npm ERR! code ENOGIT
npm ERR! Error while executing:
npm ERR! undefined ls-remote -h -t ssh://git@github.com/mercadolibre/flexbuffer-node.git
npm ERR! 
npm ERR! undefined
npm ERR! No git binary found in $PATH
npm ERR! 
npm ERR! Failed using git.
npm ERR! Please check if you have git installed and in your PATH.
```

This might break existing automated builds, in my case it was a Docker image that previously had no need for git to be installed.

luin:
Good caught! Just downgrade latest tag to 4.9.0. Will move the dependencies to npm in the next version.

kerihenare:
This is solved in #826. However, I'm also looking at [refactoring out **flexbuffer** entirely](https://github.com/luin/ioredis/compare/master...kerihenare:buffer-refactor?expand=1). It's a real small package, and **ioredis** only uses an even smaller portion of it.

gabegorelick:
A better approach may be to refactor usages of flexbuffer to use Buffer and related JS builtins directly. Especially if it's not used much. 

For example, https://github.com/luin/ioredis/blob/7736c1cf27af1cb991b95f592a5fbe89646facf6/lib/pipeline.ts#L306 appears to [copy an entire Buffer](https://github.com/mercadolibre/flexbuffer-node/blob/1487df393a30872e3e81b246711a4cf6b0b23314/flexbuffer.js#L63) just to write it back out.  It may be possible to make stuff like this zero-copy by switching to using Buffers and/or TypedArrays directly (please correct me if I'm misunderstanding the code).

dantman:
If `arg` is large it may also be more efficient to convert the smaller strings to Buffers, push things onto an array instead of writing them to a flexbuffer, and use Buffer.concat to convert it to a single buffer all at once instead of repeatedly allocating slightly bigger buffers. If you want to super-optimize that, `"\r\n"` and `"$"` could be static buffers that are only allocated once and then shared in all usages (since you won't modify the buffers when making an array of them).

dantman:
Still an important issue. Should not be closed until https://github.com/luin/ioredis/blob/master/package.json#L34 is changed.

[...]

PR #826, opened 2019-03-25
fix: switches from flexbuffer to flexbuffer2

kerihenare:
@kuba-kubula is correct in #821 that we shouldn't use the package without a license. However, linking directly to a git repo has added a new dependency on **git**.

As an alternate solution, I've forked **flexbuffer** and created a modernized drop-in replacement.

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -1,4 +1,4 @@
-import * as fbuffer from 'flexbuffer'
+import * as fbuffer from 'flexbuffer2'
 import * as commands from 'redis-commands'
 import * as calculateSlot from 'cluster-key-slot'
 import asCallback from 'standard-as-callback'
--- a/lib/pipeline.ts
+++ b/lib/pipeline.ts
@@ -278,7 +278,7 @@ Pipeline.prototype.exec = function (callback: CallbackFunction) {
   }).then(execPipeline)
 
   function execPipeline() {
-    let data: FlexBuffer | string = ''
+    let data: FlexBuffer | Buffer | string = ''
     let writePending: number = _this.replyPending = _this._queue.length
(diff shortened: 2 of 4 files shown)


kuba-kubula:
Wow @kerihenare, thanks for that. It looks as the final solution.

kuba-kubula:
Although it's distributed as TS library. Shouldn't it be distributed as both, the `lib` (TS pre-built) and `.ts`

kerihenare:
@kuba-kubula You mean the npm package doesn't include the source files? I've now republished with them.

luin:
Thanks for the pull request! However, given the pr has already changed the major code of the original flexbuffer, and it may potentially introduce inconsistent results to the module interface, I prefer to implementing the related part of flexbuffer in ioredis directly instead of requiring an additional module. What do you think?

kuba-kubula:
@luin Makes sense. @kerihenare has already started the work on the removal of flexbuffer–as–dependency, see https://github.com/luin/ioredis/issues/824#issuecomment-477446881

kerihenare:
@luin **flexbuffer2** is purely a port to ES6 and TS, no logic has really changed. It passes all of the original **flexbuffer** and **ioredis** tests. 

However, I agree it might be better to just reimplement (or port) the functionality that **ioredis** requires. I've been considering that [in this branch](https://github.com/luin/ioredis/compare/master...kerihenare:buffer-refactor?expand=1).

freaz:
@kerihenare Hi, are you planning to continue on integrating `flexbuffer` to `ioredis`?

luin:
Closed via https://github.com/luin/ioredis/pull/856.

PR #845, opened 2019-04-18
Fix logo.svg being cut off

silverwind:
This bothered me quite a bit. See rendered diff.


luin:
Sorry for the late response. I can't see any distinction between these two versions. Could you please explain in detail?

silverwind:
The issue is that the SVG uses font "DIN Condensed" which if not installed on the system will fallback to a system font. I'm on my Mac right now where this font is present and the logo looks right, but on Windows, the logo's text is cut off at `iored`.

The font should actually be converte to a path to render consistenly cross-platform, I'll see if I can do that.

silverwind:
Fix pushed. Logo text is now a `path`. Also ran it throught [svgomg](https://jakearchibald.github.io/svgomg/) for some size optimizations.

silverwind:
Another small fix pushed. Looks like `svgomg` removed `width` and `height`. Added them back in.

luin:
Awesome! That looks cool. Thank you for the pull request. 🍻

PR #858, opened 2019-05-02
Sometimes execResult is undefined.

rimiti:
Hello,

Sometimes, execResult return is undefined. In this case, we just have to return.

_Note: Commit updated._

Regards,

### DIFF
--- a/lib/transaction.ts
+++ b/lib/transaction.ts
@@ -34,6 +34,9 @@ export function addTransactionSupport (redis) {
       const promise = exec.call(pipeline)
       return asCallback(promise.then(function (result) {
         const execResult = result[result.length - 1]
+        if (typeof execResult === 'undefined') {
+          return
+        }
         if (execResult[0]) {
           execResult[0].previousErrors = []
           for (let i = 0; i < result.length - 1; ++i) {
--- a/lib/transaction.ts
+++ b/lib/transaction.ts
@@ -35,7 +35,7 @@ export function addTransactionSupport (redis) {
       return asCallback(promise.then(function (result) {
         const execResult = result[result.length - 1]
         if (typeof execResult === 'undefined') {
-          return
+          throw new Error('Pipeline cannot be used to send any commands when the `exec()` has been called on it.');
         }
         if (execResult[0]) {
           execResult[0].previousErrors = []


rimiti:
@luin Could you please merge this PR? 🙏

luin:
@rimiti Thank you for the pull request. I'm wondering in which case the pipeline will return void?

rimiti:
It occurred when I had a transaction with some del actions. I can’t share you a snipped code because it’s from my client software and the logic is too complicated to extract you an example.

Otherwise the fix doesn’t break the CI. It cant be safely merged. 

Note: if you’re looking for a maintainer, I can help you.

luin:
I just did some tests on my side, but the results seem to always be arrays no matter what the commands were in the transaction. Here's a sample:

```javascript
redis.multi().del('foo1').del('foo2').exec(console.log)
```

The output is:

```
null [ [ null, 1 ], [ null, 1 ] ]
```

Could you create a minimum reproducible sample for this fix?

rimiti:
I've re-run the tests with adding a console.log() when `execResult` is `undefined`:

``` js
if (typeof execResult === 'undefined') {         
    console.log(pipeline);
    return;
}
```

Output:
```bash
tests    |        ...
tests    |        [ Command {
tests    |            name: 'multi',
tests    |            transformed: true,
tests    |            isCustomCommand: false,
tests    |            replyEncoding: 'utf8',
tests    |            errorStack: undefined,
tests    |            args: [],
tests    |            callback: undefined,
tests    |            resolve: [Function],
tests    |            reject: [Function],
tests    |            promise: [Promise] },
tests    |          Command {
tests    |            name: 'srem',
tests    |            transformed: true,
tests    |            isCustomCommand: false,
tests    |            replyEncoding: 'utf8',
tests    |            errorStack: undefined,
tests    |            args: [Array],
tests    |            callback: undefined,
tests    |            resolve: [Function],
tests    |            reject: [Function],
tests    |            promise: [Promise] },
tests    |          Command {
tests    |            name: 'srem',
tests    |            transformed: true,
tests    |            isCustomCommand: false,
tests    |            replyEncoding: 'utf8',
tests    |            errorStack: undefined,
tests    |            args: [Array],
tests    |            callback: undefined,
tests    |            resolve: [Function],
tests    |            reject: [Function],
tests    |            promise: [Promise] },
tests    |          Command {
tests    |            name: 'del',
tests    |            transformed: true,
tests    |            isCustomCommand: false,
tests    |            replyEncoding: 'utf8',
tests    |            errorStack: undefined,
tests    |            args: [Array],
tests    |            callback: undefined,
tests    |            resolve: [Function],
tests    |            reject: [Function],
tests    |            promise: [Promise] },
tests    |          Command {
tests    |            name: 'del',
tests    |            transformed: true,
tests    |            isCustomCommand: false,
tests    |            replyEncoding: 'utf8',
tests    |            errorStack: undefined,
tests    |            args: [Array],
tests    |            callback: undefined,
tests    |            resolve: [Function],
tests    |            reject: [Function],
tests    |            promise: [Promise] },
tests    |          Command {
tests    |            name: 'srem',
tests    |            transformed: true,
tests    |            isCustomCommand: false,
tests    |            replyEncoding: 'utf8',
tests    |            errorStack: undefined,
tests    |            args: [Array],
tests    |            callback: undefined,
tests    |            resolve: [Function],
tests    |            reject: [Function],
tests    |            promise: [Promise] },
tests    |          Command {
tests    |            name: 'srem',
tests    |            transformed: true,
tests    |            isCustomCommand: false,
tests    |            replyEncoding: 'utf8',
tests    |            errorStack: undefined,
tests    |            args: [Array],
tests    |            callback: undefined,
tests    |            resolve: [Function],
tests    |            reject: [Function],
tests    |            promise: [Promise] },
tests    |          Command {
tests    |            name: 'exec',
tests    |            transformed: true,
tests    |            isCustomCommand: false,
tests    |            replyEncoding: null,
tests    |            errorStack: undefined,
tests    |            args: [],
tests    |            callback: undefined,
tests    |            resolve: [Function],
tests    |            reject: [Function],
tests    |            promise: [Promise] },
tests    |          Command {
tests    |            name: 'del',
tests    |            transformed: true,
tests    |            isCustomCommand: false,
tests    |            replyEncoding: 'utf8',
tests    |            errorStack: undefined,
tests    |            args: [Array],
tests    |            callback: undefined,
tests    |            resolve: [Function],
tests    |            reject: [Function],
tests    |            promise: [Promise] },
tests    |          Command {
tests    |            name: 'del',
tests    |            transformed: true,
tests    |            isCustomCommand: false,
tests    |            replyEncoding: 'utf8',
tests    |            errorStack: undefined,
tests    |            args: [Array],
tests    |            callback: undefined,
tests    |            resolve: [Function],
tests    |            reject: [Function],
tests    |            promise: [Promise] } ],
tests    |       _result:
tests    |        [ [ null, 'OK' ],
tests    |          [ null, 'QUEUED' ],
tests    |          [ null, 'QUEUED' ],
tests    |          [ null, 'QUEUED' ],
tests    |          [ null, 'QUEUED' ],
tests    |          [ null, 'QUEUED' ],
tests    |          [ null, 'QUEUED' ],
tests    |          [ null, [Array] ],
tests    |          undefined,
tests    |          undefined ],
tests    |       _transactions: 0,
tests    |       _shaToScript: {},
tests    |       resolve: [Function],
tests    |       reject: [Function],
tests    |       promise:
tests    |        Promise {
tests    |          [ [Array],
tests    |            [Array],
tests    |            [Array],
tests    |            [Array],
tests    |            [Array],
tests    |            [Array],
tests    |            [Array],
tests    |            [Array],
tests    |            undefined,
tests    |            undefined ] },
tests    |       exec: [Function],
tests    |       execBuffer: [Function],
tests    |       nodeifiedPromise: true,
tests    |       replyPending: 0 }".
tests    |
tests    |       at BufferedConsole.log (node_modules/@jest/console/build/BufferedConsole.js:199:10)
tests    |       at node_modules/ioredis/built/transaction.js:47:25
```

I hope it can help you.

luin:
Thanks for the logs. I can finally reproduce the issue. The key point is to send other commands after `exec()`:

```javascript
const pipeline = redis.multi().del('foo1')
pipeline.exec(console.log) // TypeError: Cannot read property '0' of undefined
pipeline.del('foo3') // pipeline shouldn't be used to send any commands when the `exec()` has been called on it.
```

I think we should catch the exception (like this pr do) but re-throw it as a more meaningful error with a message like "Transaction resolved failed. Did you call any commands in a transaction after `exec()` had been called on that transaction?"

What do you think?

rimiti:
> I think we should catch the exception (like this pr do) but re-throw it as a more meaningful error with a message like "Transaction resolved failed. Did you call any commands in a transaction after exec() had been called on that transaction?"

It think you're right, it's the better way to fix this case 👍

rimiti:
> I think we should catch the exception (like this pr do) but re-throw it as a more meaningful error with a message like "Transaction resolved failed. Did you call any commands in a transaction after exec() had been called on that transaction?"

It think there are two solution:
- First: Throwing an error when we are in this case
- Second: Adding commands sent after exec() into the pipeline array.

I prefer the second solution because, it always works.

luin:
> * First: Throwing an error when we are in this case.

This is the quickest solution for this issue.

> * Second: Adding commands sent after exec() into the pipeline array.

This breaks BC since in the past commands invoked after exec() will not be sent to Redis. So even we chose this way, it have to be happened in the next major release.

Another cons I can think of is allowing commands after exec() makes things obscure: is these command will be included in the transaction, or be excluded from the transaction?

rimiti:
> Another cons I can think of is allowing commands after exec() makes things obscure: Will these command be included in the transaction, or be excluded from the transaction?

I think, if it's obscure, it's because this is not the right way to do that. After reflection, it'll be more clear to just throw an error instead of trying to bypass transaction logic.

What disturbs me, if I run the same tests with redis package, it's works and without error.
It may be interesting to check how to the redis package treat this case ?

luin:
Just tried node_redis. The following code：

```javascript
const m = r.multi().del('foo1')
m.exec(console.log)
m.del('foo2')
```

Both node_redis & ioredis will simply ignore `del(foo2)`. node_redis returns `[0, undefined]` corresponding to del(foo1) and del(foo2). The result of del(foo2) being undefined is because the command was not sent to Redis.

I think in this case, simply throwing errors to let developers know that there are commands not being sent would be a fair solution.

rimiti:
> Both node_redis & ioredis will simply ignore del(foo2). node_redis returns [0, undefined] corresponding to del(foo1) and del(foo2). The result of del(foo2) being undefined is because the command was not sent to Redis.

Oh, it's so a silent error... 😕
Totally agreed with you, throwing an error is best solution.

rimiti:
@luin Do you want I throw an error like that instead of the return?

#### Before

```javascript
        if (typeof execResult === 'undefined') {
          return
        }
```

#### After

```javascript
        if (typeof execResult === 'undefined') {
          throw new Error('Pipeline cannot be used to send any commands when the `exec()` has been called on it.');
        }
```

What do you think?

luin:
Looks good to me!

rimiti:
It can be merged. 🎉

luin:
@rimiti The new version has been published automatically when merged. I just add the 4.9.3 to the latest tag.

Issue #536, opened 2017-10-25
Chaining custom commands in cluster mode


manast:
According to the documentation "Chaining custom commands in the pipeline is not supported in Cluster mode.".
I am wondering, is this a limitation of ioredis or of redis itself? and do you have any information or link on why this limitation exists?

manast:
@luin please, could you be so kind to share any insight you have in this issue?

luin:
Sorry for the late response.

That's not a limitation of Redis. It's just not easy to support chaining custom commands in the cluster setup: ioredis always try `evalsha` first when sending custom commands, and when the command fails, `eval` will be sent instead. However, when it comes to pipelines, to make sure all commands in a pipeline will be received by Redis in order, ioredis sends `script exists` command to check if all lua scripts have loaded in the target redis node and will load the scripts which haven't loaded before sending the pipeline.

In the cluster setup, the nodes may change (e.g. failover happens) during the process of loading scripts, so it's hard to make sure that the scripts are loaded to the correct redis node. That's why this feature haven't been implemented.

manast:
thanks for the throughout explanation.  When calling ```defineCommand```, why can't we just call ```SCRIPT LOAD``` and then after that we could just assume that the command is available?

luin:
@manast Script cache may be cleared by the `SCRIPT FLUSH` command.  Despite that, calling `SCRIPT LOAD` when defining commands indeed solves the problem (we also need to resend all `SCRIPT LOAD` again when new connection has been made).

manast:
Bump! :)

btd:
Bump

manast:
Bump!

btd:
Bump

manast:
Bump because it is an important issue!

Rush:
Bump - interested in it as well.

Gappa88:
interested

firedfox:
Bump because we are using more Redis clusters in recent projects. This problem is getting more serious.

ajwootto:
Also running into this.

btzsoft:
we have the same issue.

botzill:
Any updates on this issue? @luin 

Thx.

jl9404:
> @manast Script cache may be cleared by the `SCRIPT FLUSH` command. Despite that, calling `SCRIPT LOAD` when defining commands indeed solves the problem (we also need to resend all `SCRIPT LOAD` again when a new connection has been made). Good idea!

[...]

PR #876, opened 2019-05-23
Chaining custom commands in cluster mode

r3nat:
@luin probably a fix for #536, please take a look

### DIFF
--- a/lib/script.ts
+++ b/lib/script.ts
@@ -23,6 +23,20 @@ export default class Script {
       options.keyPrefix = this.keyPrefix
     }
 
+    // https://github.com/luin/ioredis/issues/536#issuecomment-346872649
+    if (container.constructor.name === 'Pipeline' && container.isCluster) {
+      const evalCmd = new Command('eval', [this.lua].concat(args), options);
+
+      const result = container.sendCommand(evalCmd);
+      if (isPromise(result)) {
+        return asCallback(result, callback);
+      }
+
+      asCallback(evalCmd.promise, callback);
+
+      return result;
+    }
+
     const evalsha = new Command('evalsha', [this.sha].concat(args), options)
     evalsha.isCustomCommand = true
 


luin:
It seems the scripts are sent with `eval` directly, without bandwidth optimization with `evalsha`?

r3nat:
Yep, but it seems to work.

r3nat:
@luin so what do you think? Is it good or bad? Should I change something or add some tests?

luin:
@r3nat Thank you for the pull request, but I think there are some issues still under discussion.

Above all, sending commands within a pipeline is aimed at improving the network performance. However, if we send custom commands with `eval` instead of `evalsha` just in order to make custom commands works in the cluster pipeline, we are having some nontrivial performance tradeoff, which I think is not worth, especially given Node.js send commands in a way pretty similar to pipeline (not waiting for the previous command being handled before sending the next command).

r3nat:
@luin 

>Above all, sending commands within a pipeline is aimed at improving the network performance.

Pipelines are indeed aimed on that, but transactions are all about atomicity.

Why it is possible to run `eval` inside transaction, but not possible to do literally same thing with custom commands? It should fallback to `eval`'s if `evalsha` optimization could not be applied.

BTW I've run [benchmarks](https://gist.github.com/r3nat/ded8cf587854388cec1acc0aeab19ebc). `evalsha` optimization is slightly winning in cases of really big scripts (32kb) or big number of calls (100) inside single pipeline. In other cases it is faster to just run `eval`'s. And I've tested with redis that run locally. If redis is on another machine (adding 0.1-0.5ms of rtt) then bare `eval`'s may have become better in all cases.

After all I think custom commands is a great way of thinking about scripts - they become simple redis commands.

r3nat:
@luin please take a look on message above

luin:
Thank you for the benchmark script. It makes a lot of sense.

I updated the script, added cases that sending commands without a pipeline (https://gist.github.com/luin/c8fd968f50c0a7abcdc79f0947274729). According to the results, not only should us send commands in a pipeline with `eval` in cluster mode, but we also may do that when connecting to a standalone Redis server for a better performance. However, as the result suggested, it seems better to send lua scripts without pipeline in terms of performance. So only atomicity makes it worthwhile.

For this pull request, I'm wondering if there's a better way to deal with constructor detection since there's an issue related to `constructor.name` approach. I also think it may be better to have an option like `useEvalsha: false` to tell Script use `eval` directly to make things more explicit. What do you think? Anyway, tests are required for these changes :-).

RPallas92:
Any update on this ?

Issue #883, opened 2019-06-07
Cluster UnhandledPromiseRejectionWarning


jstewmon:
I'm intentionally causing a connection error (by omitting a NAT mapping for one cluster node) to verify my application's error handling, but I'm seeing an UnhandledPromiseRejectionWarning:

```
(node:18) UnhandledPromiseRejectionWarning: Error: Connection is closed.
    at close (/home/node/app/node_modules/ioredis/built/redis/event_handler.js:102:25)
    at Socket.<anonymous> (/home/node/app/node_modules/ioredis/built/redis/event_handler.js:73:20)
    at Object.onceWrapper (events.js:288:20)
    at Socket.emit (events.js:200:13)
    at Socket.EventEmitter.emit (domain.js:471:20)
    at TCP.<anonymous> (net.js:586:12)
```

My code is executing a multi command like this:
```js
redisClient
      .multi()
      .expire(key, this.expire)
      .hgetall(key)
      .exec()
      .then(
        async ([[expErr, expVal], [hashErr, hashVal]]) => {
          if (expErr || hashErr) {
            throw expErr || hashErr;
          }
          return Object.values(hashVal);
        },
      )
      .then(x => cb(null, x))
      .catch(err => cb(err, []));
```

I believe this is the relevant debug output preceding the throw:
```
  ioredis:redis status[127.0.0.1:7000]: [empty] -> wait +5ms
  ioredis:delayqueue send 1 commands in failover queue +119ms
  ioredis:redis status[127.0.0.1:7000]: wait -> connecting +1ms
  ioredis:redis queue command[127.0.0.1:7000]: 0 -> multi([]) +1ms
  ioredis:redis queue command[127.0.0.1:7000]: 0 -> expire([ 'mykey', '2592000' ]) +0ms
  ioredis:redis queue command[127.0.0.1:7000]: 0 -> hgetall([ 'mykey' ]) +0ms
  ioredis:redis queue command[127.0.0.1:7000]: 0 -> exec([]) +1ms
  ioredis:connection error: Error: connect ECONNREFUSED 127.0.0.1:7000
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1054:14) {
  errno: 'ECONNREFUSED',
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 7000
} +9ms
  ioredis:redis status[127.0.0.1:7000]: connecting -> close +2ms
[...]

PR #884, opened 2019-06-08
fix: pipeline with transactions causes unhandled warnings

luin:
`Pipeline#exec()` is a little tricky: it has an init process when called first time (indicated by whether `Pipeline#nodeifiedPromise` is true). `Transaction#exec()` is similar to that, it also has an init process that parse the result from `Pipeline#exec()` before returning to users. However, `#nodeifiedPromise` is not checked here, so it may fail which causes the unhandled warnings when retries.

Closes: #883.

To be frank, it also take me a lot of time in order to understand the code, especially the complex relation between Pipeline & Transaction. We indeed need to do a refactor on this.

### DIFF
--- a/lib/transaction.ts
+++ b/lib/transaction.ts
@@ -31,6 +31,12 @@ export function addTransactionSupport (redis) {
       if (this._transactions > 0) {
         exec.call(pipeline)
       }
+
+      // Returns directly when the pipeline
+      // has been called multiple times (retries).
+      if (this.nodeifiedPromise) {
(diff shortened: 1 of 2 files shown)


jstewmon:
The patch resolves the issue with my reproduction setup, but the new test passes with and without the change.

luin:
@jstewmon You're right. Just fixed.

jstewmon:
LGTM! Thanks for addressing so quickly!

Issue #896, opened 2019-06-13
[ioredis] Unhandled error event


barakpx:
ioredis cluster client will print out connection issues even if I register to error handler.

`console.error('[ioredis] Unhandled error event:', error.stack);`

Even when your cluster redis instance registered for 'error' or 'node error'.
Look like getInfoFromNode duplicates Redis instance, the current duplicate implementation does not close the event listeners. Any error on the cloned instance will fall to the console.log.

This is an issue as when there are connection problems the process prints out to the console and doesn't trigger the event on the cluster instance.

kjr247:
Is this a new bug? Can I use an earlier version? What needs to be done?

barakpx:
Not sure it is new - but when you clone the listeners when you duplicate the Redis instance I think we will get the right behavior.

barakpx:
Created PR https://github.com/luin/ioredis/pull/899

Kristoffer95:
I am having this error
`
Using ts-node version 8.10.1, typescript version 3.8.3
http://localhost:3000
[ioredis] Unhandled error event: Error: write EPIPE
    at WriteWrap.afterWrite (net.js:789:14)
[ioredis] Unhandled error event: Error: write EPIPE
    at WriteWrap.afterWrite (net.js:789:14)
[ioredis] Unhandled error event: Error: write EPIPE
    at WriteWrap.afterWrite (net.js:789:14)
[ioredis] Unhandled error event: Error: write EPIPE
    at WriteWrap.afterWrite (net.js:789:14)
[ioredis] Unhandled error event: Error: read ECONNRESET
    at TCP.onStreamRead (internal/stream_base_commons.js:111:27)
[ioredis] Unhandled error event: Error: read ECONNRESET
    at TCP.onStreamRead (internal/stream_base_commons.js:111:27)
[ioredis] Unhandled error event: Error: write EPIPE
    at WriteWrap.afterWrite (net.js:789:14)
[ioredis] Unhandled error event: Error: read ECONNRESET
    at TCP.onStreamRead (internal/stream_base_commons.js:111:27)
[ioredis] Unhandled error event: Error: read ECONNRESET
    at TCP.onStreamRead (internal/stream_base_commons.js:111:27)
[ioredis] Unhandled error event: Error: read ECONNRESET
    at TCP.onStreamRead (internal/stream_base_commons.js:111:27)
[ioredis] Unhandled error event: Error: read ECONNRESET
[...]

PR #899, opened 2019-06-16
redis duplicate resubscribe event handlers (issue #896)

barakpx:


luin:
Sorry for the late response. Issue #896 has been fixed in 4.13.0 in a more targeted way. Thanks for your contribution!

PR #906, opened 2019-06-23
Custom connectors and additional typings

### LINKED ISSUE #903: Async Sentinel/Cluster Addresses
The Redis service I use provides the host and ports over an http request, and the agents IPs might shift over time. Is there a good way to embed this in the client?

So far I have written a kinda hacky wrapper:

```js
class FooConnector extends SentinelConnector {
  constructor(options, client) {
    super(options);
    this._client = client;
  }

  async connect(callback, eventEmitter) {
    try {
      const connector = await this.discover();
      connector.connect(callback, eventEmitter);
      this._client._discoverCalled = true;
      this._client.connector = connector;
    } catch (e) {
      callback(e.message);
    }
  }

  async discover() {
    this.options.sentinels = await service.getSentinelEndpoints(this.options.fetchUri);

    return new SentinelConnector(this.options);
  }
}

class FooRedis extends IORedis {
  constructor(
    cacheName,
    fetchUri = 'http://foo.com/redis/',
    options = {}
  ) {
    super({
      name: cacheName,
      fetchUri,
      sentinels: [{host: 'localhost', port: 6379}], // dummy value
      ...options
    });
  }

  connect(callback) {
    if (!this._discoverCalled)
      this.connector = new FooConnector(this.options, this);

    return IORedis.prototype.connect.call(this, callback);
  }
}
```

imatlopez:
Made the SentinelIterator and actual Iterator ~and made it not affect `this.options.sentinels` in place~ (tests were there to test the side effect).

Added `floatingSentinels` to the `SentinelConnectorOptions` which is called whenever `connect` is called, it must return an array of sentinels using a callback.

Welcome any feedback :)

resolves #903 for Sentinel at least :)


luin:
Thanks for the pull request!

It adds some overheads to add an option for this specified issue. A better way seems to be customizing SentinelConnector by overriding `SentinelConnector#connect()` to fetch dynamic endpoints first.

Currently, it could be a little bit hacky since there's no way for users to provide customized connectors to `Redis`. But it trivial to add an option to `Redis` like `connector` which will solve a broader set of problems.

imatlopez:
I added the `connector` property to the config, which is now where `Redis` uses its connector from. Hence if no property is set, the constructor will assign it based on the old behavior.

I reverted some of the changes to `SentinelConnector` but not all because I felt it is cleaner this way (?)

Added `AsyncSentinelConnector` as  an example of a custom connector that can be passed in.

imatlopez:
@luin let me know your thoughts :)

lib/connectors/SentinelConnector/index.ts
```diff
@@ -37,15 +41,19 @@ export interface ISentinelConnectionOptions extends ITcpConnectionOptions {
   updateSentinels?: boolean
 }
 
+export interface ISentinelConnectorOptions extends ISentinelConnectionOptions {
+  sentinels: Partial<ISentinelAddress>[]
+}
```
luin: It seems `ISentinelConnectionOptions` isn't used other places inside the library except for `AsyncSentinelConnector`, which I think can be moved outside (like to the `examples` folder).

If people want to reuse `ISentinelConnectorOptions` with the `sentinels` property to be omitted, they can simply use `Pick<ISentinelConnectorOptions, Exclude<keyof TypeA, 'sentinels'>>`, so given that, I think we just need to keep one of the two interfaces.

lib/connectors/AsyncSentinelConnector.ts
```diff
@@ -0,0 +1,35 @@
...
+    private fetch: AsyncSentinelFetch
+
+    constructor(options: ISentinelConnectionOptions, fetch: AsyncSentinelFetch) {
+        options.sentinels = options.sentinels || [{host: 'localhost', port: 6379}] // Placeholder
+        super(options as ISentinelConnectorOptions)
+
+        this.fetch = fetch;
+    }
+
+    public connect(callback: NodeCallback<NetStream>, eventEmitter: ErrorEmitter) {
```
luin: I prefer to making all public APIs use Promise instead of the outdated callback-style. So if we are going to export the connectors, the `connect` method needs to be changed to `public connect(eventEmitter: ErrorEmitter): Promise<NetStream>`.

lib/connectors/AsyncSentinelConnector.ts
```diff
@@ -0,0 +1,35 @@
...
+import { NetStream } from '../types';
+import { NodeCallback } from './types';
+
+export type AsyncSentinelFetch = () => Promise<Partial<ISentinelAddress>[]>
+
+export default class AsyncSentinelConnector extends SentinelConnector {
+    private fetch: AsyncSentinelFetch
+
+    constructor(options: ISentinelConnectionOptions, fetch: AsyncSentinelFetch) {
+        options.sentinels = options.sentinels || [{host: 'localhost', port: 6379}] // Placeholder
```
luin: What about not calling `super()` here (so that we don't need a placeholder)?

luin:
Awesome work! As you said, that changes made on `SentinelConnector` make the code cleaner. The most parts of codes are great, and I've left some comments about the interface of `SentinelConnector` since this the first time we export it.

lib/connectors/AsyncSentinelConnector.ts
imatlopez: you always have to call super though

lib/connectors/AsyncSentinelConnector.ts
```diff
@@ -0,0 +1,35 @@
...
+        return this.fetch().then(sentinels => {
+            if (!sentinels.length) throw new Error(EMPTY_SENTINELS_MSG);
+
+            this.options.sentinels = sentinels;
+            this.sentinelIterator = new SentinelIterator(sentinels)
+            return super.connect(eventEmitter);
+        });
+    }
+
+}
```
luin: I suggest moving this file to the `examples` folder since most users don't need it, and it trivial to implement one. What do you think?

imatlopez: 👍 now I just need to find why the tests fail

.travis.yml
```diff
@@ -12,6 +12,7 @@ services:
 - redis-server
 
 script:
+- npm run build
```
luin: Why `npm run build` is needed here?

imatlopez: test may have type errors like trying to access private properties, but the code itself is fine, so this guards against build time errors while the tests are for logical errors only.

imatlopez:
I think I dug myself a deep hole trying to change the interface to promises. I haven't figured out how to run the tests locally either so I'm gonna need some assistance 😁

imatlopez:
@luin I've noticed that the ESLint doesn't quite run in some parts of the code, fixing it by adding `typescript-eslint` added lots of conflicts. Once this goes in, would you like another PR adding it for code consistency? I personally like `Prettier` for code styling too.

luin:
@imatlopez That would be great! And Prettier is also my favorite :-)

PR #907, opened 2019-06-25
Run prettier

imatlopez:

PR #909, opened 2019-06-25
Use connector as class not value

imatlopez:
Fixes https://github.com/luin/ioredis/pull/907#issuecomment-505545220

Realized that options was being passed through from an internal object, not the one I passed in, hence internal instantiation is needed.


imatlopez:
@luin found the bug :)

Issue #921, opened 2019-07-08
Support .duplicate() on Redis.Cluster


SGrondin:
Hi,

[This issue](https://github.com/SGrondin/bottleneck/issues/129) brought to my attention that `client.duplicate()` does not exist if the client was created with `new Redis.Cluster({ ... })`.

It seems trivial to add and I don't mind making a PR for it, but I'd like to know if it's intentional or an oversight before spending time on this!

Thank you

SGrondin:
This was added in #926 and released in v4.12.0

PR #926, opened 2019-07-14
feat: add cluster.duplicate() method

SGrondin:
Hi,

See https://github.com/luin/ioredis/issues/921. This PR adds the missing `cluster.duplicate()` method.

Notes:
- I'm using `this.startupNodes.slice(0)` to ensure we pass a copy and not a reference. This matches the behavior of the options, which also get a copy thanks to `Object.assign()`.
- I'm not sure this new test I added is being executed when running `npm test`. I don't see it in the mocha output. It works as expected when I run ` npx mocha test/functional/cluster/duplicate.ts`, could someone please validate?

### DIFF
--- a/lib/cluster/index.ts
+++ b/lib/cluster/index.ts
@@ -386,6 +386,29 @@ class Cluster extends EventEmitter {
   }
 
   /**
+   * Create a new instance with the same startup nodes and options as the current one.
+   *
+   * @example
+   * ```js
+   * var cluster = new Redis.Cluster([{ host: "127.0.0.1", port: "30001" }]);
+   * var anotherCluster = cluster.duplicate();
+   * ```
+   *
+   * @public
+   * @param {(Array<string | number | object>)} [overrideStartupNodes=[]]
+   * @param {IClusterOptions} [overrideOptions={}]
+   * @memberof Cluster
+   */
+  public duplicate(overrideStartupNodes = [], overrideOptions = {}) {
+    const startupNodes =
+      overrideStartupNodes.length > 0
+        ? overrideStartupNodes
+        : this.startupNodes.slice(0);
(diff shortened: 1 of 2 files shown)


luin:
LGTM. The test should be executed since `npm test` simply calls `mocha test/**/*.ts`. "test/functional/cluster/duplicate.ts" matches the pattern. Does `npx mocha test/**/*.ts` execute the test?

SGrondin:
Wow, that's a quick response!

Unfortunately `npx mocha test/**/*.ts` fails to run any tests because of a compilation error in `test/functional/scan_stream.ts`. I'm guessing it's also why `npm test` isn't running the new test.

```
TSError: ⨯ Unable to compile TypeScript:
test/functional/scan_stream.ts:239:15 - error TS2339: Property 'sadd' does not exist on type 'Cluster'.
test/functional/scan_stream.ts:240:30 - error TS2339: Property 'sscanStream' does not exist on type 'Cluster'.
```

luin:
I see. There are some issues related to type which will be fixed in the next major version. Just pushed a temp fix. Please merge the latest commit and try again.

SGrondin:
I've rebased on top of the latest commit and fixed the code formatting issues. The build is passing but I still don't see the new test. It should be under a `describe` grouped called `#duplicate`. All I see is the other duplicate test, for the base client.

luin:
It seems that mocha ignores tests that contain TypeScript errors, which is the reason the newly added test wasn't executed, as well as some other tests. Didn't notice this issue before.

luin:
@SGrondin Just did some fixes for the issues. Merge the latest commits will do the trick.

SGrondin:
Very nice. It ran 55 tests that were previously ignored.

PR #939, opened 2019-07-30
Supports rediss:// URL

### LINKED ISSUE #846: Allow a way to configure TLS using DSN
There is currently no way to configure TLS using DSN.

For starters, couldn't this requirement be automatically derived based on the `rediss://` prefix?

monkbroc:
Support [`rediss://` URL](https://www.iana.org/assignments/uri-schemes/prov/rediss) (Redis over TLS) by defaulting option `tls: true` for those URLs.

Fixes #846

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ new Redis({
 });
 ```
 
-You can also specify connection options as a [`redis://` URL](http://www.iana.org/assignments/uri-schemes/prov/redis):
+You can also specify connection options as a [`redis://` URL](http://www.iana.org/assignments/uri-schemes/prov/redis) or [`rediss://` URL](https://www.iana.org/assignments/uri-schemes/prov/rediss) when using [TLS encryption](#tls-options):
 
 ```javascript
 // Connect to 127.0.0.1:6380, db 4, using password "authpassword":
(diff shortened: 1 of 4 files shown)


luin:

monkbroc:
Thanks for the quick merge and release! :tada:

paulmelnikow:
Hi! Unfortunately this change prevents passing in `tls` options, as the custom options are now replaced by the boolean `true`. Probably this should set `tls: true` only when `tls` is not already present in the options.

Perhaps more to the point, the reason that I need to pass in tls options is so the correct SSL certificate is used; otherwise I get this error:

```
Error: unable to verify the first certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1049:34)
    at TLSSocket.emit (events.js:182:13)
    at TLSSocket.EventEmitter.emit (domain.js:442:20)
    at TLSSocket._finishInit (_tls_wrap.js:631:8) code: 'UNABLE_TO_VERIFY_LEAF_SIGNATURE'
```

This is what my code looks like:

```js
    const options =
      this.url && this.url.startsWith('rediss:')
        ? {
            //  https://www.compose.com/articles/ssl-connections-arrive-for-redis-on-compose/
            tls: { servername: new URL(this.url).hostname },
          }
        : undefined
    this.redis = new Redis(this.url, options)
```

Would you be open to a fix that passes this value automatically? I think it will allow for more robust out-of-the-box behavior.

I'm not sure what it is about the way Compose is set up that requires this, though I think the issue is that an SSL connection to that IP address returns more than one certificate (and without it, TLS tries to validate the wrong one).

paulmelnikow:
Opened new issues for these: #947 #948

Issue #940, opened 2019-08-05
previously working rediss:// url does not work since 4.14.0


mkls:
Hello,

In 4.14.0 support for rediss:// url was introduced, but trying to connect to an url like this we get:
```
[ioredis] Unhandled error event: Error: unable to verify the first certificate
```
errors.

Before this version we used to have a workaround in our code for connecting to rediss:// urls which looked something like this:
```
const { URL } = require('url');
const redisUri = 'rediss://...'
const options = {
  lazyConnect: true,
  retryStrategy: () => 1000
};
if (redisUri.startsWith('rediss://')) {
  options.tls = { servername: new URL(redisUri).hostname };
}
const connection = new Redis(redisUri, options)
```
which was working fine. (even in local development where we had a regular redis:// url)

With this newly introduced feature I haven't been able to find the right options to make this work.

It looks like the new parser set's `tls: true` in options, and we would need more detailed config here.

luin:
Just submitted a pull request for this issue.

For now, the workaround would be changing the protocol to `redis://` after assigning `tls`:

```javascript
if (redisUri.startsWith('rediss://')) {
  options.tls = { servername: new URL(redisUri).hostname };
  redisUri = redisUri.replace('rediss://', 'redis://')
}
```

After the pull request being merged, you can parse the tls option via URL directly:

```javascript
const redisUri = 'rediss://127.0.0.1/?tls[servername]=127.0.0.1'
const options = {
  lazyConnect: true,
  retryStrategy: () => 1000
};

const connection = new Redis(redisUri, options)
```

PR #942, opened 2019-08-08
feat: support nesting options via redis://

luin:
Close: #940


paulmelnikow:
I'm having the same issue, which I mistakenly reported at #947 before I noticed this PR.

This seems like a more complicated solution than necessary. Wouldn't it be sufficient to conditionally set `tls` to true if it's not already set?

PR #950, opened 2019-08-20
feat: Automatically set tls.servername option with rediss:// URLs

### LINKED ISSUE #940: previously working rediss:// url does not work since 4.14.0

### LINKED ISSUE #948: When using rediss:/ URLs, auto-configure `tls.servername`
As mentioned at https://github.com/luin/ioredis/pull/939#issuecomment-523074291, using `ioredis` (or `node-redis`) with Compose.io's Let's Encrypt certificates generates an SSL error:

```
Error: unable to verify the first certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1049:34)
    at TLSSocket.emit (events.js:182:13)
    at TLSSocket.EventEmitter.emit (domain.js:442:20)
    at TLSSocket._finishInit (_tls_wrap.js:631:8) code: 'UNABLE_TO_VERIFY_LEAF_SIGNATURE'
```

Disabling strict SSL is a workaround, though the correct solution is to [pass the `servername` option to TLS.connect](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback), which tells the client which TLS certificate to use.

I found the solution in this article: https://www.compose.com/articles/ssl-connections-arrive-for-redis-on-compose/

It doesn't do an awesome job of explaining the cause, though I think the reason is that there are more than one certificates being used on the same IP address.

My code looks like this:

```js
const options =
  this.url && this.url.startsWith('rediss:')
    ? {
        //  https://www.compose.com/articles/ssl-connections-arrive-for-redis-on-compose/
        tls: { servername: new URL(this.url).hostname },
      }
    : undefined
this.redis = new Redis(this.url, options)
```

I'd be happy to open a PR if this default behavior would be welcome!

paulmelnikow:
Fix #948
Fix #940
Alternative to #949 and #942

### DIFF
--- a/lib/redis/index.ts
+++ b/lib/redis/index.ts
@@ -192,6 +192,7 @@ Redis.prototype.resetOfflineQueue = function() {
 
 Redis.prototype.parseOptions = function() {
   this.options = {};
+  let isTls = false;
   for (var i = 0; i < arguments.length; ++i) {
     var arg = arguments[i];
     if (arg === null || typeof arg === "undefined") {
@@ -201,11 +202,17 @@ Redis.prototype.parseOptions = function() {
       defaults(this.options, arg);
     } else if (typeof arg === "string") {
       defaults(this.options, parseURL(arg));
+      if (arg.startsWith('rediss://')) {
+        isTls = true
+      }
     } else if (typeof arg === "number") {
       this.options.port = arg;
     } else {
       throw new Error("Invalid argument " + arg);
     }
+  }
+  if (isTls) {
+    defaults(this.options, {tls: true})
   }
   defaults(this.options, Redis.defaultOptions);
 
--- a/lib/redis/index.ts
+++ b/lib/redis/index.ts
@@ -202,8 +202,8 @@ Redis.prototype.parseOptions = function() {
       defaults(this.options, arg);
     } else if (typeof arg === "string") {
       defaults(this.options, parseURL(arg));
-      if (arg.startsWith('rediss://')) {
-        isTls = true
+      if (arg.startsWith("rediss://")) {
+        isTls = true;
       }
     } else if (typeof arg === "number") {
       this.options.port = arg;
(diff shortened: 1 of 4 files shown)


paulmelnikow:
@luin Would you mind taking a look at this? It would be great to get this merged as Shields is currently blocked from upgrading. Thanks a lot!

lib/redis/index.ts
```diff
@@ -201,12 +202,18 @@ Redis.prototype.parseOptions = function() {
...
+        isTls = true;
+      }
     } else if (typeof arg === "number") {
       this.options.port = arg;
     } else {
       throw new Error("Invalid argument " + arg);
     }
   }
+  if (isTls) {
+    defaults(this.options, { tls: { servername: this.options.host } });
```
luin: I'm not sure if setting `tls` to `this.options.host` instead of `true` applies to more scenarios, but I think at least we need to check whether `this.options.host` is a hostname instead of an IP address.

As for me, I prefer the way that we just default `tls` options to `true` without setting any other properties (like `servername`) and let users handle it. The improvement we should make here is to keep the options provided by users not be overwritten.

paulmelnikow: > I'm not sure if setting `tls` to `this.options.host` instead of `true` applies to more scenarios, but I think at least we need to check whether `this.options.host` is a hostname instead of an IP address.

I'm happy to make that modification if you'd like.

Setting `tls.options.host` to the hostname of the server is necessary to use many shared Redis hosting services, including Compose.io and Redis Labs. It reduces boilerplate and seems like a reasonable default as this is not really a fringe use case, but a common scenario for TLS.

> The improvement we should make here is to keep the options provided by users not be overwritten.

Alternatively, #949 does just that!

PR #961, opened 2019-09-10
Bugfix for ssl handshake + socket hangup errors

pwasurion:
Changing the `once()` to `on()` to prevent the first error from removing the listener and causing the second event to be an unhandled error event.


luin:
Thanks for your contribution! Could you provide a test case for this?

PR #962, opened 2019-09-12
Bugfix for ssl handshake + socket hangup errors

pwasurion:
* changing the `once()` to `on()` to prevent the first error from removing the listener and causing the second event to be an unhandled error event.
* Added unit test for multiple errors emitting from socket

### DIFF
--- a/lib/redis/index.ts
+++ b/lib/redis/index.ts
@@ -305,7 +305,7 @@ Redis.prototype.connect = function(callback) {
         }
 
         stream.once(CONNECT_EVENT, eventHandler.connectHandler(_this));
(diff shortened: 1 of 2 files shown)


knoxcard:
@luin - how serious is this bug?

pwasurion:
It is critical for connecting to an encrypted redis cluster in AWS. Without the fix an unhandled error event is thrown which brings the Lambda execution down.

ricardocasares:
This is critical for our projecst as well. @knoxcard will you consider to merging this? I can contribute if something is missing.

knoxcard2:
@luin - when you get the chance, please review. We need you bud!

Issue #968, opened 2019-09-19
Question: Is there a way to log commands being sent to redis?


ThisIsMissEm:
I know I could achieve this by wrapping every method, but ideally, I don't want to do that. Essentially, in certain conditions, like development, I'm wanting to have a function called to log every command being sent from node.js to Redis. Is there a way to archive this?

ThisIsMissEm:
@tuananh okay, it looks like the typedefs for sendCommand are incorrect, I can see it used with two signatures:

```
sendCommand(command: IORedis.Command)
sendCommand(command: IORedis.Command, stream: ?, node: ?)
```

I'm not sure what `stream` and `node` are above, as there's no type information in the code for them

ThisIsMissEm:
@tuananh any thoughts on how we should go about fixing the incorrect type definitions for sendCommand?

tuananh:
> @tuananh any thoughts on how we should go about fixing the incorrect type definitions for sendCommand?

I'm not sure. types is located in another repo, not sure if anyone is maintaining it

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ioredis/index.d.ts

ThisIsMissEm:
@tuananh oh? That's odd for a project that's written in typescript?

ThisIsMissEm:
I guess as a temporary thing I can patch the @types/ioredis package to fix it.

tuananh:
> I guess as a temporary thing I can patch the @types/ioredis package to fix it.

i guess because at first @luin hasn't finished migrating everything to ts so he needs to add an options that `allowJs` in tsc. but using this option, `tsc` is unable to produce `d.ts` files.

now that ioredis is fully migrated to TS, i think we can generate `d.ts` files and publish to npm so that we dont need to maintain types.

PR #981, opened 2019-10-01
fix: generate declaration files d.ts

tuananh:
Related #968 

Now that ioredis is in TypeScript, we can set `allowJs` to false, let `tsc` generates declaration files `d.ts` and publish it to npm.

after:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "lint": "eslint --ext .js,.ts .",
     "format": "prettier --write \"{,!(node_modules)/**/}*.{js,ts}\"",
     "format-check": "prettier --check \"{,!(node_modules)/**/}*.{js,ts}\"",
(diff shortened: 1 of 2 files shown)


luin:
Thanks for the pull request! Declarations are not generated and published to npm is because there are many methods (like `redis.get()` / `redis.mget()`) are generated dynamically instead of being defined statically. So the automatically generated declarations will not contain those methods, which, I'm afraid, may mislead many developers.

Currently, they may keep using @types/ioredis before the above issue is solved.

tuananh:
Ah ok. Just 1 more question, did you mean "dynamically generated" as using in using https://www.npmjs.com/package/redis-commands to generate it?

xtianus79:
the use of generics would help with this

PR #991, opened 2019-10-12
Improved performance of Pipeline.exec

flunderpero:
Execution cost of Pipeline.exec can be reduced to 50% or lower. This is
accomplished by first filling up an Array of Buffers and then calling `Buffer.concat`
instead of calling `Buffer.concat` for each command.
In our use-case (~3000 commands in a single Pipeline) wall-clock-runtime went
down from 1.3sec to 500ms.

### DIFF
--- a/lib/pipeline.ts
+++ b/lib/pipeline.ts
@@ -302,7 +302,7 @@ Pipeline.prototype.exec = function(callback: CallbackFunction) {
     .then(execPipeline);
 
   function execPipeline() {
-    let data: Buffer | string = "";
+    let data = "";
     let buffers: Buffer[];
     let writePending: number = (_this.replyPending = _this._queue.length);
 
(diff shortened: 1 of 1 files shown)


lib/pipeline.ts
```diff
@@ -319,16 +320,25 @@ Pipeline.prototype.exec = function(callback: CallbackFunction) {
           bufferMode = true;
         }
         if (bufferMode) {
-          data = Buffer.concat([
-            typeof data === "string" ? Buffer.from(data, "utf8") : data,
+          if (!buffers) {
+            buffers = [];
+          }
+          if (typeof data === "string") {
```
AVVS: Data is ignored if its already a buffer

flunderpero: `data` cannot be a Buffer. I cleaned up the code a bit to make this clear.
Your comment made me aware that `buffers` is actually never re-set, so I did that, too.

tuananh:
can you take a look at this @luin / @AVVS ?

AVVS:
LGTM, @luin if you are ok with the PR I'd be happy to merge :)

AVVS:
I'll merge this in, I believe releases are going through /next tag and it should be safe

luin:
> I'll merge this in, I believe releases are going through /next tag and it should be safe

Thank you for merging this! I totally missed this pull request. The code LGTM.

PR #998, opened 2019-11-03
Fix race when socket connects immediately on redis.connect()

gfoltz:
I've encountered a race condition that I can reproduce consistently locally. It seems when ioredis.connect creates a new stream / socket, it comes back immediately connected. This causes ioredis to timeout because the `connect` callback never happens (because it's already connected).

This code change fixes my situation and passes all existing tests. Unfortunately, I'm not able to reproduce my situation in the test harness here, despite my best attempts to emulate what's happening to me locally.

### DIFF
--- a/lib/redis/index.ts
+++ b/lib/redis/index.ts
@@ -303,12 +303,22 @@ Redis.prototype.connect = function(callback) {
         if (typeof options.keepAlive === "number") {
           stream.setKeepAlive(true, options.keepAlive);
         }
+        /*
+         * There is a race here. In rare circumstances, the stream can
+         * arrive here already connected. In that case, we just call
+         * the connectHandler directly on the next tick.
+         */
+        const connectHandler = eventHandler.connectHandler(_this);
+        if (stream.connecting) {
+          stream.once(CONNECT_EVENT, connectHandler);
+        } else {
+          process.nextTick(connectHandler);
+        }
 
-        stream.once(CONNECT_EVENT, eventHandler.connectHandler(_this));
         stream.once("error", eventHandler.errorHandler(_this));
         stream.once("close", eventHandler.closeHandler(_this));
 
-        if (options.connectTimeout) {
+        if (options.connectTimeout && stream.connecting) {
           /*
            * Typically, Socket#setTimeout(0) will clear the timer
            * set before. However, in some platforms (Electron 3.x~4.x),


gfoltz:
@luin / @AVVS  thoughts on this?

AVVS:
@gfoltz could this issue be related to a particular version of node? I remember there were some problems where certain events wouldnt fire or would fire immediately and not on the next tick

gfoltz:
@AVVS No. I added logging in a `connect` observer on the stream/socket as soon as it gets created here: https://github.com/luin/ioredis/blob/master/lib/connectors/StandaloneConnector.ts#L72

And then added a log statement at the top of the connector callback here:
https://github.com/luin/ioredis/blob/master/lib/redis/index.ts#L289

I found that the 'connect' event fired by the stream/socket _does happen on a subsequent tick_, it just happens before the stream gets returned to the callback. 

I'm not familiar enough with the details of promises/asCallback to understand if there should be a run loop in between `resolve(this.stream)` here https://github.com/luin/ioredis/blob/master/lib/redis/index.ts#L289 and the start of the callback here https://github.com/luin/ioredis/blob/master/lib/connectors/StandaloneConnector.ts#L79

Empirically speaking, it seems like there must be.

AVVS:
I believe you are right. Another question that pops up is whether we can miss an error event because of the same issue or not

https://github.com/luin/asCallback/blob/master/lib/index.ts#L16 - asCallback would attach .then(...) meaning it definitely goes to the next tick

so if for some reason it connected before getting there the event is missed like you've described

```
    // TODO:
    // We use native Promise here since other Promise
    // implementation may use different schedulers that
    // cause issue when the stream is resolved in the
    // next tick.
    // Should use the provided promise in the next major
    // version and do not connect before resolved.
```

This is the todo from where it says. Ideal solution would be to create the socket, but do not call .connect on it and instead call it after the stream handler is resolved. That way we would be absolutely sure that event isnt missed.

Do you think you can make the change? Current solution works for the connect event, but not for the error

nevir:
> https://github.com/luin/asCallback/blob/master/lib/index.ts#L16 - asCallback would attach .then(...) meaning it definitely goes to the next tick

FWIW, `.then()` will queue a _microtask_ if the promise is already resolved (which will callback at the end of the current tick). If the event is natively queued up on the run loop (e.g. socket events), it'll still fire after than `.then()` resolves

gfoltz:
@AVVS 
>> Ideal solution would be to create the socket, but do not call .connect

By that I think you mean call `new net.Socket` instead of `createConnection`
(See https://nodejs.org/api/net.html#net_new_net_socket_options)

Then move the `connect` call into the connector callback?

That wouldn't be so bad, but I'm a little concerned about the TLS part. Right now it's calling `tls.connect` directly. I think instead it would need to call `new tls.TLSSocket(socket)` with a socket that's created with `new net.Socket`? Not sure about that though.  Testing would be difficult for me since we don't currently use TLS.

>>  Current solution works for the connect event, but not for the error
In an error case, the socket never gets connected, so what would we be missing here? I think the race is specific to the happy path.

AVVS:
> In an error case, the socket never gets connected, so what would we be missing here? I think the race is specific to the happy path.

I'm worried about unhandled error, which would result in a crash

> Then move the connect call into the connector callback?

yep, exactly that! same here in terms of real-world testing for TLS, but seems like what you are suggesting would work (new net.socket + wrap it in tls)

slvnperron:
Will this be merged anytime soon? Considering forking the repo if not. Thanks!

PR #1025, opened 2019-12-12
Ignore empty hosts returned by CLUSTER SLOTS

liukun:
When some nodes disconnected from the cluster, there will be left with some empty addresses.
nodes
Then the `CLUSTER SLOTS` will reply with empty hosts.
slots
`ioredis` then convert the empty address `:0` to `127.0.0.1:6379` internally and try to connect them.
connectionPool
Even worse, when `scaleReads=='slaves'`, it may try to sendCommand to `:0`.
So we'd better just ignore these empty addresses.


luin:
Thanks for the contribution!

Issue #1024, opened 2019-12-12
Redis Cluster keyPrefix doesn't work in redisOptions


wangyx233:
version: 4.14.1
According to doc,  I think keyPrefix in redisOptions. But it doesn't work. Like blow:
```
const redisClient = new Redis.Cluster(nodes, {
  redisOptions: {
    keyPrefix: 'foo:',
  }
});
```
After attempts, it actually works in root option, like this: 
```
const redisClient = new Redis.Cluster(nodes, {
    keyPrefix: 'foo:',
});
```
Is it by design?

zhongshan1993:
I reproduced this bug too.

cobolbaby:
+1

henrikhaugboelle:
+1

aaronleesmith:
This issue was left to die on the vine with an open PR, but it is critical to the functionality of key prefixing. The workaround is to include `keyPrefix` one level up (on the object which is the second parameter to `new Cluster`) but the type definitions will yell at you for trying to do that. So you end up having to do something like this:

```
new Cluster(nodes, clusterRetryStrategy: this.clusterRetryStrategy,
      redisOptions: {
        maxRetriesPerRequest: 1,
        keyPrefix: `${this.config.RedisNamespace}:`,
      },
      /**
       * The entry below fixes a bug where the keyPrefix is totally ignored when used in cluster mode.
       * The issue has an open PR: https://github.com/luin/ioredis/issues/1024
       */
      keyPrefix: `${this.config.RedisNamespace}:`
} as any)
```

It would be great if this PR were merged and this was resolved.

joseph-jja:
Just installed ioredis this week, so using latest.  I am still seeing this issue. Thankfully I found this issue and was able to use this workaround.

anton-johansson:
I also just noticed this issue. While the workaround does work, it's a bit tedious when using TypeScript, since `keyPrefix` is technically not valid on that location.

Any chance this issue can be re-opened? To avoid the workaround and also to avoid future confusion for other people. :)

glensc:
would be nice if the fix be ported to 4.x

florian-besser:
Just encountered this issue using the old `4.x.x` version of IoRedis. It would be great to either backport this feature **or remove keyPrefix from the TypeScript types** for 4.x.x so people are not led to believe this feature is supported, when it isn't.

PR #1041, opened 2020-01-10
ISSUE-1024 keyPrefix option support Redis Cluster

cobolbaby:
https://github.com/luin/ioredis/issues/1024

### DIFF
--- a/lib/commander.ts
+++ b/lib/commander.ts
@@ -144,7 +144,9 @@ function generateFunction(_commandName?: string, _encoding?: string) {
     }
     if (this.options.keyPrefix) {
       options.keyPrefix = this.options.keyPrefix;
-    }
+    } else if (this.options.redisOptions && this.options.redisOptions.keyPrefix) {
+      options.keyPrefix = this.options.redisOptions.keyPrefix;
+    }  
 
     return this.sendCommand(new Command(commandName, args, options, callback));
   };


cobolbaby:
@luin Could the bugfix be merged into the trunk?

eltonpao:
We're looking for this to be merged as well

murbanowicz:
What about this bug? We hit this issue today

luin:
Sorry for the late response. It was indeed an oversight but the fix is somehow a breaking change so I think we should ship this fix in the next major release. For now, we may add a usage notice to README.md to save people's time.

Thanks for the contribution @cobolbaby !

murbanowicz:
To be honest I wouldn't consider it as breaking change but bugfix. 
At this moment ioredis basically does not respect keyPrefix provided in redisOptions which is expected way of doing that (reflected in typings) which is clearly a bug.

For this moment I had to add it to top-level options and it works with Cluster but this is hopefully not long term solution...

guruprasadgvmf (approved):
lookgs good

guruprasadgvmf:
when is this shipping ?

guruprasadgvmf:
why exactly is this a breaking change?

lala7573:
This should be a bugfix.. or please update the documents :'( when is the next major release?

luin:
Hey @cobolbaby 👋 
Sorry for the late response! Resolved this issue with #1522. Thanks for the contribution!

PR #1043, opened 2020-01-13
docs: add more detail to the basic example

barakplasma:
This change only affects the project documentation.

# What
This change adds more details to the basic example. These details show a new user what to expect from the library. 

# Why
As a beginning user of this library, I didn't know what the basic functions returned until I set up a redis server and a notebook to test it out myself. I want to 

# Notes
I tested the proposed file changes at https://runkit.com/barakplasma/ioredis-basic-operations .

### DIFF
--- a/examples/basic_operations.js
+++ b/examples/basic_operations.js
@@ -20,12 +20,12 @@ redis.get("foo", function(err, result) {
   }
 });
 
-redis.del("foo");
-
 // Or ioredis returns a promise if the last argument isn't a function
 redis.get("foo").then(function(result) {
   console.log(result);
 });
+
+redis.del("foo");
 
 // Arguments to commands are flattened, so the following are the same:
 redis.sadd("set", 1, 3, 5, 7);
@@ -41,7 +41,7 @@ redis.hset("myhash", "field1", "Hello");
 redis.hgetall("myhash").then(res => console.log(res)); // Promise resolves to Object {field1: "Hello"} rather than a string, or array of strings
 
 // All arguments are passed directly to the redis server:
-redis.set("key", 100, "EX", 10);
+redis.set("key", 100, "EX", 10); // set's key to value 100 and expires it after 10 seconds
 
 // Change the server configuration
 redis.config("set", "notify-keyspace-events", "KEA");
--- a/examples/basic_operations.js
+++ b/examples/basic_operations.js
@@ -4,7 +4,7 @@ const Redis = require("ioredis");
 const redis = new Redis(process.env.redisPort, process.env.redisEndpoint, {password: process.env.redisPW});
 
 // ioredis supports all Redis commands:
-redis.set("foo", "bar"); // returns promise which resolves to string, "Ok"
+redis.set("foo", "bar"); // returns promise which resolves to string, "OK"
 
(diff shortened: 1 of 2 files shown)


examples/basic_operations.js
```diff
@@ -1,27 +1,42 @@
 "use strict";
 
-var Redis = require("ioredis");
-var redis = new Redis();
+const Redis = require("ioredis");
+const redis = new Redis(process.env.redisPort, process.env.redisEndpoint, {password: process.env.redisPW});
 
 // ioredis supports all Redis commands:
-redis.set("foo", "bar");
+redis.set("foo", "bar"); // returns promise which resolves to string, "Ok"
```
luin: The response should be `"OK"` (in uppercase)

examples/basic_operations.js
```diff
@@ -1,27 +1,42 @@
 "use strict";
 
-var Redis = require("ioredis");
-var redis = new Redis();
+const Redis = require("ioredis");
+const redis = new Redis(process.env.redisPort, process.env.redisEndpoint, {password: process.env.redisPW});
```
luin: We should recommend passing a single option since it's easier for people to understand:

```javascript
new Redis({
  port: process.env.redisPort,
  host: process.env.redisEndpoint,
  password: process.env.redisPW
});
```

examples/basic_operations.js
```diff
@@ -1,27 +1,42 @@
...
 });
 
 // Arguments to commands are flattened, so the following are the same:
 redis.sadd("set", 1, 3, 5, 7);
 redis.sadd("set", [1, 3, 5, 7]);
+redis.spop("set"); // Promise resolves to "5" or another item in the set
+
+// Most responses are strings, or arrays of strings
+redis.zadd("sortedSet", 1, "one", 2, "dos", 4, "quatro", 3, "three")
+console.log(redis.zrange("sortedSet", 0, 2, "WITHSCORES")); // Promise resolves to ["one", "1", "dos", "2", "three", "3"] as if the command was ` redis> ZRANGE sortedSet 0 2 WITHSCORES `
```
luin: We don't need the `console.log` here since it won't print the actual response. A comment for what will be resolved would be enough.

examples/basic_operations.js
barakplasma: done

barakplasma:
Hi @luin ,

Can I also update the README.md example?

examples/basic_operations.js
```diff
@@ -1,30 +1,47 @@
...
+const Redis = require("ioredis");
+const redis = new Redis({
+  port: process.env.redisPort,
+  host: process.env.redisEndpoint,
+  password: process.env.redisPW
+});
 
 // ioredis supports all Redis commands:
-redis.set("foo", "bar");
+redis.set("foo", "bar"); // returns promise which resolves to string, "OK"
```
barakplasma: I'm considering putting all of this within an async function so that we can async/await some of the commands. What do you think?

luin: Sorry for the late response! Yes I think that would be helpful but still we should keep some examples for callback-style. I'm going to merge this pr so feel free to create a new pr for the proposal. Thanks again for the contribution!

PR #1051, opened 2020-01-22
FIX: lodash update for CVE-2018-3721

Gyakusetsu:
# What 
image
image

- https://ossindex.sonatype.org/vuln/0f23ff35-235f-404f-8118-bc1580673fd0
- https://www.npmjs.com/advisories/577
- https://hackerone.com/reports/310443

# Impact

This vulnerability is garanteed to at least obtain denial of service as all the library allow the property "toString" and "valueOf" to be replaced by a "String". This breaks the express module and forces the server to either crash or return a 500 to every subsequent request.
Variable. Server crash or the server becoming unable to respond to all request is garanteed, but more significant impact like remote code execution can be achieved in some cases.


AVVS:
Hey, initially lodash was switched over to partial packages due to size concerns
Where the functions are used? I dont believe it can be exploited as the input to redis settings should be from a trusted source?

I might be missing some cases as I’m writing this based on my memory, not looking through the whole codebase

Please provide some examples where user-based input can realistically make it through to these functions?

Gyakusetsu:
@AVVS it seems like our repository scanner: sonatype-depshield is detecting the lodash-partials as outdated modules. because of subversion provided by npm
btw this is the reference: https://hackerone.com/reports/310443
image

AVVS:
I understand the reference case, but for that to be triggered - malicious input from a user needs to be passed to these functions. Question is whether it is safe to say that user input never makes it to the vulnerable functions or it can make it to the functions and, therefore, the module is vulnerable without the proposed change - could you look through the codebase and determine that?

Gyakusetsu:
@AVVS sorry, my bad. 
Vulnerability already patched: https://github.com/lodash/lodash/commit/d8e069cc3410082e44eb18fcf8e7f3d08ebe1d4a

PR #1095, opened 2020-04-05
Handle connection after connect event was emitted

### LINKED ISSUE #977: Error: connect ETIMEDOUT ----use bluebird
I use bluebird replace native promise, error happened!

```
ioredis:redis status[172.31.15.99:5818]: [empty] -> connecting +0ms
  ioredis:redis queue command[172.31.15.99:5818]: 2 -> set([ 'aaa', '1' ]) +5ms
  ioredis:connection error: Error: connect ETIMEDOUT +0ms
redisClient connect to db error: { Error: connect ETIMEDOUT
    at Socket.<anonymous> (/opt/project/umichat/umichat_manage/node_modules/_ioredis@4.14.1@ioredis/built/redis/index.js:282:31)
    at Object.onceWrapper (events.js:313:30)
    at emitNone (events.js:106:13)
    at Socket.emit (events.js:208:7)
    at Socket._onTimeout (net.js:422:8)
    at ontimeout (timers.js:498:11)
    at tryOnTimeout (timers.js:323:5)
    at Timer.listOnTimeout (timers.js:290:5) errorno: 'ETIMEDOUT', code: 'ETIMEDOUT', syscall: 'connect' }
  ioredis:redis status[172.31.15.99:5818]: connecting -> close +10s
  ioredis:connection reconnect in 50ms +3ms
  ioredis:redis status[172.31.15.99:5818]: close -> reconnecting +1ms
  ioredis:redis status[172.31.15.99:5818]: reconnecting -> connecting +50ms
```

envriment:
```
node: v8.16.0
redis server: 5.0.5
ioredis: 4.14.1
bluebird: 3.5.5
```

my test code
```
const bluebird = require('bluebird');
global.Promise = bluebird;

const instance = new Redis({
  port: 6379,
  host: 'localhost',
  db: 2,
});

instance.on('error', e => {
  console.error(`redisClient connect to db error:`, e);
});

instance.set("aaa", 1)
```

alavers:
It's possible for connect promise to resolve after the resolved
connection has already been established, which means we miss the connect
event. This can happen when Bluebird is set as the global Promise, and
we connect to an ip address host for example. This fix checks to see if
that's the case and invokes the connection handler directly. Thanks
@luin for the recommendation.

Fixes #977

### DIFF
--- a/lib/redis/index.ts
+++ b/lib/redis/index.ts
@@ -304,7 +304,11 @@ Redis.prototype.connect = function (callback) {
           stream.setKeepAlive(true, options.keepAlive);
         }
 
-        stream.once(CONNECT_EVENT, eventHandler.connectHandler(_this));
+        if (stream.connecting) {
+          stream.once(CONNECT_EVENT, eventHandler.connectHandler(_this));
+        } else {
+          process.nextTick(eventHandler.connectHandler(_this));
+        }
         stream.once("error", eventHandler.errorHandler(_this));
         stream.once("close", eventHandler.closeHandler(_this));
 
--- a/lib/redis/index.ts
+++ b/lib/redis/index.ts
@@ -306,42 +306,42 @@ Redis.prototype.connect = function (callback) {
 
         if (stream.connecting) {
           stream.once(CONNECT_EVENT, eventHandler.connectHandler(_this));
+
+          if (options.connectTimeout) {
+            /*
+             * Typically, Socket#setTimeout(0) will clear the timer
+             * set before. However, in some platforms (Electron 3.x~4.x),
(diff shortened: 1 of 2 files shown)


luin:
I think we should ignore `connectTimeout` option in this case right?
https://github.com/luin/ioredis/blob/ec76383a1a85e02aca44954b48c719a69e291e90/lib/redis/index.ts#L315

alavers:
@luin You're right! Pushed a fix

bvallelunga:
Thanks @alavers and @imatlopez for putting up a fix! Dealing with this bug right now and was hoping a fix would come out. Do you think we can get this merged and published soon?

PR #1107, opened 2020-04-21
Fix bug that scripts may not be loaded correctly in pipeline

canyousayyes:
Hello, I want to report a bug in the `pipeline.ts` and submit the bug fix as well.

---

### Description

In the script checking part:

```js
return this.redis
  .script('exists', Array.from(new Set(scripts.map(({ sha }) => sha))))
  .then(function (results) {
    var pending = [];
    for (var i = 0; i < results.length; ++i) {
      if (!results[i]) {
        pending.push(scripts[i]);
      }
    }
```

It will pass a **uniqued scripts array** (`Array.from(new Set(scripts.map(({ sha }) => sha)))`) into `SCRIPT EXISTS`, this is fine.
However it will add the scripts from the **original scripts array** (`pending.push(scripts[i])`), this could introduce a problem.

The bug seems to be introduced since #781.

---

### Example

If I defined 2 custom commands `echo` and `test`, and do:

```js
redis.pipeline()
  .echo()
  .echo()
  .test()
```

Then the `scripts` will contain `["echo", "echo", "test"]`, but the uniqued scripts array is `["echo", "test"]`.
The inconsistencies between the two arrays will make `echo` script will be loaded twice but `test` script not loading into Redis.

---

### Fix

Please review my PR for the fix. I have moved the uniqueness check when adding the scripts.
Now there's only one `scripts` array and there should be no more inconsistency issue.

I have also updated the unit test to cover this case.

### DIFF
--- a/lib/pipeline.ts
+++ b/lib/pipeline.ts
@@ -253,6 +253,7 @@ Pipeline.prototype.exec = function (callback: CallbackFunction) {
 
   // Check whether scripts exists
   const scripts = [];
+  const addedScriptHashes = {};
   for (let i = 0; i < this._queue.length; ++i) {
     const item = this._queue[i];
     if (this.isCluster && item.isCustomCommand) {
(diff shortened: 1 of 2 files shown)


luin:
Good catch! Thanks for your contribution!

canyousayyes:
Thanks for your swift response, I just need this new release for my app :)

Issue #1098, opened 2020-04-08
Allow retryStrategy to be set for Redis instances in cluster


yangyang198703:
The `retryStrategy` is explicitly disabled in https://github.com/luin/ioredis/blob/master/lib/cluster/ConnectionPool.ts#L74. We are adapting this library in our production environment and found performance issue around this default strategy.

Our production redis cluster have ~ 20 shards and each shard with 1 master and 2 slaves,  and the redis `timeout` setting is `30`, which means if a connection doesn't send any command in 30 seconds, it will be closed. 

Totally we would need to have 20 * 3 = 60 connections, if any of the connection doesn't get command sent in 30 seconds, the connection will be closed, and we will send the command to another node which might result  in a `MOVED` error, and finally all the connections to be closed and reconnected.

If QPS is not enough, we will have a lot of unnecessary disconnect and reconnection. Would suggest we don't force the retryStrategy to null.

yangyang198703:
After some investigation, I think this is the reason:

If we allow `retryStrategy` in cluster mode, all the connections to all the nodes will never be disconnected, which means, even no commands are sent, those connections would still repeat reconnecting after redis server close the connection.

I think we could do something to "lazy reconnect", just as we do `lazyConnect`, we can provide an option to set closed connection to `wait` instead of `close` or `reconnecting` when receiving `close` event from stream as if they are freshly created, it will help to reduce the cost of refreshing the cluster slots again.

@luin

yangyang198703:
Proposed a change in https://github.com/luin/ioredis/pull/1108 for addressing this issue, the easiest hack for this is to re-queue the closed nodes and set it to `wait` status under following situations:
1. The cluster is still `ready`
2. The node once successfully connected

The assumption is that the closed connection hitting above criteria should be caused by temporary network issue or idle timeout, and has high possibility to be able to reconnect, thus prevent a whole reconnect to the cluster.

If the closed connection is permanent (server down), the worst case is that the next connection attempted will still fail and it won't hit the criteria anymore - it didn't succeeded connecting to the code as condition #2.

PR #1108, opened 2020-04-22
ISSUE-1098 Allow reconnection closed nodes in cluster mode  #1098

yangyang198703:


luin:
It's hard for users to tell in advance whether a node is lost temporarily or permanently so I don't think we should make this behavior configurable.

The current logic should be able to re-discover the unjoined node or did you find anything wrong?

yangyang198703:
> It's hard for users to tell in advance whether a node is lost temporarily or permanently so I don't think we should make this behavior configurable.
> 
> The current logic should be able to re-discover the unjoined node or did you find anything wrong?

Thanks for replying!

Yes, the current logic works but with a problem of performance. It will cause  disconnections / reconnections of all the nodes if any single node lost connection.

yangyang198703:
> It's hard for users to tell in advance whether a node is lost temporarily or permanently so I don't think we should make this behavior configurable.
> 
I agg
> The current logic should be able to re-discover the unjoined node or did you find anything wrong?

PR #1115, opened 2020-04-28
tls: add tlsSni option - closes #1114

### LINKED ISSUE #1114: [Cluster] TLS-SNI: different servername per node
Hi everyone,

First of all, thank you for this amazing library and the work done here.

I've encountered an issue when trying to connect to a Redis Cluster via TLS where each node possesses its own SNI.

In fact I had no problem connecting to a single node using the `tls.servername` option.

But the `tls` option being common to all nodes I found no way of declaring different SNI per node.

I have made a fix by implementing a new boolean `tlsSni` option that I will introduce in a PR following the opening of this issue.

Thank you for your time.

alexandre-abrioux:
Hi everyone,

This PR introduces a new boolean option `tlsSni` that will basically, when enabled, copy the Redis hostname to `tls.servername` before the connection initialization,
allowing each node of a Redis Cluster to possess its own SNI.

This PR has been opened following the issue #1114 

It is also related to #948

Thank you for your time!

### DIFF
--- a/API.md
+++ b/API.md
@@ -12,75 +12,80 @@
 <a name="Redis"></a>
 
 ## Redis ⇐ <code>[EventEmitter](http://nodejs.org/api/events.html#events_class_events_eventemitter)</code>
-**Kind**: global class  
-**Extends**: <code>[EventEmitter](http://nodejs.org/api/events.html#events_class_events_eventemitter)</code>, [<code>Commander</code>](#Commander)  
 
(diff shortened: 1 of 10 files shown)


alexandre-abrioux:
Added a test for cluster connections.

For clusers I had to copy the `options.host` in `options.hostOriginal` before it gets passed throught the `dnsLookup` function, so that the original hostname could get passed to the `tls.servername` option instead of its IP.

It should be ready to be reviewed now.
Thank you!

alexandre-abrioux:
changed fork branch, PR moved here:
https://github.com/luin/ioredis/pull/1137

PR #1118, opened 2020-05-01
Add Examples for using Redis Streams Commands 

Kareem-Emad:
Added an example for how to use the commands through ioredis interface:
- XADD
- XREAD
- XLEN
https://redis.io/topics/streams-intro

### DIFF
--- a/examples/redis_streams.js
+++ b/examples/redis_streams.js
@@ -1,0 +1,34 @@
+const IORedis = require('ioredis')
+const redis = new IORedis()
+
+// you may find this read https://redis.io/topics/streams-intro
+// very helpfull as a starter to understand the usescases and the parameters used
+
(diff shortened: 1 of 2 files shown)


examples/redis_streams.js
```diff
@@ -0,0 +1,41 @@
+const IORedis = require('ioredis')
+const { promisify } = require('util')
+const redis = new IORedis()
+// this to make them work with await, however you could still
+// use the async version with .then
+const xadd = promisify(redis.xadd).bind(redis)
```
luin, line 6: ioredis has built-in support for promise so I think we can just use `await redis.xadd(xxx)` right?

Kareem-Emad: @luin  yes you are right, fixed

Issue #1097, opened 2020-04-06
Add support for Auth with username & password (Redis 6)


gkorland:

techpanga:
I am able to get this worked by passing below in new Redis({})... Hope this helps.

> password: ['user','password']

gkorland:
@techpanga is seems like it can work as a workaround, but it doesn't feel like good solution for the long term

gkorland:
@techpanga also it seems like all the user must have `+info` on their ACL otherwise it fails when creating the connection.

```
[ioredis] Unhandled error event: ReplyError: NOPERM this user has no permissions to run the 'info' command or its subcommand
    at parseError (/home/guy/redisclients/TestIoRedis/node_modules/redis-parser/lib/parser.js:179:12)
    at parseType (/home/guy/redisclients/TestIoRedis/node_modules/redis-parser/lib/parser.js:302:14)
```

techpanga:
@gkorland , Yes. it is workaround only. Thank you.

luin:
Hey @gkorland @techpanga, actually giving `password` a `['user', 'password']` does work as expected and I think what we need to do is just documenting this usage.

For the `NOPERM` error, if the user doesn't have `+info` you should disable `enableReadyCheck` option like `new Redis({ enableReadyCheck: false })`.

wdyt?

asyrique:
@luin This workaround only works if you're initializing ioredis using the `new Redis({options})`approach and passing in the `password` key in the options object.

However, we're initializing ioredis using the url string approach, passing in something like `redis://foo:bar@somehost:someport`, and this now fails (previously with Redis <6, we'd pass in the password without a user, i.e. `redis://:bar@somehost:someport`).

I'd be happy to submit a PR for this that would fix the the URL string parsing as well as adding an extra optional option `user` to the options object, if you're not already working on it.

Let me know!

luin:
Yeah gotcha. PR is welcome

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

reshreshus:
A workaround would be to use a connection string
```ts
    RedisModule.forRoot({
      type: 'single',
      url: 'redis://default:password@localhost:6379',
    }),
```

PR #1130, opened 2020-05-10
Adds auth support for Redis 6

asyrique:
This PR contains the following changes:
- Adds the `username` optional key to the `IRedisOptions` interface and instantiation options.
- Adds 2 extra error handlers for handling Redis 6 default user not needing password and user passing username and password to a Redis <=5 instance (https://github.com/luin/ioredis/pull/1130/files#diff-43fd638f7be3d3664e2c4ee36dbdf059R32-R50)
- Update URL parser to handle for username passed in redis URL
- Tests for redis 5 and redis 6 using the new auth code.

Fixes luin/ioredis#1097

### DIFF
--- a/lib/redis/index.ts
+++ b/lib/redis/index.ts
@@ -279,7 +279,7 @@ Redis.prototype.connect = function (callback) {
     this.condition = {
       select: options.db,
       auth:
-        options.username && options.username.length > 0
+        options.username
           ? [options.username, options.password]
           : options.password,
       subscriber: false,
--- a/lib/redis/index.ts
+++ b/lib/redis/index.ts
@@ -42,6 +42,7 @@ const debug = Debug("redis");
  * it to reduce the latency.
  * @param {string} [options.connectionName=null] - Connection name.
  * @param {number} [options.db=0] - Database index to use.
+ * @param {string} [options.username=null] - If set, client will send AUTH command with this user and password when connected.
  * @param {string} [options.password=null] - If set, client will send AUTH command
  * with the value of this option when connected.
  * @param {boolean} [options.dropBufferSupport=false] - Drop the buffer support for better performance.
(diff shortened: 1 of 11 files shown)


asyrique:
@luin This should be ready to be reviewed.
I didn't know whether this merited a patch version bump or a minor version bump, so defer to you on that.

Let me know if there's anything I need to fix in this PR.

asyrique:
@luin Is there anything I can change / help to do that can help get this merged? I think I've covered all of the cases in the tests I wrote, but if there's a test case I missed out, if you can describe the case, I can write a test for it.

Thanks and appreciate all the work you do to keep this library fast and secure!

lib/utils/index.ts
```diff
@@ -257,7 +257,11 @@ export function parseURL(url) {
 
   const result: any = {};
   if (parsed.auth) {
-    result.password = parsed.auth.split(":")[1];
+    const parsedAuth = parsed.auth.split(":");
+    if (parsedAuth[0].length > 0) {
```
luin, line 261:
```suggestion
    if (parsedAuth[0]) {
```

I think `parsedAuth[0]` is sufficient to detect empty strings.

lib/redis/index.ts
```diff
@@ -277,7 +278,10 @@ Redis.prototype.connect = function (callback) {
 
     this.condition = {
       select: options.db,
-      auth: options.password,
+      auth:
+        options.username && options.username.length > 0
```
luin, line 282:
```suggestion
        options.username
```

We don't need to check `.length` to detect empty strings right?

luin:
Sorry for the late response. Code-wise it looks good to me 👍, just left two comments. I'll test thoroughly and merge the PR this weekend.

lib/utils/index.ts
asyrique: The reason I used `.length` rather than `parsedAuth[0]` only is because I've been bitten before by the issue where the string `"0"` can be coerced to `0` and thus become falsey. I just tested it and it does seem like within the `if()`, `"0"` is also considered truthy, so this suggestion is valid.

https://developer.mozilla.org/en-US/docs/Glossary/truthy

lib/redis/index.ts
asyrique: Same reasoning as above, and agree with the change.

PR #1134, opened 2020-05-16
fix: revert parsing username via URI due to potential breaking changes

### LINKED ISSUE #1132: v4.17.0 breaking change for Redis parser
Release v4.17.0 fails to parse username:password combinations for redis version 5 and lower. The Rediss / Redis URL connection strings are no longer valid.

```
[ERROR] The server returned "wrong number of arguments for 'auth' command". You are probably passing both username and password to Redis version 5 or below. You should only pass the 'password' option for Redis version 5 and under.
[ioredis] Unhandled error event: ReplyError: NOAUTH Authentication required.
```

luin:
Closes #1132.

Previously, users can specify the username in Redis connect URI unintentionally so parsing username leads potential breaking changes in this case. We still support `username` parameter so users can opt-in ACL explicitly.

### DIFF
--- a/lib/utils/index.ts
+++ b/lib/utils/index.ts
@@ -258,9 +258,6 @@ export function parseURL(url) {
   const result: any = {};
   if (parsed.auth) {
     const parsedAuth = parsed.auth.split(":");
-    if (parsedAuth[0]) {
-      result.username = parsedAuth[0];
-    }
     result.password = parsedAuth[1];
   }
   if (parsed.pathname) {
--- a/test/functional/auth.ts
+++ b/test/functional/auth.ts
@@ -170,23 +170,6 @@ describe("auth", function () {
       redis = new Redis({ port: 17379, username, password });
     });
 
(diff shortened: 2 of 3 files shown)


billinghamj:
Could this be brought back with a flag to allow proper URI parsing? Or in a v5 release?

luin:
@billinghamj I thought we can launch this in v5 but a flag sounds good to me.

bpo:
@luin what do you think about just skipping the username assignment if the username is default?

        if (parsedAuth[0] && parsedAuth[0] != "default") {

I think that would preserve backward compatibility for older servers without requiring a flag or a major release.

luin:
@bpo Actually people may put arbitrary usernames ex "user" for whatever reasons in the URL as they thought it doesn't matter and will be ignored by ioredis. So your approach works in most normal cases but still introduces risks of breaking their assumption.

It's debatable what kind of changes should be considered as breaking changes, but I think we should be cautious anyway so I'd prefer adding a flag or next major release.

gkorland:
@luin do you still plan to put it back?

luin:
> @luin do you still plan to put it back?

Added back via #1284 😄

Issue #1174, opened 2020-08-04
Why flatten the args?


cnwangjie:
When I called `redis.set('a', [1])`, I should receive `['a', [1]]` in argument transformer. But I got `['a', 1]`. 

After I reviewed the code I found `flatten(args)` in https://github.com/luin/ioredis/blob/master/lib/command.ts#L187

It confused me. Is there anyone who could explain why to do that? Thank you.

luin:
It's more of a historical reason due to the convention of node_redis where for convenience users can provide an arbitrary number of parameters with an array. The sugar is less needed nowadays though given the prevalence of ES6 that users can now use the spread operator.

cnwangjie:
Okay, thank you for your reply. And are you willing to provide an option to disable this behavior?

luin:
I'd rather not to provide it as it seems a little weird to change method signatures based on an option. However, the parameters are only flattened for 1 level so `redis.set('a', [[1]])` should work as expected (though also a little weird :-).

cnwangjie:
okay, but I just made a fork to try to solve this problem and submitted a pr (#1175). If you don't like it I can only push a fork version for my case. ;)

PR #1175, opened 2020-08-05
feat: provide an option to allow prevent flatten args

cnwangjie:
fix: #1174


luin:
Sorry for the late response! Don't think it's intuitive to change function signatures with an option and since users can always wrap arguments with an array so I'd not add the complexity here. Thanks anyway for the contribution!

PR #1180, opened 2020-08-11
clean types warnings

gkorland:


lib/cluster/index.ts
```diff
@@ -448,7 +448,7 @@ class Cluster extends EventEmitter {
    */
   private refreshSlotsCache(callback?: CallbackFunction<void>): void {
     if (this.isRefreshing) {
-      if (typeof callback === "function") {
+      if (callback) {
```
luin, line 451: Why do we need to change this? Does it cause type warnings?

luin:
Closing as this has already been merged in v5 😄. Thanks again!

PR #1190, opened 2020-09-09
Replace lodash.defaults with native Object.assign

TrySound:
This utility does not add much value but really confuses newcomers like
me. Also all per function `lodash.*` packages are deprecated so better
to avoid lodash usage if possible.


TrySound:
Hi @luin. Is this good to go?

TrySound:
Friendly ping @luin

TrySound:
Hi @luin What do you think about this change?

TrySound:
Friendly ping @luin

lib/cluster/index.ts
```diff
@@ -910,15 +915,12 @@ const scanCommands = [
...
-        {
-          objectMode: true,
-          key: key,
-          redis: this,
-          command: command,
-        },
-        options
-      )
+      Object.assign({}, options, {
+        objectMode: true,
```
luin, line 919: This can cause a different result as `objectMode` will always be `true` regardless of `options`.

luin (changes requested):

luin:
Sorry @TrySound I overlooked this PR. I agree with the direction to remove all lodash packages in favor of native ones. Will take a look!

lib/cluster/index.ts
TrySound: So all good here?

luin:
The only pitfall is this introduces a potential backward compatibility issue when the provided option contains an `undefined` value due to the difference between `_.defaults` and `Object.assign`:

```javascript
> const userOptions = { port: undefined }
> const defaultOptions = { port: 6379 }
> lodash.defaults({}, userOptions, defaultOptions)
{ port: 6379 }
> Object.assign({}, defaultOptions, userOptions)
{ port: undefined }
```

So I think we should include this in the next major version. WDYT?

TrySound:
I agree

PR #1197, opened 2020-09-25
refactor: replace lodash with native alternatives

azinoviev:


azinoviev:
@luin what do you think about this PR?

lib/command.ts
```diff
@@ -184,7 +183,7 @@ export default class Command implements ICommand {
     this.replyEncoding = options.replyEncoding;
     this.errorStack = options.errorStack;
 
-    this.args = flatten(args);
+    this.args = [].concat(...args);
```
Uzlopak, line 186: This seems not to be a equivalent transformation.

lib/cluster/ConnectionPool.ts
```diff
@@ -71,24 +71,20 @@ export default class ConnectionPool extends EventEmitter {
...
-            // before sending commands to the node.
-            enableOfflineQueue: true,
-            readOnly: readOnly,
-          },
-          node,
-          this.redisOptions,
-          { lazyConnect: true }
-        )
-      );
+      redis = new Redis({
```
Uzlopak, line 74: This is imho not equivalent. The equivalent would be
```typescript

      redis = new Redis({
        // Never try to reconnect when a node is lose,
        // instead, waiting for a `MOVED` error and
        // fetch the slots again.
        retryStrategy: null,
        // Offline queue should be enabled so that
        // we don't need to wait for the `ready` event
        // before sending commands to the node.
        enableOfflineQueue: true,
        readOnly: readOnly,
        ...node,
        ...this.redisOptions,
        lazyConnect: true,
      });
```

lib/utils/index.ts
```diff
@@ -260,7 +259,7 @@ export function parseURL(url) {
     options.allowUsernameInURI && options.allowUsernameInURI !== "false";
   delete options.allowUsernameInURI;
 
-  const result: any = {};
+  const result: any = { ...options };
```
Uzlopak, line 262: This is not equivalent. Better do an Object.assign in line 287 instead of removing it.

lib/command.ts
Uzlopak: Should be probably according to

https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_flatten

this.args = args.reduce( (a, b) => a.concat(b), []);

lib/redis/index.ts
```diff
@@ -217,17 +222,17 @@ Redis.prototype.resetOfflineQueue = function () {
...
+  this.options = { ...Redis.defaultOptions };
   let isTls = false;
   for (let i = 0; i < arguments.length; ++i) {
     const arg = arguments[i];
     if (arg === null || typeof arg === "undefined") {
       continue;
     }
     if (typeof arg === "object") {
-      defaults(this.options, arg);
+      Object.assign(this.options, arg);
```
Uzlopak, line 233: this.options is mutating, but you dont ensure, that the values in this.options are preserved

```suggestion
this.options = {...args, ...this.options};
```

lib/redis/index.ts
```diff
@@ -238,9 +243,8 @@ Redis.prototype.parseOptions = function () {
     }
   }
   if (isTls) {
-    defaults(this.options, { tls: true });
+    this.options = { tls: true, ...this.options };
```
Uzlopak, line 246: here you do it correct

lib/redis/index.ts
```diff
@@ -217,17 +222,17 @@ Redis.prototype.resetOfflineQueue = function () {
...
     const arg = arguments[i];
     if (arg === null || typeof arg === "undefined") {
       continue;
     }
     if (typeof arg === "object") {
-      defaults(this.options, arg);
+      Object.assign(this.options, arg);
     } else if (typeof arg === "string") {
-      defaults(this.options, parseURL(arg));
+      Object.assign(this.options, parseURL(arg));
```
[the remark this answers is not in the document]
Uzlopak: maybe your solution is actually correct. I would also expect that the parseURL is overwriting this.options.

lib/redis/index.ts
Uzlopak: maybe your solution is actually correct. I would also expect that the parseURL is overwriting this.options.

JappiPatel:
## Reviving this PR: Critical Security Vulnerabilities in Lodash Dependencies

I'd like to revive this 2020 PR given the **critical security vulnerabilities** that have emerged in the lodash sub-packages used by ioredis.

### Current State (ioredis@5.10.1)
The lodash sub-packages have **10+ known CVEs** with CVSS scores up to **9.8 (Critical)**:

| Package | Version | Critical CVEs |
|---------|---------|---------------|
| lodash.defaults | 4.2.0 | **CVE-2026-4800** (CVSS 9.8), **CVE-2019-10744** (CVSS 9.1) |
| lodash.isarguments | 3.1.0 | **CVE-2021-23337** (CVSS 7.2), **CVE-2020-8203** (CVSS 7.4) |

### Most Critical Issue

**CVE-2026-4800** (CVSS 9.8 - Critical):
- Command Injection vulnerability in lodash `_.template()` function
- Bypass of previous CVE-2021-23337 fix
- Allows arbitrary code execution at template compilation time
- **Fixed only in lodash 4.18.1+**, but lodash.defaults@4.2.0 (from 2016) will never receive updates

**CVE-2019-10744** (CVSS 9.1 - Critical):
- Prototype Pollution vulnerability
- Can modify Object.prototype affecting all objects in the application

### Why This Matters Now

These lodash sub-packages are **abandoned**. The maintainers consolidated everything into the main `lodash` package years ago, but these sub-packages were never updated with security fixes. Every new CVE discovered in lodash core affects these old sub-packages, but they'll never receive patches.

### Real-World Impact
- Security scanners (Black Duck, Snyk, SAP SVM) flag these **daily**
- Organizations using ioredis cannot resolve these alerts
- Many companies are forced to use `npm overrides` as a workaround, which is fragile

### Request
Given the **4+ year gap** and increasing security concerns, could we please revisit this on urgent basis? The security impact is now significantly higher than when this was originally opened.

**References:**
- CVE-2026-4800: https://nvd.nist.gov/vuln/detail/CVE-2026-4800
- CVE-2019-10744: https://nvd.nist.gov/vuln/detail/CVE-2019-10744
- [Lodash security advisories](https://github.com/lodash/lodash/security/advisories)

Thank you for maintaining ioredis! 🙏

PavelPashov:
@JappiPatel thanks for flagging this. Could you please open an issue with the details from your comment so we can track it properly? This PR may still be useful, but since it targets the v4 branch, it would probably make more sense to open a new PR or continue with [#1996](https://github.com/redis/ioredis/pull/1996).

JappiPatel:
> @JappiPatel thanks for flagging this. Could you please open an issue with the details from your comment so we can track it properly? This PR may still be useful, but since it targets the v4 branch, it would probably make more sense to open a new PR or continue with [#1996](https://github.com/redis/ioredis/pull/1996).

Thanks! Will open an issue and comment on #1996 instead. Appreciate the quick response!

PavelPashov:
Closing as superseded by #2107

PR #1224, opened 2020-10-30
fix: Fix autopipeline and downgrade p-map to support Node 6. [#1216]

### LINKED ISSUE #1216: The dependencie 'p-map' require node >= 10
I find that i can't run it in my project, my node is 6,  is it caused by this problem?

ShogunPanda:
## What's in there?

[x] Fixes a severe autopipeline bug that prevented correct routing.
[x] Fixes #1216 by downgrading `p-map` to `2.1.0`, which is the last one that supports Node 6.
[x] Changes the Travis matrix to run against all Node even version >= 6, plus the current unstable one.

## Note to reviewers

The autopipeline bug was not catch from the test suite since testing for cluster goes against a mock cluster.
Shall we change this to test against a real cluster?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
 language: node_js
 
 node_js:
+  - "6"
   - "8"
-  - "9"
   - "10"
-  - "11"
   - "12"
+  - "14"
+  - "15"
 
 services:
   - redis-server
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ node_js:
   - "10"
   - "12"
   - "14"
-  - "15"
 
 services:
   - redis-server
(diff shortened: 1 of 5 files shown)


mcollina (approved):
lgtm

AVVS:
Node 6 tests wont run due to dev deps requiring higher node version

As for using real cluster - its definitely an option

ShogunPanda:
Ok, I'll remove it from the matrix then.
About the cluster, let me look at Travis documentation!

ShogunPanda:
Node 6 removed.
I think this PR can land to fix the bugs.
I will take care of removing mock cluster in another PR.

AVVS:
As soon as tests finish I'd be able to merge

PR #1226, opened 2020-11-02
fix: Fixed autopipeline performances.

ShogunPanda:
## What's in there?

This PR introduces new internal structures to compare the target nodes for different slots.
In particular, it assigns a numeric ID to each group of nodes (where by "group" I mean a master and its slaves).
This ID is then searchable using both the group name itself or by slot.

The change is needed to improve autopipeline performances, which degraded after the fix in #1224.

## Notes to reviewers

`playground.js` and `.gitignore` changes will be removed when the PR is ready for review/merge.


mcollina:
Good for me!

lib/cluster/index.ts
```diff
@@ -842,6 +849,24 @@ class Cluster extends EventEmitter {
           }
         }
 
+        // Assign to each node keys a numeric value to make autopipeline comparison faster.
+        this._groupsIds = {};
```
AVVS, line 853: Maybe Object.create(null) - should be slightly faster?

ShogunPanda: Done!

ShogunPanda:
I have rebased this after latest master. Are we good to merge this?

luin (approved):

Issue #1228, opened 2020-11-02
getBuffer returns string when enableAutoPipelining is true


adamnoakes:
Hi, firstly, thanks for releasing the new auto pipelining feature.

Whilst testing on my project, I noticed that `getBuffer` returns the result converted to string (rather than a `Buffer`) with `enableAutoPipelining` set to `true`. Is this expected/ a known limitation with the pipelining implementation?

adamnoakes:
Been looking into this a bit further and seems to be caused by *Buffer commands being created with Buffer omitted
https://github.com/luin/ioredis/blob/master/lib/commander.ts#L65-L71
```typescript
commands.forEach(function (commandName) {
  Commander.prototype[commandName] = generateFunction(commandName, "utf8");
  Commander.prototype[commandName + "Buffer"] = generateFunction(
    commandName,
    null
  );
});
```

This `commandName` value is then used for calling the command again from the pipeline
https://github.com/luin/ioredis/blob/master/lib/commander.ts#L110-151
```typescript
...
    // Create a new pipeline and make sure it's scheduled
    return executeWithAutoPipelining(this, commandName, args, callback);
  };
```

At which point the context that it's the Buffer variant of the command (and encoding is null) is lost.

mcollina:
cc @ShogunPanda

ShogunPanda:
I checked the PR. It LGTM!

laverdet:
+1

PR #1231, opened 2020-11-03
Fix autopipeling for buffer function + added built in commands

adamnoakes:
A fix for https://github.com/luin/ioredis/issues/1228
Also came across https://github.com/luin/ioredis/issues/621 whilst looking into this, and realised similar issue could exist for commands created via createBuiltinCommand, i added a suggested addBuiltinCommand to fix this.

### DIFF
--- a/lib/autoPipelining.ts
+++ b/lib/autoPipelining.ts
@@ -17,7 +17,11 @@ export const notAllowedAutoPipelineCommands = [
   "unpsubscribe",
 ];
 
-function findAutoPipeline(client, _commandName, ...args: Array<string>): string {
+function findAutoPipeline(
+  client,
+  _commandName,
+  ...args: Array<string>
+): string {
   if (!client.isCluster) {
     return "main";
   }
@@ -68,8 +72,13 @@ function executeAutoPipeline(client, slotKey: string) {
   });
 }
 
-export function shouldUseAutoPipelining(client, commandName: string): boolean {
+export function shouldUseAutoPipelining(
+  client,
+  functionName: string,
+  commandName: string
+): boolean {
   return (
+    functionName &&
     client.options.enableAutoPipelining &&
     !client.isPipeline &&
     !notAllowedAutoPipelineCommands.includes(commandName) &&
@@ -79,6 +88,7 @@ export function shouldUseAutoPipelining(client, commandName: string): boolean {
 
 export function executeWithAutoPipelining(
   client,
+  functionName: string,
   commandName: string,
   args: string[],
   callback
@@ -94,10 +104,13 @@ export function executeWithAutoPipelining(
           return;
         }
 
-        executeWithAutoPipelining(client, commandName, args, callback).then(
-          resolve,
-          reject
-        );
+        executeWithAutoPipelining(
+          client,
+          functionName,
+          commandName,
+          args,
+          callback
+        ).then(resolve, reject);
       });
     });
   }
@@ -138,7 +151,7 @@ export function executeWithAutoPipelining(
       resolve(value);
     });
 
-    pipeline[commandName](...args);
+    pipeline[functionName](...args);
   });
 
   return asCallback(autoPipelinePromise, callback);
--- a/lib/commander.ts
+++ b/lib/commander.ts
@@ -135,14 +135,14 @@ Commander.prototype.defineCommand = function (name, definition) {
  */
 Commander.prototype.sendCommand = function () {};
 
-function generateFunction(_functionName: string | null, _encoding: string);
+function generateFunction(functionName: string | null, _encoding: string);
 function generateFunction(
-  _functionName: string | null,
+  functionName: string | null,
   _commandName: string | void,
   _encoding: string
 );
 function generateFunction(
-  _functionName: string | null,
+  functionName: string | null,
   _commandName?: string,
   _encoding?: string
 ) {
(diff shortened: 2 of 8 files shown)


lib/commander.ts
```diff
@@ -57,21 +57,46 @@ Commander.prototype.getBuiltinCommands = function () {
...
+ * @return {object} functions
+ * @public
+ */
+Commander.prototype.addBuiltinCommand = function (commandName) {
+  Commander.prototype[commandName] = generateFunction(
+    commandName,
+    commandName,
+    "utf8"
+  );
+  Commander.prototype[commandName + "Buffer"] = generateFunction(
```
adamnoakes, line 78: I think https://github.com/luin/ioredis/blob/master/lib/pipeline.ts#L64 needs to be change to:
```typescript
Pipeline.prototype = Object.create(Commander.prototype);
```
for this to actually work.
e.g. https://repl.it/repls/DesertedCookedApplicationframework#index.js

adamnoakes:
@ShogunPanda

lib/autoPipelining.ts
```diff
@@ -68,8 +72,13 @@ function executeAutoPipeline(client, slotKey: string) {
...
 }
 
-export function shouldUseAutoPipelining(client, commandName: string): boolean {
+export function shouldUseAutoPipelining(
+  client,
+  functionName: string,
+  commandName: string
+): boolean {
   return (
+    functionName &&
```
ShogunPanda, line 81: In general looks fine to me.
The only doubt I have is here: why you also check for functionName here?

lib/commander.ts
```diff
@@ -57,21 +57,46 @@ Commander.prototype.getBuiltinCommands = function () {
...
 };
 
+/**
+ * Create add builtin command
+ *
+ * @param {string} commandName - command name
+ * @return {object} functions
+ * @public
+ */
+Commander.prototype.addBuiltinCommand = function (commandName) {
```
ShogunPanda, line 72: Why this new function?

adamnoakes: This is not really necessary for this PR, but whilst looking into the code/ previous issued i noticed #621 and realised similar issue probably exists for commands created via createBuiltinCommand, this would be a possible fix for that also but think it  would require https://github.com/luin/ioredis/blob/master/lib/pipeline.ts#L64 to be change to:
```typescript
Pipeline.prototype = Object.create(Commander.prototype);
```
for this to actually work.

lib/autoPipelining.ts
adamnoakes: commands created by `createBuiltinCommand` would have no functionName, this would prevent the auto pipelining breaking for commands created in this way

ShogunPanda: I see.
But why do you want to prevent autopipeline to run against them? Autopipeline could support them.

lib/commander.ts
ShogunPanda: Got it about the new function.
Why replacing the prototype rather than merging it?

lib/commander.ts
ShogunPanda: Why? 
The repl test will never work as you defining `newFn` after the prototype would have been assigned. 
Are you trying to keep the prototype in sync?

adamnoakes: It does work with (as it's using Object.create), that's what the repl test is demonstrating. Maybe it's clearer like this:
image

lib/autoPipelining.ts
adamnoakes: I don't want to prevent, but the best way i can see to support it without creating an API that exposes the internals of this is via the `addBultinCommand` method i have created

lib/commander.ts
ShogunPanda: Ok, that's what I thought then. You use Object.create so that the prototype of the children gets new addition of the prototype of the parent.
Looks fine to me!

lib/autoPipelining.ts
ShogunPanda: 👍

lib/commander.ts
ShogunPanda: 👍

ShogunPanda:
The PR LGTM to me, go on and file it!

adamnoakes:
> The PR LGTM to me, go on and file it!

Ok great, i'll try get this PR ready in the next couple days

lib/commander.ts
adamnoakes: Ended up doing this slightly differently as ran into issues with Redis inheriting from EventEmitter. I saw that defineCommand was doing a similar thing that i wanted to do so have done a similar implementation by adding the command to the instance rather than the prototype, then maintaining a set for the added commands.

ShogunPanda: Wonderful. Much cleaner i think!

SimonWoolf:
This seems to have fallen through the cracks - @luin is anything holding up this PR from being merged? (It would also be useful for us). Thanks! 🙂

artur-ma:
@luin anything new about it ?

rahbaner:
Hi @luin , any update on this? A friendly ping

luin (approved):

luin:
Sorry for the delay! I overlooked the notifications for this thread. Awesome works @adamnoakes & @ShogunPanda!

rahbaner:
Hi @luin, thanks, I will try this change on Monday. Thanks @adamnoakes and @ShogunPanda

Issue #1215, opened 2020-10-26
`_addedScriptHashesCleanInterval` is not always cleared on `redis.quit()`


trentm:
Hi. I believe https://github.com/luin/ioredis/commit/3b22f8f89c3239fbbbc16da692651f6a35eca53a added a bug where the new `this._addedScriptHashesCleanInterval = setInterval(...)` (https://github.com/luin/ioredis/commit/3b22f8f89c3239fbbbc16da692651f6a35eca53a#diff-d3d9763dec19cf3dd977c437e0be3e6068b13161ee5a2b9b0bf225bddd2d83f8R188) isn't cleared via `clearInterval()` in all cases. This results in a dangling active handle that will hang a node process.

Here is a repro:

```js
var Redis = require('.'); // ioredis
var redis = new Redis();
redis.on('ready', function () {
    console.log('ready');
    redis.quit();
    console.log('done');
});
```

```
% node hang.js
ready
done
<hangs here>
```

or with ioredis debug output:

```
% DEBUG=ioredis:* node hang.js
  ioredis:redis status[localhost:6379]: [empty] -> connecting +0ms
  ioredis:redis status[127.0.0.1:6379]: connecting -> connect +6ms
  ioredis:redis write command[127.0.0.1:6379]: 0 -> info([]) +2ms
  ioredis:redis status[127.0.0.1:6379]: connect -> ready +4ms
ready
  ioredis:redis write command[127.0.0.1:6379]: 0 -> quit([]) +1ms
done
  ioredis:redis status[127.0.0.1:6379]: ready -> close +3ms
  ioredis:connection skip reconnecting since the connection is manually closed. +0ms
  ioredis:redis status[127.0.0.1:6379]: close -> end +1ms
<hangs here>
```

This does not happen if one uses `redis.disconnect()` instead of `redis.quit()`.

When calling `redis.quit()`, the closeHandler (in event_handler.ts) is called with `self.manuallyClosing == true` which does **not** call the  `clearInterval` added to `Redis.prototype.disconnect()`.

Some options I see (though I am not very familiar with ioredis, so I'm definitely happy to be overriden):

1. Add `this._addedScriptHashesCleanInterval.unref()` (https://nodejs.org/api/timers.html#timers_timeout_unref) to the `connect` method.  That'll prevent node from hanging, but *will* persist the setInterval after a program is presumably finished with the ioredis session.
2. Refactor the use of the `closeHandler` to explicitly call `clearInterval`.
[...]

PR #1232, opened 2020-11-06
fix: clear script caches interval on each connection attempt

ejose19:
Follow-up fix for: #1215

After using tools like ```why-is-node-running```, I saw:

```ts
this._addedScriptHashesCleanInterval = setInterval(() => {
  this._addedScriptHashes = {};
}, this.options.maxScriptsCachingTime);
```

appearing multiple times (after ```quit``` was called), which means ```quit``` was only clearing the last of them, but there were many that were overwritten by that same call when connect is called multiple times (as in not  connecting the first time). This PR ensures that the interval is cleared before assigning a new one to the variable holding them.

### DIFF
--- a/lib/cluster/index.ts
+++ b/lib/cluster/index.ts
@@ -185,6 +185,7 @@ class Cluster extends EventEmitter {
         return;
       }
 
+      clearInterval(this._addedScriptHashesCleanInterval);
       this._addedScriptHashesCleanInterval = setInterval(() => {
         this._addedScriptHashes = {};
       }, this.options.maxScriptsCachingTime);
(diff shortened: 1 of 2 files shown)


AVVS:
Hey, this looks awesome. Any chance you are able to add a test case for this so that we can ensure there are no regressions in the future?

ejose19:
I couldn't think of any test for this particular fix, as the way I found it is either using a low level tool ```why-is-node-running```, or noticing why the server was hanging on ```SIGTERM```, but the tests are ignoring background intervals. If you know of any test case for it using the existing test tools I may be able to work on it.

ShogunPanda:
@ejose19 Thanks for filing this.

After your bug report, I've investigated separately and I found the same fix, which I incorporated in #1226.

@AVVS I've tried to add a test case in that PR as well, but honestly I was not able to reproduce the failure as well.

PR #1233, opened 2020-11-07
Sanitize errors [#1229]

### LINKED ISSUE #1229: Error contains password in plaintext
Hi, I'm using ioredis like this (simplified):
```typescript
import Redis from 'ioredis'

const password = process.env.REDIS_PASSWORD // e.g. 'abc123'
const redis = new Redis({ password })

redis.on('error', (error) => {
  logger.error('Redis error', error)
})
```

Due to a human error, the password used was not the correct one for my service, so ioredis could not connect.
This resulted in the following error:
```typescript
Error {
  command: {
    name: 'auth',
    args: ['abc123']
  }
}
```
Since this error gets emitted and ultimately logged, it ended up in plaintext in my service logs, which for some could equate to a security incident.

I've since made sure to omit`error.command.args` before logging. But is this expected behavior? Is there a better way to handle this?

smartinio:
Fixes #1229


smartinio:
@luin Friendly reminder. Let's get this merged?

artur-ma:
@luin is it planned to be merged?

lib/errorSanitizer.ts
```diff
@@ -0,0 +1,31 @@
...
+    Object.create(Object.getPrototypeOf(instance)),
+    instance
+  );
+}
+
+export function sanitizeError(originalError: RedisError): RedisError {
+  if (!originalError?.command?.args) {
+    return originalError;
+  }
+  const error = clone(originalError);
```
marcbachmann, line 16: This can't be taken in like that. It breaks the stack traces.
I'd rather just use `originalError.args = undefined`

```js
const err = new Error()
const err2 = Object.assign(
  Object.create(Object.getPrototypeOf(err)),
  err
);

console.log(err.stack)
// 'Error\n    at <anonymous>:1:13'
console.log(err2.stack)
// undefined
```

[v5](#1504) Is in the works. Maybe we can check whether we can conditionally set it during error construction.

marcbachmann:
Weird. The `command` attribute isn't following the type declaration of the `AbortError` in [DefinitelyTyped/types/redis-errors](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/redis-errors/index.d.ts#L24-L27).

https://github.com/luin/ioredis/blame/master/lib/redis/event_handler.ts#L104-L108
https://github.com/luin/ioredis/blame/master/lib/redis/event_handler.ts#L4

I'd have voted that we're making the args non-enumerable. People can still access it, but it shouldn't be present with a shallow clone or property iteration.

lib/errorSanitizer.ts
smartinio: Removed the pseudo cloning and opted for mutating the error instead. Don't really have time to spend refactoring this into something cleaner.

Issue #1235, opened 2020-11-10
ioredis won't throw errors, prints to console instead - Unhandled error event: Error: connect ECONNREFUSED


vktrl:
Error handling seems to be broken when trying to query a non-existent database. To reproduce:

```
const Redis = require('ioredis');

const settings = {
  host: "localhost",
  port: 1234 // any invalid port
};

const redis = new Redis(settings);

(async function() {
  try {
    await redis.info()
  } catch (err) {
    // no error is caught, ioredis prints straight to console
    console.log("caught: ", err)
  }
})()
```

Results in console getting flooded with:
```
[ioredis] Unhandled error event: Error: connect ECONNREFUSED 127.0.0.1:6379
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1146:16)
```

Version 4.19.2

peteraw77:
Experiencing the same issue, tried passing a reconnectOnError function but no use

peteraw77:
The proposed solution in #1209 seemed to resolve this for me.

vktrl:
I don't see how this helps. Ioredis should throw an error so I could do something about it.

vktrl:
> The proposed solution in #1209 seemed to resolve this for me.

I don't see how this helps. Ioredis should throw an error so I could do something about it.

smartinio:
@vktrl Redis extends `EventEmitter`, so it doesn't `throw` the error. It `emit`s the error. You need to attach an error handler using e.g:
```typescript
redis.on('error', (e) => doSomething())
```

vktrl:
> @vktrl Redis extends `EventEmitter`, so it doesn't `throw` the error. It `emit`s the error. You need to attach an error handler using e.g:
> 
> ```ts
> redis.on('error', (e) => doSomething())
> ```

Oh, I see, thanks!

cope:
And what can we do in redis.on('error' to make this stop? I tried .disconnect and .quit and nothing helps 😞

Issue #1209, opened 2020-10-16
Unhandled error event: Error: connect EHOSTDOWN


OPScorp:
Hi,
I am getting Error: connect EHOSTDOWN and terminating of server process when connecting to unavailable Redis server.
Sometimes on start and sometimes on reconnecting try.

Problem in this file: [StandaloneConnector.js](https://github.com/luin/ioredis/blob/master/lib/connectors/StandaloneConnector.ts)
in function connect.

process.nextTick(() => {
                if (!this.connecting) {
                    reject(new Error(utils_1.CONNECTION_CLOSED_ERROR_MSG));
                    return;
                }
                try {
                    if (options.tls) {
                        this.stream = tls_1.connect(connectionOptions);
                    }
                    else {
                        this.stream = net_1.createConnection(connectionOptions);
                    }
                    // Need to add this code!!
                    this.stream.once("error", err => {
                        reject(err);
                    });
                }
                catch (err) {
                    reject(err);
                    return;
                }
                resolve(this.stream);
            });

P.S.: this is final fix. Hope help someone with this.

peteraw77:
Is there any reason you haven't made a PR with this change? It solved the issue for me and I think it would be good to bring it to the maintainer's attention.

andywelters:
I have the same issue.  Let's make a PR.

PR #1236, opened 2020-11-12
Handle stream error in StandaloneConnector.js

peteraw77:
Solution proposed by @OPScorp in #1209 . Solves #1235 as well. I'm not sure why he didn't PR this himself but it was helpful for myself and others so I would like to get it merged.


AVVS:
@peteraw77 thanks for the PR, but it does definitely need a test case. Error handlers are typically assigned later on in the chain once the connection was created. I dont remember the reasoning for not having an error handler in the create helper, either way a test covering this would be required to get this merged

peteraw77:
@AVVS where would be the right place to put the test case?

peteraw77:
Added a test in test/functional/connection.ts but I'm having a hard time getting the tests to reproduce the error. This was my best attempt but it's definitely not ideal.

test/functional/connection.ts
```diff
@@ -240,6 +240,21 @@ describe("connection", function () {
         done();
       });
     });
+
+    it("should reject on initial connection error", function (done) {
+      const stub = sinon
+        .stub(net.Socket.prototype, "connect")
+        .callsFake((options) => {
+          net.Socket.prototype.emit("error");
```
AVVS, line 248: this need to be async, .connect() doesnt throw in the same tick - wrap this in process.nextTick

https://nodejs.org/api/net.html#net_socket_connect

peteraw77: Thanks, it seems to be capturing the behaviour I intended it to now

francescov1:
Is there an expected timeline on when this will be merged? This issue has been causing endless headaches for our team

drobnikj:
Ping here. I would be great to merge this fix soon. It generates Unhandled errors, which are hard to debug. In our case it was an error: 
```
[ioredis] Unhandled error event: Error: read EFAULT
    at TCP.onStreamRead (internal/stream_base_commons.js:209:20)
    at TCP.callbackTrampoline (internal/async_hooks.js:126:14)
```
It takes us time to find that bug comes from this issue.
Thanks!

francescov1:
> Ping here. I would be great to merge this fix soon. It generates Unhandled errors, which are hard to debug. In our case it was an error:
> 
> ```
> [ioredis] Unhandled error event: Error: read EFAULT
>     at TCP.onStreamRead (internal/stream_base_commons.js:209:20)
>     at TCP.callbackTrampoline (internal/async_hooks.js:126:14)
> ```
> 
> It takes us time to find that bug comes from this issue.
> Thanks!

Same here, it's such a simple, obvious but crucial fix...

kapalkat:
Please merge that change as we had to write dedicated service to catch those unhandled errors...

ctrlaltdylan:
Also pinging! If I was a maintainer I would merge. But alas I am a simple consumer of this great package.

luin:
Sorry for the late response. I didn't notice this PR... Will take a look soon!

luin:
`reject(err)` here actually always does nothing as the Promise has already resolved at line 83. This is equivalent to `this.stream.once('error', () => {})`, all it does is ignoring the stream error instead of handling them properly. So this PR should be considered as a workaround.

I created a PR to address the root issue: #1296. I'll leave this open and may merge this if I can't get #1296 done very soon. Thank you for the contributions!

ctrlaltdylan:
You the man @luin !

PR #1243, opened 2020-11-23
Fix typo for stream in README.md

Motsols:
Found a typo in readme where it said `stream#pause()` instead of `stream.pause()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -601,7 +601,7 @@ const stream = redis.hscanStream("myhash", {
 You can learn more from the [Redis documentation](http://redis.io/commands/scan).
 
 **Useful Tips**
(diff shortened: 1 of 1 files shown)


luin:
Actually we should change both to `Stream#` as `#` indicates instance methods. https://stackoverflow.com/questions/736120/why-are-methods-in-ruby-documentation-preceded-by-a-hash-sign

Motsols:
Interesting, I thought it was a typo. Thanks for taking the time to look up an explanation.

If I understand it correctly, by writing Stream#pause() readers will know that they should do 
```
const stream = redis.scanStream();
stream.pause()
``` 

Instead of something like this?
`const pausedStream = redis.scanStream().pause();`

Motsols:
Updated according to comment and all checks are green

PR #1253, opened 2020-12-07
Connect to Redis through SRV record as a service discovery layer

ohadisraeli:
Connect to Redis using a DNS service discovery layer, the SRV record holds a URL and port that allows changing the port and host without application code change.

### DIFF
--- a/lib/cluster/index.ts
+++ b/lib/cluster/index.ts
@@ -31,6 +31,7 @@ import Redis from "../redis";
 import Commander from "../commander";
 import Deque = require("denque");
 import { Pipeline } from "..";
+import * as dns from "dns";
 
 const debug = Debug("cluster");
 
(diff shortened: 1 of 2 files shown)


lib/cluster/index.ts
```diff
@@ -930,17 +1039,34 @@ class Cluster extends EventEmitter {
...
+    if (this.options.useSRVRecord) {
+      return Promise.all(
+        hostnames.map((hostname) => this.resolveSrv(hostname))
+      ).then((ips) => {
+        const hostnameToIP = zipMap(hostnames, ips);
+        let result = [];
+        startupNodes.map((node) => {
+          if (hostnameToIP.has(node.host)) {
+            if (ips[0][0].name) {
+              result.push({ host: ips[0][0].name, port: ips[0][0].port });
```
luin, line 1051: Looks like `hostnameToIP` is not actually used? I suppose it should be something like `hostnameToIP.get(node.host)` instead of `ips[0]`?

Also a nit: I think `.filter().map()` should make the code more clear.

lib/cluster/index.ts
```diff
@@ -890,6 +956,49 @@ class Cluster extends EventEmitter {
     });
   }
 
+  private resolveSrv(hostname: string): Promise<any> {
+    return new Promise((resolve, reject) => {
+      dns.resolveSrv(hostname, function (err, addrs) {
+        if (err) {
+          /* no SRV record, try domain as A */
+          resolve(err);
```
luin, line 964: I don't see where do we handle this error?

lib/cluster/index.ts
```diff
@@ -890,6 +956,49 @@ class Cluster extends EventEmitter {
...
+              } else {
+                reject(error);
+              }
+            }
+          };
+          // cb1(null,addrs[0])
+          let gSRV = groupSrvRecords(addrs);
+          pending = gSRV.length;
+          gSRV.forEach(function (addr) {
+            resolveHost(addr.name, function (e, a) {
```
luin, line 985: I'd suggest `resolveHost` returns a Promise so we can use `Promise.all` to avoid nested callbacks.

luin (changes requested):
Thanks for your contribution! It looks promising, but can you add some tests and documentation for this change?

Also added some reviews.

PR #1292, opened 2021-02-27
Update min Node version to 8

azinoviev:
It looks like Node 6 was dropped from CI.


luin:
Hi @azinoviev, thanks for the contribution! We do consider dropping support for Node versions as a breaking change so will do this in the next major version.

luin:
Resolved in v5. Thanks for raising this up!

Issue #1211, opened 2020-10-19
When connecting using an ACL user (Redis 6), connection will fail unless the user has permissions to run the INFO command.


simonprickett:
Hi there,

I've been working with Redis 6 ACL users and ioredis, and I've noticed that a connection can't be made successfully unless the ACL user is given access to the `INFO` command.  Consider this ACL user, who has access only to the `GET` command:

```
127.0.0.1:6379> acl getuser simon
1) "flags"
2) 1) "on"
   2) "allkeys"
3) "passwords"
4) 1) "34fb46c847bb9df96e5205a39d382f648a6e8dce1e014cd85b4ca6a88d88ed03"
5) "commands"
6) "-@all +get"
7) "keys"
8) 1) "*"
```

When I run this test code:

```javascript
const Redis = require('ioredis');
const redis = new Redis({
  host: '127.0.0.1',
  port: 6379,
  username: 'simon',
  password: 'redis'
});

const redisDemo = async () => {
  const reply = await redis.get('foo');
  console.log(reply);
};

redisDemo();
```

I get the following error:

```
[ioredis] Unhandled error event: ReplyError: NOPERM this user has no permissions to run the 'info' command or its subcommand
...
```

Updating the ACL user's permissions to allow use of the `INFO` command fixes this with no code changes in my test program:

```
127.0.0.1:6379> acl getuser simon
1) "flags"
2) 1) "on"
   2) "allkeys"
3) "passwords"
4) 1) "34fb46c847bb9df96e5205a39d382f648a6e8dce1e014cd85b4ca6a88d88ed03"
5) "commands"
6) "-@all +get +info"
7) "keys"
8) 1) "*"
```

I think this should either be mentioned in the documentation, or handled in the code in a way that the connection is still made, but the `INFO` command is never run?

I am using ioredis 4.17.3 from npm.

ejose19:
You can already disable the info call with `enableReadyCheck: false` on `RedisOptions`. However users may not want to opt-out completely, and after reviewing https://github.com/Rua-Yuki/ioredis/blob/4aaa3918f287fa3fca9a79bcdf4ec7d4b1be2203/lib/redis/index.ts#L492-L528 I see only the `Persistence` section is being used, so we may want to change the default (or add another option) to use `info persistence` instead of just `info`, and users can grant `+info|persistence` limiting the information given to the client.

luin:
Fixed in v5. Sorry for the late response.

PR #1293, opened 2021-03-04
refactor: limit `info` on readyCheck to `persistence` section

ejose19:
Related: #1211

Limits the `info` call on `readyCheck` to just `persistence` section, allowing users to specify the exact acl (`+info|persistence`) that `ioredis` needs on `readyCheck`

### DIFF
--- a/lib/redis/index.ts
+++ b/lib/redis/index.ts
@@ -526,7 +526,7 @@ Redis.prototype.flushQueue = function (error, options) {
  */
 Redis.prototype._readyCheck = function (callback) {
   const _this = this;
-  this.info(function (err, res) {
+  this.info('persistence', function (err, res) {
     if (err) {
       return callback(err);
     }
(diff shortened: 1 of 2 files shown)


luin (commented):
Is dump.rdb necessary?

ejose19:
@luin not really, didn't notice that file. Should be good now.

marcbachmann:
Maybe we could also remove `client.serverInfo` completely https://github.com/luin/ioredis/blob/cfb04a062b380bad5655b6f97b4259f328f1ee4a/lib/redis/event_handler.ts#L91

luin:
Turns out Sentinel connector uses the `role` field which is under the `replication` section: https://github.com/luin/ioredis/blob/5cd4f57/lib/connectors/SentinelConnector/index.ts#L74. So a couple of ideas:

1. Just don't do anything. If the user doesn't have permission to run `info`, they can disable the ready check as a workaround.
2. Only provide `presistence` in the standalone mode, and run without arguments in the sentinel mode.

Not sure which one is the best. I'd prefer 1 but maybe we can catch the NOPERM error and print a hint to tell users they can disable the ready check.

> Maybe we could also remove client.serverInfo completely

Yeah we can remove it.

PR #1304, opened 2021-03-19
fix: support parallel script execution in pipelines

### LINKED ISSUE #1303: Parallel .multi requests with custom scripts result in NOSCRIPT error
In a project I'm working on we're using ioredis with redis streams and some custom lua scripts.
Locally and also on our CI, we've observed that there are `NOSCRIPT` errors and most times they occur in the same tests.

I've digged down into the issue and found out that ioredis doesn't queue the second command as it thinks the script is already initialized. The problematic line is here: https://github.com/luin/ioredis/blob/master/lib/pipeline.ts#L321

Most likely we only want to set the variable once we get a successful `exists` check back from redis.
So we could move the assignment after the `load` promise: https://github.com/luin/ioredis/blob/master/lib/pipeline.ts#L341

Reproduction:
```js
// DEBUG=ioredis.* node index.js
async function start () {
  const Redis = require('ioredis')
  const random = Math.random()
  const redis = new Redis()
  redis.defineCommand('something', {numberOfKeys: 0, lua: `return "${random}"`})
  const [[first], [second]] = await Promise.all([
    redis.multi([['something']]).exec(),
    redis.multi([['something']]).exec()
  ])

  console.log({first: first[0], second: second[0]})

  redis.disconnect()
}

start()
```

Output:
```
  ioredis:redis status[localhost:6379]: [empty] -> connecting +0ms
  ioredis:redis queue command[localhost:6379]: 0 -> script([ 'exists', '6ddff55aad13ecdf2763faa3a884ec066ebf0724' ]) +2ms
  ioredis:redis queue command[localhost:6379]: 0 -> multi([]) +2ms
  ioredis:redis queue command[localhost:6379]: 0 -> evalsha([ '6ddff55aad13ecdf2763faa3a884ec066ebf0724', '0' ]) +0ms
  ioredis:redis queue command[localhost:6379]: 0 -> exec([]) +0ms
  ioredis:redis status[127.0.0.1:6379]: connecting -> connect +5ms
  ioredis:redis write command[127.0.0.1:6379]: 0 -> info([]) +0ms
  ioredis:redis status[127.0.0.1:6379]: connect -> ready +3ms
  ioredis:connection send 4 commands in offline queue +0ms
  ioredis:redis write command[127.0.0.1:6379]: 0 -> script([ 'exists', '6ddff55aad13ecdf2763faa3a884ec066ebf0724' ]) +0ms
  ioredis:redis write command[127.0.0.1:6379]: 0 -> multi([]) +0ms
  ioredis:redis write command[127.0.0.1:6379]: 0 -> evalsha([ '6ddff55aad13ecdf2763faa3a884ec066ebf0724', '0' ]) +0ms
  ioredis:redis write command[127.0.0.1:6379]: 0 -> exec([]) +1ms
  ioredis:redis write command[127.0.0.1:6379]: 0 -> script([ 'load', 'return "0.42066807125457295"' ]) +0ms
  ioredis:redis write command[127.0.0.1:6379]: 0 -> multi([]) +2ms
  ioredis:redis write command[127.0.0.1:6379]: 0 -> evalsha([ '6ddff55aad13ecdf2763faa3a884ec066ebf0724', '0' ]) +0ms
  ioredis:redis write command[127.0.0.1:6379]: 0 -> exec([]) +0ms
{
  first: null,
  second: ReplyError: NOSCRIPT No matching script. Please use EVAL.
      at parseError (/node_modules/redis-parser/lib/parser.js:179:12)
      at parseType (/node_modules/redis-parser/lib/parser.js:302:14)
}
  ioredis:redis status[127.0.0.1:6379]: ready -> close +5ms
  ioredis:connection skip reconnecting since the connection is manually closed. +8ms
  ioredis:redis status[127.0.0.1:6379]: close -> end +0ms
```

marcbachmann:
Fixes #1303

Support parallel .multi script execution by delaying the assignment of the load confirmation.

btw. regarding the project setup: the package-lock.json links to `http://artprod.dev.bloomberg.com/artifactory/api/npm/npm-repos/` as npm registry, which isn't publicly available. Therefore `npm install` and `npm ci` fails locally. By replacing all the urls with the npm registry, the npm install works again.

### DIFF
--- a/lib/pipeline.ts
+++ b/lib/pipeline.ts
@@ -313,12 +313,15 @@ Pipeline.prototype.exec = function (callback: CallbackFunction) {
 
     const script = this._shaToScript[item.args[0]];
 
-    if (!script || this.redis._addedScriptHashes[script.sha]) {
+    if (
+      !script ||
+      this.redis._addedScriptHashes[script.sha] ||
+      scripts.includes(script)
+    ) {
       continue;
     }
 
     scripts.push(script);
-    this.redis._addedScriptHashes[script.sha] = true;
   }
 
   const _this = this;
@@ -330,7 +333,12 @@ Pipeline.prototype.exec = function (callback: CallbackFunction) {
   if (this.isCluster) {
     return pMap(scripts, (script) => _this.redis.script("load", script.lua), {
       concurrency: 10,
-    }).then(execPipeline);
+    }).then(function () {
+      for (let i = 0; i < scripts.length; i++) {
+        _this.redis._addedScriptHashes[scripts[i].sha] = true;
+      }
+      return execPipeline();
+    });
   }
 
   return this.redis
@@ -352,7 +360,12 @@ Pipeline.prototype.exec = function (callback: CallbackFunction) {
         })
       );
     })
-    .then(execPipeline);
+    .then(function () {
+      for (let i = 0; i < scripts.length; i++) {
+        _this.redis._addedScriptHashes[scripts[i].sha] = true;
+      }
+      return execPipeline();
+    });
 
   function execPipeline() {
     let data = "";
--- a/test/functional/pipeline.ts
+++ b/test/functional/pipeline.ts
@@ -320,6 +320,28 @@ describe("pipeline", function () {
         });
       });
     });
+
+    it("should support parallel script execution", function (done) {
+      const random = `${Math.random()}`;
+      const redis = new Redis();
+      redis.defineCommand("something", {
+        numberOfKeys: 0,
+        lua: `return "${random}"`,
+      });
+      Promise.all([
+        redis.multi([["something"]]).exec(),
+        redis.multi([["something"]]).exec(),
+      ])
+        .then(([[first], [second]]) => {
+          expect(first[0]).to.equal(null);
+          expect(first[1]).to.equal(random);
+          expect(second[0]).to.equal(null);
+          expect(second[1]).to.equal(random);
+          redis.disconnect();
+          done();
+        })
+        .catch(done);
+    });
   });
 
   describe("#length", function () {


lib/pipeline.ts
```diff
@@ -313,12 +313,15 @@ Pipeline.prototype.exec = function (callback: CallbackFunction) {
 
     const script = this._shaToScript[item.args[0]];
 
-    if (!script || this.redis._addedScriptHashes[script.sha]) {
+    if (
+      !script ||
+      this.redis._addedScriptHashes[script.sha] ||
+      scripts.includes(script)
```
marcbachmann, line 319: We might  want to replace the `.includes` with an object lookup or set.

lib/pipeline.ts
```diff
@@ -352,6 +355,11 @@ Pipeline.prototype.exec = function (callback: CallbackFunction) {
         })
       );
     })
+    .then(function (shas) {
```
marcbachmann, line 358: This also needs to be added to the cluster load promise chain

lib/pipeline.ts
```diff
@@ -330,7 +333,13 @@ Pipeline.prototype.exec = function (callback: CallbackFunction) {
   if (this.isCluster) {
     return pMap(scripts, (script) => _this.redis.script("load", script.lua), {
       concurrency: 10,
-    }).then(execPipeline);
+    })
+      .then(function () {
```
mcollina, line 337: I would combine those two lines

lib/pipeline.ts
```diff
@@ -330,7 +333,13 @@ Pipeline.prototype.exec = function (callback: CallbackFunction) {
...
+        for (let i = 0; i < scripts.length; i++) {
+          _this.redis._addedScriptHashes[scripts[i].sha] = true;
+        }
+      })
```
mcollina, line 341: This promise is not strictly needed so I would just do a `return execPipeline()` at its end.

mcollina (approved):
lgtm

luin:
Awesome! Thanks for the contributions!

> We could also change the behavior by assigning a promise and then reuse that in other requests.
Also doable in another PR if desired.

Functionality-wise that sounds good to me. Not sure if it worth the effort (ex considering the maintainability) though as it anyway only affects cases without cache. However, PR will be welcome if you see it as appropriate. Thanks again!

PR #1312, opened 2021-03-28
use tls.ConnectionOptions instead of tls.SecureContextOptions

bwalendz:
Change TLS options to be consistent across the library and align with `tls.connect()` parameter types.

Allows sentinel connections to use all available TLS options. Right now the typings won't allow you to use the full range of TLS options for `sentinelTLS` that `tls.ConnectionOptions` provides.

**Currently:**

```typescript
const redis = new ioredis({
  sentinels: [{
    host: 'localhost',
    port: 26379,
  }],
  name: 'mymaster',
  sentinelTLS: {
    servername: 'localhost', // <- NOT ALLOWED, type error
  },
});
```

```typescript
const redis = new ioredis({
  host: 'localhost',
  port: 6379,
  tls: {
    servername: 'localhost', // <- ALLOWED
  },
});
```

`ConnectionOptions` is used by `tls.connect()`

```typescript
function connect(options: ConnectionOptions, secureConnectListener?: () => void): TLSSocket;
```

`ConnectionOptions` extends `SecureContextOptions`, `CommonConnectionOptions`

```typescript
interface ConnectionOptions extends SecureContextOptions, CommonConnectionOptions
```

### DIFF
--- a/lib/connectors/SentinelConnector/index.ts
+++ b/lib/connectors/SentinelConnector/index.ts
@@ -6,7 +6,7 @@ import {
   sample,
   Debug,
 } from "../../utils";
-import { connect as createTLSConnection, SecureContextOptions } from "tls";
+import { connect as createTLSConnection, ConnectionOptions } from "tls";
 import {
   ITcpConnectionOptions,
   isIIpcConnectionOptions,
@@ -42,7 +42,7 @@ export interface ISentinelConnectionOptions extends ITcpConnectionOptions {
   preferredSlaves?: PreferredSlaves;
   connectTimeout?: number;
   enableTLSForSentinelMode?: boolean;
-  sentinelTLS?: SecureContextOptions;
+  sentinelTLS?: ConnectionOptions;
   natMap?: INatMap;
   updateSentinels?: boolean;
 }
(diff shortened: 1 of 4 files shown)


luin:
Hey Brian, thanks for the contribution!

Can you elaborate a little more on the use cases that need to pass options other than ones defined in `SecureContextOptions`?

I tend to prevent users from passing `CommonConnectionOptions` via `sentinelTLS` as they may accidentally pass `port`/`host` which should be dynamic and resolved from Redis sentinel servers. Hard coding these dynamic options will make it not be able to connect to the server.

If `servername` is the only exception maybe we can just allow it?

bwalendz:
Ideally the sentinel tls options and redis tls options should be independent and controllable. Whitelisting a few common fields might be acceptable to not break existing configs.

The use case is you cannot connect to a TLS terminated reverse proxy for sentinel when its CA is unverified. So there are a few options you need to allow that like `servername` or `rejectUnauthorized` or `SNICallback`, etc for the sentinel connection. Then the redis connection can be controlled how it wants as well, like a non-TLS connection or more advanced options. I assume most use cases is TLS on for both redis and sentinel but would be useful to be able to control each independently.

Issue #1314, opened 2021-03-29
Sentinel failover not detected when connection hangs


mjomble:
I have stumbled onto the same issue as these:

* https://github.com/luin/ioredis/issues/556
* https://github.com/luin/ioredis/issues/1021
* https://github.com/luin/ioredis/issues/1059

After some investigating, I concluded that ioredis currently relies on Redis closing the connection as described [here](https://redis.io/topics/sentinel-clients#sentinel-failover-disconnection)

However, when the failover is initiated with the Redis `DEBUG SLEEP` command or `docker pause`, the connection simply hangs, but doesn't terminate.

This could be solved by subscribing to sentinel messages on the `+switch-master` channel. Described in the [Sentinel docs](https://redis.io/topics/sentinel) as "the message most external users are interested in"

I've created a reproducible example: https://github.com/mjomble/ioredis-sentinel-issue
This example listens to the message outside ioredis.
Once received, it uses internal/undocumented fields to call `client.connector.stream.destroy()` because `redis.disconnect(true)` (which calls `stream.end()`) leaves the connection open in this scenario.

Ideally, this could all happen inside ioredis. I could probably submit a PR if needed.

luin:
Hi Andres, thanks for raising this and the detailed explanation! A PR would be welcome.

mjomble:
@luin I'm thinking of two different approaches here:

1. Listen to the `+switch-master` message on all sentinels. 
2. Listen to the message only on the first successfully reached sentinel (the same one used to resolve the master).

The first one seems more reliable, but could get complicated when some sentinels are up and some are down. Also, it might not work well with the existing SentinelIterator logic.
The second one should integrate more naturally with SentinelIterator, but might miss the message if the sentinel hangs as well. Which can happen easily if it's on the same machine as the master. Maybe we should add a ping loop to it?

Any thoughts/recommendations?

luin:
Yeah it's more complex than I thought. I'd go with subscribing to all sentinels because in case of a partition, as you said, it's very likely that the connected sentinel and the old master are in the same network, so we won’t be able to get events from that sentinel.

As for implementation details, I think SentinelConnector will get a lastActiveSentinel property, which defaults to null. Once a node is resolved, the connector will subscribe to all sentinels provided by user (I don’t think it need to be dynamic in v1 as it seems non-trivial to implement). Not sure if it’s necessary but a reasonable connection count limit may be applied to avoid user provides too many sentinels.

[...]

PR #1315, opened 2021-03-30
Use async/await

mjomble:
@luin I started working on the PR for #1314, but I'm finding the callback-based code a little difficult to work with.
I noticed that the async/await syntax is not used anywhere in this project.
Is there a reason for not using it? Or would you accept a PR that converts some of the code to this syntax?
I've included a small change in this draft PR, but I'll convert more if I get a green light 🙂


luin:
hi @mjomble, thanks for the work!

ioredis still supports Node.js 6 (yeah I know it's pretty outdated...) so async function can't be used otherwise it will introduce a breaking change to codebases that uses v6.

We do have a plan for the next major version of ioredis with a batch of breaking changes, which will definitely include dropping the support of v6 (or maybe also v8...), so we should do the conversion at that point.

mjomble:
With the current TypeScript configuration (target: es6) the async/await code gets transpiled into generator functions that are also supported in Node.js 6, so it should be fine.

To be sure, I ran the entire ioredis test suite on Node v6 and all tests passed (after I downgraded mocha to 6.2.3 since mocha 7 doesn't support Node 6 either).

mjomble:
I pushed a few more commits. These also passed all tests on Node v6.

luin:
Yeah that makes sense then. My only concern is whether the generated generators have any negative effects in terms of like performance or bundle size, especially given we're going to release a new major version soon so Node.js 6 won't be supported.


I'll take a look at this PR this week. Good catch for the mocha compatible issue btw!

mjomble:
> My only concern is whether the generated generators have any negative effects in terms of like performance or bundle size

The transpiled SentinelConnector/index.js file did grow from 10205 to 10747 bytes in my local build. However, I assume this is a server-side library that doesn't normally get bundled into browser bundles.

As for performance, I'm not sure. But in any case, the current changes are limited to SentinelConnector. I imagine it could theoretically only cause performance problems if there were hundreds or thousands of disconnects and reconnects per second. Whereas sentinels typically wait several seconds before declaring a master unreachable.
A more realistic scenario with lots of rapid connector calls is automatic tests, which seemed to run equally fast with or without these changes.

luin:
Fair enough. I just did a quick diff to the generated code and in general it lgtm. Will take a close look this week.

PR #1316, opened 2021-03-30
Added Node.js v6 to Travis conf, downgraded mocha to 6.2.3

mjomble:
As Node.js v6 is still officially supported, it should probably be tested against in Travis.

As far as I know, the only blocker is mocha 7, which dropped Node 6 support. However, mocha 6 seems to be working fine with this project as well. I guess we'll see what happens when Travis checks this PR.


mjomble:
@luin Got it green on Node 6 now 🙂

PR #1320, opened 2021-04-01
Added timeoutPerRequest option

### LINKED ISSUE #61: Retries per request limit
I'm migrating to ioredis from node_redis and cannot find max_attempts feature analog, is this about me bad in searching? 
It's limit for retries per each request after which it will just fail.

I can implement it on my own as a wrapper but I need to stop requests so that they will not continue to try requesting and this feature is also missing.

mjomble:
Closes #61

Based on the fact that this feature was discussed back in 2015 (#61) and is still not implemented, I'm assuming the implementation in this PR is a very naive one and there's a bunch of special cases it's not covering. But maybe it can still get the ball rolling 🙂

@luin you've said that "It's not easy to support timeoutPerRequest option efficiently". Could you elaborate on this?
Is `setTimeout` inefficient? Or would we need to add a bunch of additional logic to this PR that would slow things down?


luin:
> you've said that "It's not easy to support timeoutPerRequest option efficiently". Could you elaborate on this?

Hmm...good question. I thought having hundreds of timers would cause performance impact, but seems it isn't the case. Another reason I didn't add this feature was I thought it could be easily implemented on client side via ex Promise.race, and I didn't want to mislead users to think that the timeout was applied on the server side (like a Lua script can't be interrupted if it timed out). However, as users pointed out that if they used third-party libraries they won't be able to wrap Promise.race so it may be worth to add this feature.

mjomble:
I now noticed that the other timeout options follow a consistent naming pattern:

* connectTimeout
* disconnectTimeout
* slotsRefreshTimeout

so I renamed this option to `commandTimeout`.
I also added a `sentinelCommandTimeout` option that's passed on to the internally created Redis client for sentinels.

I've also documented `commandTimeout`, but I noticed that no sentinel-specific options have currently been documented, so I left `sentinelCommandTimeout` undocumented for now. Perhaps the sentinel options could be documented all at once in a separate PR.

luin:
Awesome! Great work Andres!

varungbt:
From the code I see that when a command timeout occurs, the promise is rejected. Is there a way to listen for command timeout events?

PR #1328, opened 2021-04-08
Detect failover from +switch-master messages

### LINKED ISSUE #1314: Sentinel failover not detected when connection hangs

mjomble:
Fixes #1314

While working on this, I forgot about the `lastActiveSentinel` suggestion from the other issue. At this point, I'm not sure if it's needed, but I can still add it if it is.

I've tested this with various failure scenarios in a Docker cluster and it successfully reconnected in all of them. I also simulated many of the scenarios in new test cases.

### DIFF
--- a/lib/connectors/SentinelConnector/index.ts
+++ b/lib/connectors/SentinelConnector/index.ts
@@ -1,3 +1,4 @@
+import { EventEmitter } from "events";
 import { createConnection } from "net";
 import { INatMap } from "../../cluster/ClusterOptions";
 import {
@@ -12,10 +13,11 @@ import {
   isIIpcConnectionOptions,
 } from "../StandaloneConnector";
 import SentinelIterator from "./SentinelIterator";
-import { ISentinelAddress } from "./types";
+import { IRedisClient, ISentinelAddress, ISentinel } from "./types";
 import AbstractConnector, { ErrorEmitter } from "../AbstractConnector";
 import { NetStream } from "../../types";
 import Redis from "../../redis";
+import { FailoverDetector } from "./FailoverDetector";
 
 const debug = Debug("SentinelConnector");
 
@@ -39,6 +41,7 @@ export interface ISentinelConnectionOptions extends ITcpConnectionOptions {
   sentinelPassword?: string;
   sentinels: Array<Partial<ISentinelAddress>>;
   sentinelRetryStrategy?: (retryAttempts: number) => number | void | null;
+  sentinelReconnectStrategy?: (retryAttempts: number) => number | void | null;
   preferredSlaves?: PreferredSlaves;
   connectTimeout?: number;
   disconnectTimeout?: number;
@@ -47,11 +50,14 @@ export interface ISentinelConnectionOptions extends ITcpConnectionOptions {
   sentinelTLS?: ConnectionOptions;
   natMap?: INatMap;
   updateSentinels?: boolean;
+  sentinelMaxConnections?: number;
 }
 
 export default class SentinelConnector extends AbstractConnector {
   private retryAttempts: number;
+  private failoverDetector: FailoverDetector | null = null;
   protected sentinelIterator: SentinelIterator;
+  public emitter: EventEmitter | null = null;
 
   constructor(protected options: ISentinelConnectionOptions) {
     super(options.disconnectTimeout);
@@ -82,6 +88,14 @@ export default class SentinelConnector extends AbstractConnector {
       this.sentinelIterator.reset(true);
     }
     return roleMatches;
+  }
+
+  public disconnect(): void {
+    super.disconnect();
+
+    if (this.failoverDetector) {
+      this.failoverDetector.cleanup();
+    }
   }
 
   public connect(eventEmitter: ErrorEmitter): Promise<NetStream> {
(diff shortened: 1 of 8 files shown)


lib/connectors/SentinelConnector/index.ts
```diff
@@ -39,6 +41,7 @@ export interface ISentinelConnectionOptions extends ITcpConnectionOptions {
   sentinelPassword?: string;
   sentinels: Array<Partial<ISentinelAddress>>;
   sentinelRetryStrategy?: (retryAttempts: number) => number | void | null;
+  sentinelReconnectStrategy?: (retryAttempts: number) => number | void | null;
```
mjomble, line 44: Not sure about this name, could be confusing.

But I'm planning to submit a PR that documents all the sentinel options after this gets merged, so that might help.

lib/connectors/SentinelConnector/index.ts
```diff
@@ -269,19 +294,108 @@ export default class SentinelConnector extends AbstractConnector {
...
+  }
+
+  private async activateFailoverDetector() {
+    if (!this.failoverDetector) {
+      return;
+    }
+
+    for (const client of this.failoverDetector.getClients()) {
+      // Apply reconnect strategy to sentinels now that we're no longer looking for master
+      client.options.retryStrategy = this.options.sentinelReconnectStrategy;
```
mjomble, line 386: Not too sure about changing options after the client has been initialised, could lead to something breaking in the future.

But I also couldn't think of a better solution at the moment.
During the initial master resolution, `retryStrategy` must be `null` so we can quickly skip unreachable sentinels.
A cleaner solution would be to disconnect and then reconnect with a new `retryStrategy`, but it seems wasteful not to reuse an already established connection.

luin: Yeah, the current implementation looks good to me.

> Not too sure about changing options after the client has been initialised

Generally speaking, we don't support modifying options after the client is initialized. Although most options should take effect and work as expected, there are no tests to cover that. In the next major version, I think we can make options read-only (or most likely just flag them as deprecated so users don't have to update their existing working code) and provide setters to change options that we support in an explicit way.

> It seems wasteful not to reuse an already established connection.

Connecting to Redis should be pretty lightweight so I think the right behavior and maintainability & readability of the code should take priority over reusing existing connections.

lib/connectors/SentinelConnector/types.ts
```diff
@@ -1,4 +1,32 @@
...
+  ): void;
+  on(event: "error", callback: (error: Error) => void): void;
+  on(event: "reconnecting", callback: () => void): void;
+  disconnect(): void;
+}
+
+export interface ISentinel {
+  address: Partial<ISentinelAddress>;
+  isConnected: boolean;
+  getClient: () => IRedisClient;
```
luin, line 31: We already support the `lazyConnect` option which I think has a similar purpose. With lazyConnect, clients will connect to the server only at the first time they need to send a command. Seems this can save `isConnected` and `getClient`.

lib/connectors/SentinelConnector/index.ts
```diff
@@ -269,19 +294,108 @@ export default class SentinelConnector extends AbstractConnector {
...
+        result = await this.resolveSlave(client);
       } else {
-        return await this.resolveMaster(client);
+        result = await this.resolveMaster(client);
       }
+
+      if (result) {
+        this.initFailoverDetector({
+          address: endpoint,
+          isConnected: true,
```
luin, line 319: As I mentioned in the last comment, the connection should be pretty lightweight (especially given this is a one-timething) so we can just disconnect here and create a new one in `initFailoverDetector` for better code readability. No strong opinion though 😄

luin (commented):
Impressive! The code looks nice! I left two suggestions, lmk your thoughts. Otherwise LGTM!

mjomble:
I made the suggested changes and the code got much simpler 😄
Ready for another review.

luin (approved):

luin:
Thanks @mjomble! I just invited you as a collaborator so from now on you'll be able to approve and merge others' pull requests (which will automatically publish a new version). Feel free to participate anytime when you have time!

Issue #1248, opened 2020-12-05
We got the error of `All keys in the pipeline should belong to the same slots allocation group` when enableAutoPipelining set true


yunnysunny:
The error stack is below:
```javascript
    at Pipeline.exec (/myapp/node_modules/ioredis/built/pipeline.js:250:29)
    at executeAutoPipeline (/myapp/node_modules/ioredis/built/autoPipelining.js:41:14)
    at /myapp/node_modules/ioredis/built/autoPipelining.js:59:13
    at tryCatcher (/myapp/node_modules/standard-as-callback/built/utils.js:11:23)
    at /myapp/node_modules/standard-as-callback/built/index.js:19:49
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```
When update our version to 4.19.2 and set `enableAutoPipelining` true, the error come up. It make us annoying. And then we disabled the `enableAutoPipelining`, the error dismissing.

We also met with an uncaught exception that make the process exited:
```javascript
TypeError: Cannot read property 'Symbol(callbacks)' of undefined
    at Immediate.executeAutoPipeline (/myapp/node_modules/ioredis/built/autoPipelining.js:39:31)
    at processImmediate (internal/timers.js:441:21)
```
It increased our determination to disable `enableAutoPipelining`.

simoneb:
Can you please provide a reproduction for this issue? A code example would work. Thanks

Paul-Isache:
Checked on latest version of ioredis ,4.22.0 down to 4.19.0, with 2 master and 3 slaves, and I am not able to reproduce it.

PR #1335, opened 2021-04-13
fix(cluster): autopipeline when prefix is used

### LINKED ISSUE #1264: autoPipelining true causes "All keys in the pipeline should belong to the same slots allocation group"
Doing some testing with `autoPipelining` turned on on a cluster with 3 master nodes 3 slaves, we're seeing the error `All keys in the pipeline should belong to the same slots allocation group` being thrown.  As far as I can see the slots cache is successfully refreshed and the shards are well balanced.  Any ideas or thoughts?

Thanks for your help!
Mo

kemmerer:
Previously the building of pipelines ignored the key prefix.
It was possible that two keys, foo and bar, might be set into
the same pipeline. However, after being prefixed by a configured
"keyPrefix" value, they may no longer belong to the same
pipeline.

This led to the error:
"All keys in the pipeline should belong to the same slots
allocation group"

This may fix #1264 and #1248.

### DIFF
--- a/lib/autoPipelining.ts
+++ b/lib/autoPipelining.ts
@@ -28,7 +28,10 @@ function findAutoPipeline(
   }
 
   // We have slot information, we can improve routing by grouping slots served by the same subset of nodes
-  return client.slots[calculateSlot(args[0])].join(",");
+  const prefixedKey = client.options.keyPrefix
+    ? `${client.options.keyPrefix}${args[0]}`
+    : args[0];
+  return client.slots[calculateSlot(prefixedKey)].join(",");
 }
 
 function executeAutoPipeline(client, slotKey: string) {
--- a/test/functional/cluster/autopipelining.ts
+++ b/test/functional/cluster/autopipelining.ts
@@ -30,6 +30,10 @@ describe("autoPipelining for cluster", function () {
       if (argv[0] === "get" && argv[1] === "foo6") {
         return "bar6";
       }
+
+      if (argv[0] === "get" && argv[1] === "baz:foo10") {
+        return "bar10";
+      }
     });
 
     new MockServer(30002, function (argv) {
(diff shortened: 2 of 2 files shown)


lib/autoPipelining.ts
```diff
@@ -28,7 +28,10 @@ function findAutoPipeline(
   }
 
   // We have slot information, we can improve routing by grouping slots served by the same subset of nodes
-  return client.slots[calculateSlot(args[0])].join(",");
+  const prefixedKey = client.options.keyPrefix
```
TysonAndre, line 31: This approach seems reasonable.

Looking at lib/commander.ts
- When there is NO pipelining, `new Command` is used to transform args (e.g. add key prefixes) in its constructor
- When there is pipelining, args are passed without any transformation

```js
    // No auto pipeline, use regular command sending
    if (!shouldUseAutoPipelining(this, commandName)) {
      return this.sendCommand(
        new Command(commandName, args, options, callback)
      );
    }

    // Create a new pipeline and make sure it's scheduled
    return executeWithAutoPipelining(this, commandName, args, callback);
```

TysonAndre: Looking at lib/pipeline.ts for the `"All the keys in a pipeline command should belong to the same slot"` I don't see any issues -  that checks if everything would go to the same shard using the prefixed keys (from Command.getKeys, Command.iterateKeys)

- e.g. if a pipeline only had one key, it would use the computed keys `this._iterateKeys((key) => options.keyPrefix + key);` to determine the correct slots, and there would only be an issue if there were multiple keys and those would get sent to different slots if the prefix wasn't added.

I'd guess the new test case would be less prone to similar coincidences in the future if the unit test was called with more distinct keys

TysonAndre:
This can be closed - the changes in this were among the changes merged in https://github.com/luin/ioredis/pull/1393

TysonAndre:
> This can be closed - the changes in this were among the changes merged in 1393

When anyone has time, so that people don't see stale PRs or worry about autopipelining needlessly.

marcbachmann:
closing as requested

PR #1359, opened 2021-05-31
perf: Serialize error stack only when needed

aivopaas:
Improving Error stack serialization performance when `showFriendlyErrorStack=true`.

With current implementation there's about 30% overhead because the stack trace is serialized even when it's not actually needed.

With proposed change the overhead is somewhere between 1-5% on Node 14. There were improvements related to stack trace serialization in Node 12.

On the attached benchmark results, first run is on old code, second run on new code. Benchmark code also committed.

image

### DIFF
--- a/benchmarks/errorStack.ts
+++ b/benchmarks/errorStack.ts
@@ -1,0 +1,38 @@
+import { cronometro } from "cronometro";
+import Redis from "../lib/redis";
+
+let redis;
+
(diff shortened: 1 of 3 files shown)


aivopaas:
Opened issue #1361 for the discussion

luin:
Awesome! Thanks for the contribution!

Issue #1362, opened 2021-06-03
4.27.0 introduced a change that prevents the event loop from being cleaned up properly


evanlucas:
https://github.com/evanlucas/ioredis-sentinel-hang is how I was able to reproduce it. We rely on calling `quit` in all of our jobs/workers/services and if the event loop is not cleaned up, it prevents node from cleanly exiting the process. It looks like it is the new failover detection mechanism that was added. I'm seeing additional TCP sockets being opened, but never closed.

This only seems to occur when using redis sentinel.

Thanks!

AVVS:
confirming, it opens a bunch of new connections that are never cleaned up (see FailoverDetector)

luin:
Hey @evanlucas, thanks for the reproducible repo! Calling `.disconnect()` should properly clean up everything. We should take a look for a fix for `quit`.

evanlucas:
Thanks for jumping on this and getting a non-breaking release published! I really appreciate y'all's time! Did y'all want to leave this issue open? or is it good to close?

luin:
Hey @evanlucas, let's leave this open as the root issue hasn't been fixed.

esatterwhite:
Any progress on this one. It would be fantastic to get this one resolved

juris:
In a Kubernetes cluster, when one of the Sentinels restarts and gets a new IP address, client starts throwing errors like this one:
```
[ioredis] Unhandled error event: Error: connect EHOSTUNREACH 10.0.55.221:26379
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1161:16)
    at TCPConnectWrap.callbackTrampoline (node:internal/async_hooks:130:17)
```
Issue disappears with `failoverDetector: false`

silverwind:
There's definitely something fishy going on when `failoverDetector` is enabled. After a failover, ioredis will start logging `Unhandled error event` to stderr, despite me having registered the `error` event earlier which should prevent all such logging. I think it must be keeping around a partially terminated `redis` instance in memory.

PR #1363, opened 2021-06-05
fix(SENTINEL): disable active failover detection by default

luin:
There is an issue related to connections used for active failover detection for sentinels not being cleaned up properly: #1362. Given that not many users realize the usefulness of this feature, and the feature does create more connections, we disable it by default and allow users to enable it with the option`failoverDetector`.


AVVS:
LGTM. 

Ideally somebody would later tackle graceful shutdown by implementing ‘close’ command where quit is sent and failover detector is also closed

AVVS (approved):

PR #1371, opened 2021-06-21
Updated Docs

metju90:
Documented that if `retry_stategy` is provided, errors are not emitted


metju90:
I've spent half a day trying to figure out why errors are not being emitted. Only to find the following comment in `node_modules/redis/index.js:`

```
 // Only emit the error if the retry_stategy option is not set
    if (!this.options.retry_strategy) {
        this.emit('error', err);
    } 
```

PR #1376, opened 2021-07-01
add RedisReadable and RedisWritable

Uzlopak:
This PR adds RedisReadable and RedisWritable to ioredis

These are convenient classes, which bring Redis Streams to the Node ecosystem.


Uzlopak:
I refactored some code of mine, and have to write some unit tests in your format.

To Be done:

- [ ] write unit tests
- [ ] check if maxlen is correctly refactored
- [ ] check if it can be instantiated correctly (does the export in index.ts is enough?)
- [ ] write documentation

PR #1393, opened 2021-07-28
Fix more edge cases in autopipelining when using redis cluster

### LINKED ISSUE #1392: `mget` with Redis.Cluster and `enableAutoPipelining: true` causes "Error: All the keys in a pipeline command should belong to the same slot"

TysonAndre:
Fixes #1392

This is a superset of the fixes in https://github.com/luin/ioredis/pull/1391

The other issue this PR fixes was that this was passing an array instead of a string to
cluster-key-slot when `args[0]` was an array

Handle pathological cases such as redis.get([], ['key1', 'key2'])
by flattening the entire args array.

Prior to this PR, in the below example, the array was hashed instead of the first string in the array,
meaning that 'v{profile}1' and 'b' were incorrectly put in the same pipeline

```js
const IORedis = require('./built');
const process = require('process');
async function main() {
    // Using instructions from https://redis.io/topics/cluster-tutorial
    // redis-cli --cluster create 127.0.0.1:7000 127.0.0.1:7001 127.0.0.1:7002 --cluster-replicas 0
    const redis = new IORedis.Cluster([{ host: '127.0.0.1', port:7000 }], { enableAutoPipelining: true });
    const results = await Promise.all([
      redis.mget(['v{profile}1', 'v{profile}2', 'v{profile}3', 'v{profile}4']),
      redis.get('b'),
    ]);
    console.log(results);
    process.exit(0);
}
main();
```

### DIFF
--- a/lib/autoPipelining.ts
+++ b/lib/autoPipelining.ts
@@ -1,6 +1,8 @@
 import * as PromiseContainer from "./promiseContainer";
+import * as commands from "redis-commands";
 import * as calculateSlot from "cluster-key-slot";
 import asCallback from "standard-as-callback";
+import { flatten } from "./utils/lodash";
 
 export const kExec = Symbol("exec");
 export const kCallbacks = Symbol("callbacks");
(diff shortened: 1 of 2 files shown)


TysonAndre:
Checking node_modules/redis-commands/commands.json for `keyStart == keyStop` may be a cheap way to detect commands such as `get` which have only one key without needing to flatten a potentially large array, but it seems like the `redis-commands` API doesn't treat commands.json as part of the public API. Not sure if that would be micro-optimizing too much.

PR #1396, opened 2021-07-29
fixbug[cluster/index.ts] GetInfoFromNode slotRangeStart and slotRangeEnd WrongType

thaiwu0107:
GetInfoFromNode slotRangeStart and slotRangeEnd WrongType


luin:
Addressed in https://github.com/luin/ioredis/pull/1504

Issue #1027, opened 2019-12-14
Execute script with caching but no definedCommand


ethanresnick:
I have a class that accepts an `ioredis` instance as a parameter and uses it to make some redis calls. One of those calls requires running a script, and I'd like to be able to take advantage of the logic that first attempts to run the script by SHA, rather than sending the script text each time. However, I don't want my class to have to call `defineCommand` on the ioredis instance it's passed, as it doesn't "own" that instance, and shouldn't be mutating it in ways that are visible from the outside.

As far as I can tell, the Script.ts file that contains the EVALSHA logic is not exported anywhere/not part of this module's public interface. Could that class be exported, so that I can create a script and run it (by passing the instance to script.execute) without having to mutate the ioredis instance (or import the Script class despite it not being part of the public API)?

ethanresnick:
I feel like this should be reopened, as it hasn't been addressed at all (even to say it's invalid/won't be supported). 

Also, @luin, maybe it'd be worth turning down the aggressiveness of the stale bot a little bit? It's somewhat annoying to spend time opening an issue only for a bot to automatically close it (and without a way to reopen it myself).

TysonAndre:
Another use case for this:

Consider an application using a `Proxy<Redis>` or `Proxy<RedisCluster>` in case a config needs to change at runtime (e.g. to change the hostname, or to switch to Redis Cluster, or to change a prefix) (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) 

In that case, when the redis instance belonging to the Proxy is switched to a newly constructed instance, all of the dynamic properties defined with defineCommand would be lost.

> As far as I can tell, the Script.ts file that contains the EVALSHA logic is not exported anywhere/not part of this module's public interface. Could that class be exported, so that I can create a script and run it (by passing the instance to script.execute) without having to mutate the ioredis instance (or import the Script class despite it not being part of the public API)?

Having a low level api would be something I'd personally consider useful for
- third party libraries that prefer avoiding mutating the passed in redis instance
- applications where the redis instance is a Proxy

PR #1404, opened 2021-08-04
Proposal: Make Script and Script.execute a stable api

TysonAndre:
For #1027

Having a stable low level api to evaluate scripts would be something I'd
personally consider useful for:

1. Third party libraries that prefer avoiding mutating the passed in redis instance
2. Applications where the redis instance is a Proxy.

If the concept makes sense to ioredis maintainers, more tests will be added.

Aside: Would adding Commander.getKeyPrefix make sense? Or should this stop passing in the keyPrefix to the Script constructor and use something like `container?.options?.keyPrefix` instead?

### DIFF
--- a/API.md
+++ b/API.md
@@ -13,7 +13,7 @@
 
 ## Redis ⇐ <code>[EventEmitter](http://nodejs.org/api/events.html#events_class_events_eventemitter)</code>
 
-**Kind**: global class  
+**Kind**: global class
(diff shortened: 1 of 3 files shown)


luin:
> Would adding Commander.getKeyPrefix make sense?

What would it look like with this approach? Should `Script` call that method?

TysonAndre:
> What would it look like with this approach? Should `Script` call that method?

Yes, that's what I had in mind so that a script could be called by any Commander and automatically use the right key prefix if they weren't overridden in an options object

PR #1408, opened 2021-08-06
fix(cluster): lazyConnect with pipeline

### LINKED ISSUE #1407: Cluster lazyConnect will not auto connect in pipeline/multi/enableAutoPipelining

jseagull:
this will fix #1407

### DIFF
--- a/lib/pipeline.ts
+++ b/lib/pipeline.ts
@@ -16,7 +16,7 @@ import Commander from "./commander";
 function generateMultiWithNodes(redis, keys) {
   const slot = calculateSlot(keys[0]);
   const target = redis._groupsBySlot[slot];
-  
+
   for (let i = 1; i < keys.length; i++) {
     if (redis._groupsBySlot[calculateSlot(keys[i])] !== target) {
       return -1;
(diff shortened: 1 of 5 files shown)


lib/pipeline.ts
```diff
@@ -241,6 +242,7 @@ Pipeline.prototype.execBuffer = deprecate(function () {
 Pipeline.prototype.exec = function (callback: CallbackFunction) {
   // Wait for the cluster to be connected, since we need nodes information before continuing
   if (this.isCluster && !this.redis.slots.length) {
+    if (this.redis.status === "wait") this.redis.connect();
```
TysonAndre, line 245:
```suggestion
    if (this.redis.status === "wait") this.redis.connect().catch(noop);
```

I think that without the Promise.catch, there may be an UnhandledPromiseRejection that would terminate node processes with a `process.on('unhandledRejection')` listener that terminates.

Looks reasonable otherwise, but I'm not a maintainer

jseagull: Added to all three location.

luin (approved):

luin:
Thanks @jseagull @TysonAndre !

PR #1416, opened 2021-08-18
fix: handle malicious keys for hgetall

### LINKED ISSUE #1267: hgetall mishandles malicious keys (__proto__)
This isn't a global prototype pollution attack, but does mean it's possible to get inconsistent behaviour:

```js
await client.hset('test_key', ['__proto__', 'hello']);
console.log('hget:', await client.hget('test_key', '__proto__'))); // "hello"
console.log('hgetall:', await client.hgetall('test_key'))); // does not include __proto__: hello
```

It's because the reply transformer which is applied does not check for special field names: https://github.com/luin/ioredis/blob/master/lib/command.ts#L404

Since this isn't doing a recursive set, it can't cause global prototype pollution, but it _can_ cause unexpected inconsistent behaviour as shown above. Obviously this is mainly a concern if allowing user-provided field names (e.g. a database explorer UI). Since the database itself can handle this name, I think the client library should be able to handle it too.

The easiest fix would be something like:

```js
    const obj = {};
    for (let i = 0; i < result.length; i += 2) {
      const key = result[i];
      const value = result[i + 1];
      if (obj[key]) { // can only be truthy if the property is special somehow, like '__proto__' or 'constructor'
        Object.defineProperty(obj, key, {
          value,
          configurable: true,
          enumerable: true,
          writable: true,
        });
      } else {
        obj[key] = value;
      }
    }
    return obj;
```

But a better fix would be to change the API to return a `Map` instead.

None of the other commands seem to have the same issue; `hgetall` is the only command which uses `setReplyTransformer`, and the `ArgumentTransformer`s are fine (e.g. `hset('foo', JSON.parse('{"__proto__":"yay"}'))` works)

---

Relatedly, it would be nice to be able to call `hgetall` _without_ the reply transformer (returning a raw key/value alternating list) for situations where the intent is to iterate through all fields anyway. It's interesting that the transformer has [a condition checking for arrays](https://github.com/luin/ioredis/blob/master/lib/command.ts#L401); is it possible for the server to send a different data type? The [main documentation says it will always be a list](https://redis.io/commands/hgetall).

luin:
Closes #1267

### DIFF
--- a/lib/command.ts
+++ b/lib/command.ts
@@ -427,7 +427,20 @@ Command.setReplyTransformer("hgetall", function (result) {
   if (Array.isArray(result)) {
     const obj = {};
     for (let i = 0; i < result.length; i += 2) {
-      obj[result[i]] = result[i + 1];
+      const key = result[i];
+      const value = result[i + 1];
+      if (obj[key]) {
+        // can only be truthy if the property is special somehow, like '__proto__' or 'constructor'
+        // https://github.com/luin/ioredis/issues/1267
+        Object.defineProperty(obj, key, {
+          value,
+          configurable: true,
+          enumerable: true,
+          writable: true,
+        });
+      } else {
+        obj[key] = value;
+      }
     }
     return obj;
   }
--- a/test/functional/hgetall.ts
+++ b/test/functional/hgetall.ts
@@ -1,0 +1,12 @@
+import Redis from "../../lib/redis";
+import { expect } from "chai";
+
+describe("hgetall", function () {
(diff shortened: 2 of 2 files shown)


lib/command.ts
```diff
@@ -427,7 +427,20 @@ Command.setReplyTransformer("hgetall", function (result) {
   if (Array.isArray(result)) {
     const obj = {};
```
silverwind, line 428: Could probably just use `Object.create(null)` here to avoid the performance hit from the additional check below.

Issue #1417, opened 2021-08-18
Idea: Add new `hgetallMap` method to return `Map<String,String>`


TysonAndre:
(and `hgetallMapBuffer: Map<String, Buffer>`)

Motivation:
- Make it easier to write consise code that avoids prototype pollution
- Often, I'd end up converting `Object.entries(await redis.hgetall(key))` to a Map anyway.
  (easier to make callers write `map.has(fieldName)` than `Object.prototype.hasOwnProperty.call(obj, fieldName)`, since `obj[fieldName]` would be wrong for fields that are part of the object prototype)

Related to #1267 

(Aside: would `key in obj` work better than `obj[key]` for the check if the property was already defined, in the PR fixing that issue for `hgetall` (to account for `Object.prototype.customProperty = false;`)? Otherwise, the PR looked correct.)

luin:
> (and hgetallMapBuffer: Map<String, Buffer>)

I definitely think it's useful to add it. However, I'm not sure if this is the right time to add it. For now, although ioredis is 100% written in TypeScript, many types are missing (ex there are no typings for `Redis#get`). I think we should solve the typing issue first and then add more methods to Redis/Commander since the process of adding types may result in breaking changes.

> (Aside: would key in obj work better than obj[key] for the check if the property was already defined...
Good point! I didn't think of that. Can you take a look https://github.com/luin/ioredis/pull/1418 ?

PR #1418, opened 2021-08-18
fix: improve proto checking for hgetall [skip ci]

luin:
As mentioned in https://github.com/luin/ioredis/issues/1417

### DIFF
--- a/test/functional/hgetall.ts
+++ b/test/functional/hgetall.ts
@@ -32,5 +32,6 @@ describe("hgetall", function () {
     expect(Object.keys(ret).sort()).to.eql(
       ["__proto__", CUSTOM_PROPERTY].sort()
     );
+    expect(Object.getPrototypeOf(ret)).to.eql(Object.prototype);
   });
 });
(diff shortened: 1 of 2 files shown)


test/functional/hgetall.ts
```diff
@@ -1,12 +1,36 @@
...
+      CUSTOM_PROPERTY,
+      "world"
+    );
     const ret = await redis.hgetall("test_key");
     expect(ret.__proto__).to.eql("hello");
-    expect(Object.keys(ret)).to.eql(["__proto__"]);
+    expect(ret[CUSTOM_PROPERTY]).to.eql("world");
+    expect(Object.keys(ret).sort()).to.eql(
+      ["__proto__", CUSTOM_PROPERTY].sort()
+    );
```
TysonAndre, line 34: Unrelated to this change: Could also check that `expect(Object.getPrototypeOf(ret)).to.eql(Object.prototype);` to confirm the original prototype isn't overridden, though the Object.keys() check pretty much implies that

TysonAndre (commented):
lgtm, documentation of https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in#parameters `in` operator is what I'd expect it to be

PR #1420, opened 2021-08-24
use null prototype instead of additional check

silverwind:
Followup on https://github.com/luin/ioredis/pull/1416, this is a more efficient way of fixing the issue by creating a object that does not inherit anything.


silverwind:
Stricly speaking this is an incompatible change as null objects behave slightly different then "regular" objects in some regards, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create#custom_and_null_objects.

PR #1430, opened 2021-09-03
Adding ability to use as uncached dependency

mmedvedik:
Mostly for unit tests with mocks.
ie: with proxyquire noPreserveCache()


luin:
@mmedvedik 👋  Sorry for the late response. ioredis v5 already dropped the support of customized Promise implementation, so closing this. Thanks for the contributions anyway!

PR #1466, opened 2021-11-23
fix: unhandled Promise rejections in pipeline.exec [skip ci]

TysonAndre:
1. Always return the same Promise in pipeline.exec.

   An example of why the original Promise should be returned, rather than
   a different Promise resolving to the original Promise:

   ```javascript
   process.on('unhandledRejection', (reason) => console.log('unhandledRejection', reason));
   const x = Promise.reject(new Error());
   x.catch((e) => console.log('caught x', e));

   const causesUnhandledRejection = Promise.resolve().then(() => x);
   ```
2. In the negligible chance the `script exists`/`script load` command
   failed, give up, catch the rejected Promise with `finally`,
   and try to run the commands in the pipeline anyway.
   (e.g. networking issues, corrupted bytes, misbehaving proxy/server, etc)

   This isn't the best approach, but it's hopefully better than an unhandled Promise rejection,
   resource/memory leak, or hanging request due to a redis Promise never
   resolving or rejecting for an individual command in this edge case.

   **If the maintainers have an idea for a better approach (e.g. rejecting requests in the pipeline with a single instance of a new error type) I'd be open to it**
3. There are calls to `this.exec(...);` all over the Pipeline that don't
   check if the returned Promise is caught which I believe could lead to
   unhandled Promise rejections on retrying failed commands.

   Also, lib/autopipelining.js also called pipeline.exec without
   checking for errors.

   The fact there is now exactly one Promise instance that can be returned
   means that this no longer can cause unhandled Promise rejections.
   (see example snippet in commit description)
   (`standard-as-callback` is catching rejections)
4. pipeline.exec can explicitly call reject for ioredis's Redis.Cluster
   client

----

**NOTE: This is theoretical. Due to the async nature of stack traces returning the stack trace of the place where the Error instance was created, the only stack traces I've seen are in redis-parser.**
I've also seen unhandled rejections when loading scripts in an application using pipelining and Redis.Cluster, but I haven't figured out if this is the cause of the issues in that specific application

### DIFF
--- a/lib/pipeline.ts
+++ b/lib/pipeline.ts
@@ -240,7 +240,15 @@ Pipeline.prototype.execBuffer = deprecate(function () {
   return execBuffer.apply(this, arguments);
 }, "Pipeline#execBuffer: Use Pipeline#exec instead");
 
-Pipeline.prototype.exec = function (callback: CallbackFunction) {
+// NOTE: To avoid an unhandled promise rejection, this will unconditionally always return this.promise,
+// which always has the rejection handled by standard-as-callback
+// adding the provided rejection callback.
+//
+// If a different promise instance were returned, that promise would cause its own unhandled promise rejection
+// errors, even if that promise unconditionally resolved to **the resolved value of** this.promise.
+Pipeline.prototype.exec = function (
+  callback: CallbackFunction
+): Promise<Array<any>> {
   // Wait for the cluster to be connected, since we need nodes information before continuing
   if (this.isCluster && !this.redis.slots.length) {
     if (this.redis.status === "wait") this.redis.connect().catch(noop);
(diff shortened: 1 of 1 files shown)


lib/pipeline.ts
```diff
@@ -338,17 +346,19 @@ Pipeline.prototype.exec = function (callback: CallbackFunction) {
...
-      }
-      return execPipeline();
-    });
+    })
+      .then(function () {
+        for (let i = 0; i < scripts.length; i++) {
+          _this.redis._addedScriptHashes[scripts[i].sha] = true;
+        }
+      })
+      .then(execPipeline, execPipeline);
```
luin, line 357: I think this will swallow errors thrown from `pMap` (as it doesn't work exactly as `.finally(execPipeline))`.

Should it be `.then(execPipeline, this.reject);`?

lib/pipeline.ts
```diff
@@ -371,8 +381,9 @@ Pipeline.prototype.exec = function (callback: CallbackFunction) {
       for (let i = 0; i < scripts.length; i++) {
         _this.redis._addedScriptHashes[scripts[i].sha] = true;
       }
-      return execPipeline();
-    });
+    })
+    .then(execPipeline, execPipeline);
```
luin, line 385: Same as above, should be `.then(execPipeline, this.reject);` IMO

PR #1468, opened 2021-11-23
Start testing with Node 16 LTS

TysonAndre:
https://docs.travis-ci.com/user/languages/javascript-with-nodejs/ provides node 16, and node 16 LTS is out.


TysonAndre:
@luin - Some of these are subtle timing issues where I'm not familiar with the implementation. Would it make sense to introduce tests for node 16 in CI, but skip it if process.versions.node >= 16.0.0 with a FIXME comment

luin:
> @luin - Some of these are subtle timing issues where I'm not familiar with the implementation. Would it make sense to introduce tests for node 16 in CI, but skip it if process.versions.node >= 16.0.0 with a FIXME comment

Yeah I think so.

luin:
> The build passes now but travis hasn't run for some reason.

Travis has a credit thing which has run out and will be refilled tomorrow. Considering moving to GitHub CI but anyway it should work tomorrow.

Issue #1405, opened 2021-08-04
Custom Lua scripts are not loaded in cluster mode


stavalfi:
I created a local cluster on my macbook pro with 3-8 nodes using redis-cli and redis-server (From homebrew).

When I shutdown all nodes and re-create them (manually) - 9/10 runs this bug occurs:

```
const client = new IORedis.Cluster([
  {
    host: 'localhost',
    port: 7000,
  },
])

const scriptName = 'script1'
const script1 = 'return {KEYS[1]}'
client.defineCommand(scriptName, { numberOfKeys: 1, lua: script1 })
console.log(await client[scriptName]('aaaa', 'hi', console.log))
```

# Error

```
(node:54658) UnhandledPromiseRejectionWarning: ReplyError: NOSCRIPT No matching script. Please use EVAL.
    at parseError (/Users/stavalfi/projects/singer/.yarn/cache/redis-parser-npm-3.0.0-7ebe40abcb-45dbcb05be.zip/node_modules/redis-parser/lib/parser.js:179:12)
    at parseType (/Users/stavalfi/projects/singer/.yarn/cache/redis-parser-npm-3.0.0-7ebe40abcb-45dbcb05be.zip/node_modules/redis-parser/lib/parser.js:302:14)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:54658) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:54658) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

# (Very?) Bad Workaround

It took me a while to figure out how to fix it: I need to manually update all nodes about the script

```
const masters = client.nodes('master')
await Promise.all(
  masters.map(node => Promise.all([node.script('load', script1), node.script('load', persistMollyLiveBmss)])),
)
```

I'm not sure I can use this fix in production.

---------

# More info

Note: In non-cluster runs, all works great: using local,remote,docker redis-servers.

```
System:
    OS: macOS 10.15.1
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
[...]

PR #1497, opened 2022-01-31
Fix script loading behavior on script flush with pipelines

marcbachmann:
Lua scripts get loaded once and not retried unless a client side interval expires: https://github.com/luin/ioredis/blob/master/lib/pipeline.ts#L376

This causes issues with redis instances restarting or failing over.
There's no way of knowing whether a script is still present in the new instance.
We should reset the local state and just load them again when a reconnect happens and rely on the existing code that checks whether the scripts are present on the server.

~I ran into that issue while testing the reliability of a service, but it might also fix https://github.com/luin/ioredis/issues/1405~ See #1499 for the cluster fixes.

There's still a small time window where `NOSCRIPT` errors occur as messages are already queued for execution. ~Previously the script handler only executed `EVAL` in the non-pipeline mode. This now changed and always gets executed.~

### Changelog
- 🐛 Reset loaded script hashes to force a reload of scripts after reconnect of redis

### DIFF
--- a/test/functional/pipeline.ts
+++ b/test/functional/pipeline.ts
@@ -353,6 +353,44 @@ describe("pipeline", function () {
         })
         .catch(done);
     });
+
+    it("should reload scripts on redis restart (reconnect)", async function () {
+      const redis = new Redis({ connectionName: "load-script-on-reconnect" });
+      const redis2 = new Redis();
+      redis.defineCommand("exeecafterreconnect", {
+        numberOfKeys: 0,
+        lua: `return "OK"`,
+      });
+
+      const [[err, res]] = await redis.multi([["exeecafterreconnect"]]).exec();
+      expect(err).to.equal(null);
+      expect(res).to.equal("OK");
+
+      const client = await redis.client("list").then((clients) => {
+        const myInfo = clients
+          .split("\n")
+          .find((client) => client.includes("load-script-on-reconnect"));
+
+        const match = / addr=([^ ]+)/.exec(myInfo);
+        if (match) return match[1];
+      });
+
+      await redis2.script("flush");
+      await redis2.client("kill", "addr", client);
+
+      // Wait for reconnect, at the moment scripts are not loaded
+      // if the pipeline starts before ioredis reconnects
+      await redis.ping();
+
+      const [[err2, res2]] = await redis
+        .multi([["exeecafterreconnect"]])
+        .exec();
+
+      expect(err2).to.equal(null);
+      expect(res2).to.equal("OK");
+      redis.disconnect();
+      redis2.disconnect();
+    });
   });
 
   describe("#length", function () {
--- a/lib/redis/index.ts
+++ b/lib/redis/index.ts
@@ -316,6 +316,10 @@ Redis.prototype.connect = function (callback) {
     // Make sure only one timer is active at a time
     clearInterval(this._addedScriptHashesCleanInterval);
 
+    // Scripts need to get reset on reconnect as redis
+    // might have been restarted or some failover happened
+    this._addedScriptHashes = {};
+
     // Start the script cache cleaning
     this._addedScriptHashesCleanInterval = setInterval(() => {
       this._addedScriptHashes = {};


test/functional/pipeline.ts
```diff
@@ -353,6 +353,47 @@ describe("pipeline", function () {
...
+          .find((client) => client.includes("load-script-on-reconnect"));
+
+        const match = / addr=([^ ]+)/.exec(myInfo);
+        if (match) return match[1];
+      });
+
+      await redis2
+        .pipeline([
+          ["script", "flush"],
+          ["client", "kill", "addr", client],
```
marcbachmann, line 381: I'll seperate those calls, so the error here is handled.

marcbachmann:
Regarding the tests, maybe listing the redis version during the docker run could be useful.

Somehow the node 6 tests failed before some small fix.
Probably because it uses redis 3.2.6 https://packages.debian.org/stretch/redis-server

But that should have worked with the [`CLIENT KILL ID clientId` and `CLIENT KILL addr:port` commands](https://redis.io/commands/client-kill).

artur-ma:
Isnt this case is handled by this code? 
https://github.com/luin/ioredis/blob/b8177479c348aa4bbd467fa944d61fe9b35aec19/lib/script.ts#L42

If the script is missing then we will get the NOSCRIPT error, which will force ioredis to load it.

marcbachmann:
Not in pipeline mode. That gets skipped as the result of sendCommand is no promise.
`await redis.pipeline(['something']).exec()`

```js
    const result = container.sendCommand(evalsha);
    // This is false, result is an instance of Pipeline
    if (isPromise(result)) {
```

marcbachmann:
I did another commit to handle the NOSCRIPT errors in https://github.com/luin/ioredis/pull/1497/commits/53472a7f269c40587dfdadd76ee7e956ef8bd8fe

Resetting the scripts is still useful with this as without it, it would execute `EVAL` all the time, which is quite expensive.

marcbachmann:
I'm still getting `NOSCRIPT` errors even with the changes above. 🤔

lib/script.ts
```diff
@@ -34,26 +33,21 @@ export default class Script {
 
     const evalsha = new Command("evalsha", [this.sha].concat(args), options);
     evalsha.isCustomCommand = true;
+    evalsha.promise = evalsha.promise.catch((err: Error) => {
+      if (err.toString().indexOf("NOSCRIPT") === -1) {
+        throw err;
+      }
+      const command = new Command("eval", [this.lua].concat(args), options);
```
marcbachmann, line 40: Replacing that with a `SCRIPT LOAD` would be another task.
Does this work in cluster mode?
```js
redis._addedScriptHashes[this.sha] = true
return redis.pipeline(['script', 'load', this.lua], ['evalsha', this.sha]])
  .exec()
  .then(([,res]) => res)
```

artur-ma:
> Resetting the scripts is still useful with this as without it, it would execute EVAL all the time, which is quite expensive.

IMO this is a serious statement, we have really high load on our services with a lot of lua scripts(we use redis with redisLabs as our main persistent DB), from time to time we loose connection so we have to reconnect. if it will cause sending the script string each time and run EVAL, it may be a disaster  :)
also it will be hard to understand where this extra latency comes from after reconnect.

marcbachmann:
Yes. This PR now fixes this behavior. New commands coming in will first load the script again. And at the moment it always does an exists check: https://github.com/luin/ioredis/blob/f275bc24de3825f80415a69ff227a45251dd1a3b/lib/pipeline.ts#L282

With the update, we only have the issue that already queued messages will do an eval. Somewhat like https://github.com/luin/ioredis/pull/1497#discussion_r795706271 would fix this.

lib/script.ts
marcbachmann: I'd have other fixes that uses pipeline-based script loading: https://github.com/marcbachmann/ioredis/compare/fix-redis-reconnect-NOSCRIPT-errors...marcbachmann:pipelined-evalsha?expand=1
As it's a bit bigger, and I don't like to merge it into this pr.

The only disadvantage is that it doesn't use `script exits` checks anymore and just uses `script load` after the interval expired. I guess redis already hashes by sha internally and only loads the script if it's not present. So that should be fine.
But as there's no roundtrip anymore and script caching time can be set much higher with it. NOSCRIPT errors are also retried in there (should also work better for the cluster support).

marcbachmann: This is now ready in https://github.com/luin/ioredis/pull/1499
@artur-ma want to try? 😁

luin:
Hey @marcbachmann 👋 

Thanks for the contribution. I didn't realize this issue so it's a great fix!

The reason we don't retry on `NOSCRIPT` error in pipeline mode is we want to ensure the command order, so for example, in `.pipeline().mycustom('foo').set('foo', 'bar').exec()`, `mycustom` shouldn't be executed after the `set` command. Breaking this rule results in a breaking change, and I don't think we should include this change in a bug fix PR.

> Resetting the scripts is still useful with this as without it, it would execute EVAL all the time, which is quite expensive.

Not sure if I understand this, why not resetting the scripts results in `EVAL` being executed all the time? Won't it be always `EVALSHA` first and then `EVAL` when failed? So only first several commands will execute `EVAL` in a small time window.

#1499 looks promising, as I still need some time reviewing it, for this PR, should we just delete the `_addedScriptHashes` flag for the sha in case of `NOSCRIPT` error so following custom commands in pipeline mode should work?

marcbachmann:
You're welcome. Yes, then let's remove the fallback logic again and simplify that PR. #1499 might be a better approach to handle the script caches.

> The reason we don't retry on NOSCRIPT error in pipeline mode is we want to ensure the command order, so for example, in .pipeline().mycustom('foo').set('foo', 'bar').exec(), mycustom shouldn't be executed after the set command. Breaking this rule results in a breaking change, and I don't think we should include this change in a bug fix PR.

Good to know, not sure this was mentioned anywhere.
That could make sense in some cases. But redis doesn't cancel commands followed after a `NOSCRIPT` of the evalsha command. Therefore multiple commands that belong together will always be an issue, even when they are in the same pipeline or within a `MULTI` transaction.
With that in mind, I'm not sure that is the better approach. I guess it completely depends on the use-case.
E.g. I'm doing an `xack` and some other script in the same pipeline. If the second script never executes, I'll have a worse state as the exactly-once delivery isn't guaranteed for those commands that failed.

> Won't it be always EVALSHA first and then EVAL when failed? So only first several commands will execute EVAL in a small time window.

Yes, but it will execute the `EVAL` for all commands in the offline queue. Mainly because `EVAL` is queued **after** all the other commands already the queue. With #1499 we'll have some trade off of loading the script once on every reconnect, but will resolve those issues.

> for this PR, should we just delete the _addedScriptHashes flag for the sha in case of NOSCRIPT error so following custom commands in pipeline mode should work?

👍

marcbachmann:
This is now ready to merge

luin:
> With that in mind, I'm not sure that is the better approach. I guess it completely depends on the use-case.

Fair point. I 💯 agree with that. I think we should then just priority code maintainability over the behavior choice, though introducing breaking changes means we have to wait for the schedule of the next major version.

PR #1499, opened 2022-02-01
Pipeline-based script loading and retries

marcbachmann:
This PR builds on https://github.com/luin/ioredis/pull/1497 (first three comments)
The last two commits are relevant for the new script load behavior.

### Changelog
- 🚀 Scripts are loaded into the pipeline instead in advance
  - It improves performance during first load
  - Improves cluster support with scripts
- 🐛 Automatically retry failed evalsha commands if the script is missing in redis, this improves the resiliency
- 🐛 Use the same script caching logic for standalone execution of scripts
- 🐛 Only send the scripts to a socket exactly once


test/functional/scripting.ts
```diff
@@ -125,96 +127,92 @@ describe("scripting", function () {
...
+    redis.defineCommand("test", {
+      numberOfKeys: 1,
+      lua: 'return redis.call("get", KEYS[1])',
+    });
+
+    await redis.test("preload");
+    await redis.script("flush");
+    const spy = sinon.spy(redis, "sendCommand");
+    const commands = await getCommandsFromMonitor(redis, 4, async () => {
+      const [a, b] = await redis.multi().test("foo").test("bar").exec();
```
marcbachmann, line 203: A test with manual transactions would be nice. 
But I don't think this is supported well at the moment.
e.g.
```js
redis.pipeline([['multi'], ['test', 'foo'], ['exec']]).exec()
```

lib/script.ts
```diff
@@ -32,28 +31,97 @@ export default class Script {
...
+// Standalone mode (cluster)
+function usingCluster(script, cluster, args, options, callback) {
+  const evalsha = createEvalShaWithFallback(script, args, options, cluster);
+  asCallback(evalsha.promise, callback);
+  return cluster.sendCommand(evalsha);
+}
+
+// Standalone mode (regular)
+function usingStandalone(script, redis, args, options, callback) {
+  if (redis._addedScriptHashes[script.sha]) {
```
marcbachmann, line 114: Instead of doing those decisions in here, we could move it into the `command.toWritable` function.
But we don't have any stream context there. https://github.com/luin/ioredis/blob/b8177479c348aa4bbd467fa944d61fe9b35aec19/lib/redis/index.ts#L790

lib/script.ts
```diff
@@ -32,28 +56,32 @@ export default class Script {
...
-        }),
-        callback
+    evalsha.promise = evalsha.promise.catch((err: Error) => {
+      if (err.toString().indexOf("NOSCRIPT") === -1) {
+        throw err;
+      }
+
+      // Resend the same custom evalsha command that gets transformed to an eval
+      // in case it's not loaded yet on the connectionDo an eval as fallback, redis will hash and load it
+      const resend = new this.Command(
```
marcbachmann, line 73: We could remove this fallback if order of delivery needs to be guaranteed.
It will still clear the cache for the next command that gets sent to redis.
https://github.com/luin/ioredis/pull/1499/files#diff-3618aaced6e025d978d2cdde354e756687916385e9024868d0ad3fff887c55c6R24

lib/redis/index.ts
```diff
@@ -787,7 +770,14 @@ Redis.prototype.sendCommand = function (command: Command, stream: NetStream) {
         command.args
       );
     }
-    (stream || this.stream).write(command.toWritable());
+
+    if (stream) {
+      if (stream.isPipeline)
+        stream.write(command.toWritable(stream.destination.redis.stream));
```
marcbachmann, line 776: For the pipeline we could move the buffer logic into the Redis class itself.
Maybe even persist the pipeline itself in the queue instead of calling sendCommand that often.

lib/script.ts
luin: I'm thinking to include this breaking change in the next major version https://github.com/luin/ioredis/pull/1504 so we can just retry custom commands in pipeline as well.

luin:
Hey @marcbachmann 👋,

Sorry for the delay. I'm planning v5 in https://github.com/luin/ioredis/pull/1504. So do you mind change the target branch of this PR to `v5` and resolve the conflicts?

marcbachmann:
@luin I've rebased this.
At least one cluster.disconnect test is missing as I've removed the script based ones:
https://github.com/luin/ioredis/pull/1499/files#diff-25757dbbc9670af28c244c85ae7b36c1224b1963223299f4df48963785141a6fL415

luin:
Awesome! Nice and clean improvements!

Issue #1377, opened 2021-07-03
when not to use enableAutoPipelining?


stavalfi:
From the docs, it sounds like there are only advantages. So why ioredis does not internally use it without any user-configurations?

* I have a single node cluster with 1 shard (master). Maybe in my specific use case there are only advatanges for `enableAutoPipelining`?

Also, If I send `multi`, will the commands inside this `multi` will be send in a "bigger" `multi` (incase there are some commands waiting in the queue)?

p.s. maybe the answers should be posted in the documentation.

----

Thanks for this super great feature!! Good job!!

TysonAndre:
> From the docs, it sounds like there are only advantages. So why ioredis does not internally use it without any user-configurations?

**Not a maintainer,** but there were still some open issues, some of which were recently fixed in 4.27.7 (#1391) - maybe they're waiting for more real-world testing or a major release before making it a default

Second, if 100 commands are automatically pipelined, and there's an error, you don't know how many failed, which may be fine in some applications but undesirable in others (e.g. if there were 100 increment commands, you wouldn't know how many (0-100) of the increments actually succeeded or actually failed - having 100 requests not knowing if they succeeded or failed after a spurious networking error is worse (in some applications) than just 1 request with an unknown result (reconnecting then sending the other 99))

> 2. If I send multi, will the commands inside this multi will be send in a "bigger" multi (incase there are some commands waiting in the queue)?

I think the multi is atomic and can't be added in a pipeline together with other commands from briefly reading the lib/autopipelining code (array of commands that forbid autopipelining). I'm not certain if that's correct.

> 3. With enableAutoPipelining=true, does ioredis send at most only 1 request at the same time?

No, that describes what happens when it's false. It sends many requests at a time without waiting for a response when it's true

stavalfi:
Thanks for your time @TysonAndre!

> if 100 commands are automatically pipelined, and there's an error, you don't know how many failed

How can it be? I will get the result-status of every one of my requests so I will know if a specific request succeed or not. Do I have any mistake?


> With enableAutoPipelining=true, does ioredis send at most only 1 request at the same time?

> No, that describes what happens when it's false. It sends many requests at a time without waiting for a response when it's true

I'm confused. it's the point of autopipeline to wait for more commands (while there is an ongoing request) and when there will be a response, collect all the pending requests and then go to redis only once? 

It seems like I understand the opposite from your explanantion.

TysonAndre:
[...]

PR #1504, opened 2022-02-06
Next major version: V5

luin:
Closes #1292

Major changes:
- [x] Drop support of Node.js < 12.
- [x] Drop support of plugging in third-party Promise Library. (As native Promise is pretty performant and Bluebird is deprecated.)
- [x] Upgrade TypeScript to the latest version.
- [x] Official TypeScript declarations.
- [x] Support ESM + CJS.
- [x] Scripts will retry in case of `NOSCRIPT` in pipeline mode. https://github.com/luin/ioredis/pull/1499
- [x] #1293
- [x] #1041 (May not merge directly, another approach is to add a warning)


AVVS:
@luin any chance we can squeeze in resp3 (dont remember this being supported, but might be wrong) / client side caching here? glad to help!

AVVS:
and another idea is native opentelemetry instrumentation

luin:
> @luin any chance we can squeeze in resp3 (dont remember this being supported, but might be wrong) / client side caching here? glad to help!

It hasn't been supported 🤣 . For myself, I don't feel that resp3 will bring a lot of real value, and I'm not sure as a driver, what kind of API needs to be provided to support the new features that resp3 brings. Similarly, client side caching does not seem to be as meaningful for a language that uses GC.

But I haven't thought much about this, so open to suggestions! If we want to support resp3, v5 seems to be a good chance to do so.

> and another idea is native opentelemetry instrumentation

Can you elaborate a little bit more on this? Does it mean to provide an interface to get the average command execution time or something similar?

AVVS:
> Can you elaborate a little bit more on this? Does it mean to provide an interface to get the average command execution time or something similar?

it sort of allows you to perform diagnostics better and get a complete trace of a request when every part of your request chain is instrumented, typically you'd use async_hooks, but can be anything

https://github.com/elastic/apm-agent-nodejs/blob/main/lib/instrumentation/modules/ioredis.js is an example of an existing instrumentation, which can be bridged to opentracing/opentelementry, fed to something like https://github.com/jaegertracing/jaeger for visualization/search

AVVS:
> It hasn't been supported 🤣 . For myself, I don't feel that resp3 will bring a lot of real value, and I'm not sure as a driver, what kind of API needs to be provided to support the new features that resp3 brings. Similarly, client side caching does not seem to be as meaningful for a language that uses GC.

I don;t think GC is relevant in terms of client-side caching here, in my opinion main benefit here is lower latency to access information, along with guarantees that information is still what server has. https://redis.io/topics/client-side-caching
for certain queries you can opt-in to cache the response. when server-side data changes - server will notify clients about that and clients will get new info

marcbachmann:
Resp3 changes response formats and we should mention that. eg. booleans are converted to true/false instead of 1/0: https://redis.io/topics/lua-api#lua-to-resp3-type-conversion

Therefore I'd schedule resp3 with another version later or support opt-out.

lib/Redis.ts
```diff
@@ -0,0 +1,785 @@
...
+      (!stream &&
+        this.status === "connect" &&
+        commands.exists(command.name) &&
+        commands.hasFlag(command.name, "loading"));
+    if (!this.stream) {
+      writable = false;
+    } else if (!this.stream.writable) {
+      writable = false;
+      // @ts-expect-error
+    } else if (this.stream._writableState && this.stream._writableState.ended) {
```
marcbachmann, line 644: Is this still needed with node 10?

luin:
Raising the Node.js version bar to v12 as it's the oldest version that is still in maintenance status. With this change, we are able to upgrade all dependencies to the latest versions, and we can also get rid of `lodash.flatten`.

Also, I think we should delay RESP3 for the next major version (v6) as seems it's a big change and we need some time thinking of it. @AVVS does it make sense?

luin:
This PR now includes all changes I want to make in v5, and I'm going to update some examples and documentation in the next week, and then I think we are good to release v5 as a beta version (in `next` dist channel) 🚀.

Let me know if you have something in your mind that can be done before we release v5 as a stable version @AVVS @marcbachmann. cc @shaharmor @TysonAndre 👋 

BTW going to consider RESP3 in the next major version 🤣

lib/utils/Commander.ts
```diff
@@ -0,0 +1,227 @@
...
+
+export interface CommanderOptions {
+  keyPrefix?: string;
+  showFriendlyErrorStack?: boolean;
+}
+
+const DROP_BUFFER_SUPPORT_ERROR =
+  "*Buffer methods are not available " +
+  'because "dropBufferSupport" option is enabled.' +
+  "Refer to https://github.com/luin/ioredis/wiki/Improve-Performance for more details.";
```
marcbachmann, line 20: That page doesn't exist anymore

luin: Good catch! The content was: https://github.com/luin/ioredis/wiki/_compare/c3aeabf0594e0c68b9c5ddf39d8cd7fa4876604a. Going to remove the `dropBufferSupport` option as hiredis is not used anymore.

luin:
> I think this could have a big performance impact because of buffer allocation in the tcp client. A benchmark would be nice for that.
> 
> edit: maybe not as writes still happen using strings

We had a benchmark at the end of https://github.com/luin/ioredis/wiki/_compare/c3aeabf0594e0c68b9c5ddf39d8cd7fa4876604a. There is no difference with the pure JavaScript parser whether the option is enabled or not. The option was useful because there was a non-trivial cost for hiredis (a parser written in C) to pass buffers from native to JavaScript. It's not an issue with the pure JavaScript parser that we are using.

> But maybe it could be interesting to migrate to the onread function introduced in node 12 in [socket.connect](https://nodejs.org/api/net.html#socketconnectoptions-connectlistener).
That way the buffer reuse would be possible.

Oh, good to know! Not sure how much it will improve the performance though.

TysonAndre:
> Let me know if you have something in your mind that can be done before we release v5 as a stable version

The biggest thing I could think of for a major version would be changing the default to `enableAutoPipelining: true` to avoid the potential bottleneck.

Things it would affect:

- `select` to manually select a db number (1505)
- applications where failures due to networking errors are important to avoid? https://github.com/luin/ioredis/issues/1377#issuecomment-890546745
   (In practice, I assume the number of in flight commands would be low if redis is in the same local network)

lib/cluster/index.ts
```diff
@@ -477,11 +479,9 @@ class Cluster extends Commander {
     let targetSlot = node ? node.slot : command.getSlot();
     const ttl = {};
     const _this = this;
-    // @ts-expect-error
-    if (!node && !command.__is_reject_overwritten) {
-      // TODO WeakMap
-      // @ts-expect-error
-      command.__is_reject_overwritten = true;
+    if (!node && !REJECT_OVERWRITTEN_COMMANDS.has(command)) {
```
marcbachmann, line 482: Keep in mind that the `command.promise` & `command.reject` could get overwritten in some cases.
But at the moment it's only done in the [pipeline](https://github.com/luin/ioredis/blob/b45a3e82d9f3daf096803936192b4397a6e05b46/lib/Pipeline.ts#L150) and for that `node` should be present here.
It's probably good enough at the moment, but `.has(command.promise)` might be better.

luin: Good point. Will think about it!

luin:
Renamed to the main branch. Published v5.0.0-beta.1.

luin:
> The biggest thing I could think of for a major version would be changing the default to `enableAutoPipelining: true` to avoid the potential bottleneck.
> 
> Things it would affect:
> 
> * `select` to manually select a db number (1505)
> * applications where failures due to networking errors are important to avoid? [When not to use enableAutoPipelining? #1377 (comment)](https://github.com/luin/ioredis/issues/1377#issuecomment-890546745)
>   (In practice, I assume the number of in flight commands would be low if redis is in the same local network)

Yeah I think that makes sense. Let's delay this to the next major release.

luin:
Hey @marcbachmann @TysonAndre 👋 , thanks for the contributions to ioredis. I've just invited you to become collaborators on this repository. Feel free to help review new pull requests in the future if you have time (and it doesn't matter at all if you don't have time 😄). Appreciate it 🙏. You can refer to https://github.com/luin/ioredis/blob/main/.github/CONTRIBUTING.md for some notes.

luin:
> Why disable it by default?

Hey @shaharmor , The option at some level introduces side effects as it consumes resources, especially during a failover, that by default every 5 seconds all nodes may have to be recreated (if `cluster slots` gives different results). Given that, I think developers need to be aware of this option and only enable it explicitly when they are aware of its purpose. WDYT?

PR #1516, opened 2022-03-01
fix unhandled promise rejection in cluster

qishibo:
The reproduction code  as follows, I want to perform unified processing when Redis executing, such as prohibiting some commands, so I rewrite the `sendCommand` method so that some commands can be intercepted [by reject or throw Errors], but the single Redis mode and Cluster mode have inconsistent responses: 

```javascript
var Redis = require("ioredis");
const { sendCommand } = Redis.prototype;

Redis.prototype.sendCommand = function(...options) {
    // record command log
    console.log('command name', options[0].name);

    // if I want to prohibit command like 'set' in blockList
    // if blockList.includes(options[0].name) {
    if (options[0].name == 'set') {
        return reject(new Error("You can not exec 'set' command!"));
    }

    sendCommand.call(this, ...options)
      .then(r => {resolve(r)})
      .catch(e => {reject(e)})
  })
}

// cluster mode
var redis = new Redis.Cluster([{ port: 7000, host: '127.0.0.1' }], {
  redisOptions: {password: 'xxx'},
});
// single Redis mode
// var redis = new Redis(6379, '127.0.0.1');


redis.on('ready', r => {
  var key = 'test_key_' + Math.random();

  redis.set(key, Math.random()).then(r => {
    console.log('set result', r);
  }).catch(e => {
    console.log('fail to set===', e.message);
  })
})
```
If I `reject` in sendCommand, the single Redis mode can catch the error when executing `set` and logs `fail to set`, 
but when in Cluster mode,  I got ` UnhandledPromiseRejectionWarning: Unhandled promise rejection. ` and I cannot handle the promise rejection anywhere.

This PR is to solve the problem and make them have the consistent responses

### DIFF
--- a/lib/cluster/index.ts
+++ b/lib/cluster/index.ts
@@ -730,7 +730,7 @@ class Cluster extends EventEmitter {
         }
       }
       if (redis) {
-        redis.sendCommand(command, stream);
+        redis.sendCommand(command, stream).catch(function(e) {command.reject(e);});
       } else if (_this.options.enableOfflineQueue) {
         _this.offlineQueue.push({
           command: command,


luin:
Hey @qishibo 👋 

Thanks for the contribution! `#sendCommand()` is a little bit counterintuitive as it has to return `command.promise` instead of a random Promise.

 In terms of your case, I think the following code should work:


```js
Redis.prototype.sendCommand = function (...options) {
  // record command log
  const command = options[0];
  console.log("command name", options[0].name);

  // if I want to prohibit command like 'set' in blockList
  // if blockList.includes(options[0].name) {
  if (options[0].name == "set") {
    return command.reject(new Error("You can not exec 'set' command!"));
  }

  return sendCommand.apply(this, options);
};
```

Does this solve your issue?

qishibo:
hi @luin , thank you for your reply! it works fine for cluster, but failed in single mode because `redis.get()` returns undefined instead of a promise, code should change to below, this will work for both cluster and single
```
Redis.prototype.sendCommand = function (...options) {
  const command = options[0];
  if (options[0].name == "set") {
    command.reject(new Error("You can not exec 'set' command!"));
    return command.promise; // return promise in single mode to avoid redis.get() return undefined
  }

  return sendCommand.apply(this, options);
};
```

Tips: ioredis is quite awesome

luin:
Thanks for the follow-up 👍. I think we can close this PR then.

PR #1517, opened 2022-03-02
feat: support ESM bundle

luin:
This PR generally follows https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c that it provides a wrapper for ESM instead of compiling TypeScript twice for different module targets in order to avoid [dual package hazard](https://nodejs.org/api/packages.html#packages_dual_package_hazard).


luin:
After a second thought, don't think the benefit is worth the complexity. Will be replaced by https://github.com/luin/ioredis/pull/1521

PR #1528, opened 2022-03-12
fix: remove unused Command#isCustomCommand

luin:
It was introduced to prevent sending custom commands in cluster+pipeline: https://github.com/luin/ioredis/blob/80f4a454fb0d24392e4fa27fbd3f6f22b0bf938e/lib/pipeline.js#L250, and it's not used anymore.

### Test plan
Make sure the property is not used anymore.

### DIFF
--- a/lib/Command.ts
+++ b/lib/Command.ts
@@ -140,7 +140,6 @@ export default class Command implements Respondable {
   isReadOnly?: boolean;
 
   args: CommandParameter[];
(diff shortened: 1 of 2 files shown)


luin:
@marcbachmann 👋 Can you take a look at this change if you have time?

marcbachmann (approved):
Nice. I also don't see any reason we should keep that 👏

PR #1532, opened 2022-03-15
feat: improve typings for smismember

### LINKED ISSUE #1429: Command SMISMEMBER is missing from types definition
Following #1281, the built-in command SMISMEMBER is in the redis client prototype, but is missing from the @types/ioredis definitions, hence breaking compatibility with TypeScript. Using @ts-ignore as a temporary workaround.

luin:
Closes #1429.

Changes:

1. Improved typings for `smismember` so it returns `number[]` instead of `unknown[]`.
2. Improved typings for`geopos`, `lmpop`, `blmpop`, `bzpopmin`, `bzpopmax`, and `xread`.
3. Make the declaration order stable to make further diffs easier.
5. Make numeric parameters also accept strings. Otherwise, it was a little tricky to figure out where the issue was (the correct fix was to change `'1000'` to `1000`):

CleanShot 2022-03-15 at 13 08 26@2x

### DIFF
--- a/bin/generateRedisCommander/index.js
+++ b/bin/generateRedisCommander/index.js
@@ -5,9 +5,9 @@ const overrides = require("./overrides");
 const { getCommanderInterface } = require("@ioredis/interface-generator");
 
 const ignoredCommands = ["monitor", "multi"];
-const commands = require("@ioredis/commands").list.filter(
-  (name) => !ignoredCommands.includes(name)
-);
+const commands = require("@ioredis/commands")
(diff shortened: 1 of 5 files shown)


marcbachmann (approved):
Looks good, types match the one of the docs.
I'm just not sure about the `typeMaps.number` change as I don't know the internals yet... but that's covered in your 4. point.

luin:
We use the `COMMAND DOCS` command to get parameters and return types for every command. `typeMaps` is used to map those types to JavaScript types.

PR #1534, opened 2022-03-16
Fix changelog url in readme

Koroeskohr:


luin:
Thanks for the fix!

Koroeskohr:
such speed! cheers

PR #1544, opened 2022-03-27
Loosen callback typings

luin:
In @types/ioredis, callbacks are defined as `(err: Error | null, res: T) => void`, whereas in ioredis v5, they are `(err?: Error | null, res?: T) => void`.

The intention was to indicate the `res` can be absent when there is an error. However, given TypeScript doesn't provide a way to achieve "if there is no error, then res should be fulfilled, developers will need to check `res` in every callback even if they have already checked `err`:

```ts
  redis.get('foo', (err, res) => {
    if (err) reject(err);
    // Can't use `else` here
    if (typeof res !== 'undefined') resolve(res);
  });
});
```

I feel it's a bit inconvenient, and apparently, people have different opinions on it: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43583.

I lean towards keeping in line with the Node.js way: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/884d68c/types/node/fs.d.ts#L2416 though no strong opinion.

Open to any suggestions! 😄

### DIFF
--- a/lib/Redis.ts
+++ b/lib/Redis.ts
@@ -773,7 +773,7 @@ class Redis extends Commander {
           );
           return callback(null, {});
         }
-        return callback(err);
+        return callback(err, undefined);
       }
       if (typeof res !== "string") {
         return callback(null, res);
--- a/lib/cluster/index.ts
+++ b/lib/cluster/index.ts
@@ -827,7 +827,7 @@ class Cluster extends Commander {
 
     duplicatedConnection.cluster(
       "SLOTS",
-      timeout((err: Error, result) => {
+      timeout((err, result) => {
         duplicatedConnection.disconnect();
         if (err) {
           return callback(err);
@@ -842,7 +842,7 @@ class Cluster extends Commander {
             result.length,
             this.status
           );
-          callback();
+          callback(null);
           return;
         }
         const nodes: RedisOptions[] = [];
@@ -900,7 +900,7 @@ class Cluster extends Commander {
         }
 
         this.connectionPool.reset(nodes);
-        callback();
+        callback(null);
       }, this.options.slotsRefreshTimeout)
     );
   }
@@ -922,7 +922,7 @@ class Cluster extends Commander {
         return callback(err);
       }
       if (typeof res !== "string") {
-        return callback();
+        return callback(null);
       }
 
       let state: string;
@@ -939,7 +939,7 @@ class Cluster extends Commander {
         debug("cluster state not ok (%s)", state);
         callback(null, state);
       } else {
-        callback();
+        callback(null);
       }
     });
   }
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -1,7 +1,7 @@
 import { Socket } from "net";
 import { TLSSocket } from "tls";
 
-export type Callback<T = any> = (err?: Error | null, result?: T) => void;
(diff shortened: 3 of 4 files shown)


fabrykowski:
Why not model the type exactly after ist usage?
```typescript
export type Callback<T = any> = (err: Error, result: never) => void | (err: undefined, result: T) => void;
```
By the way, I believe that the more "TypeScript-y" default type for `T` would be `unknown`, resulting in
```typescript
export type Callback<T = unknown> = (err: Error, result: never) => void | (err: undefined, result: T) => void;
```

luin:
@fabrykowski Don't think that works. TypeScript asks to [give the parameters types](https://www.typescriptlang.org/play?ts=4.6.2#code/C4TwDgpgBAwghgGwQIzgYwNZQLxQBR4QBORAXFAKIkD2RANFERAM4CuCw5AdhAG7EBKHAD4ovagEsAJkIA++QiXKsuUiADMJPKQyZsO3VgFtkgkWMkyA3AFgAUPbTUuzYFDjl4SVJhz5i9Iws7MBC2KIA3vZQMVAS6v4kQlF2sWlQTi7UCBAAdAjUAOaKRAK2qbEAvlAQCMzQKemxmczZeQXFeiFl0VX2lfZAA).

luin:
In that case seems `result` [is never be `never`](https://www.typescriptlang.org/play?ts=4.6.2#code/KYDwDg9gTgLgBDAnmYcDCBDANlgRhgYwGsAeAFTgF44BXAOyLogHc6A+KuACgDo+MoAcwDOALjgBtYFCjiAojOgAaOFGDCaWGOLrAAbtIC6cAD6Tps2nQAmwAGYBLXdZVqNW8WUMBKKhz0QDtYA3ACwAFARBBB0wvAY4pg4+MQkdDQAtrjSHNRcFq7qmjC+lBwA3hFw1XDRsRBYwDxYEIJcAIxwALSqRVreYeEAvhFAA) 😢

fabrykowski:
If we check for the existence of the error, then `result` does become `never`.
Unfortunately if we don't check, the type of `result` is `never | T`, which TypeScript simplifies to `T`.
Of course `result` can be set to `undefined`:
```typescript
export type Callback<T = unknown> = (...args: [err: Error, result: undefined] | [err: undefined, result: T]) => void;
```
[playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBDAnmYcDCBDANlgRhgYwGsAeAFTgF44BXAOyLogHc6A+KuACgDo+MoAcwDOALjgBtYFCjiAojOgAaOFGDCaWGOPoATYADMAlnWC6AunAA+k6bNp19x07pVqNW8WXMBKKhwA3CCNdAG4AWAAoKIIIOmF4DHFMHHxiEjoaAFtcaQ5qLjs3dU0YP0oOAG8ouFq4IwNuOz9qyLr2uFj4iCxgHiwIQUKZHwi2uoBfOGAsYVRWjrqu4R6+gaH3UtGayaiJqJi4hLhcZOw8QlJMnLzOYahijzL-OAWlo9X+wa4ARjgAWlUJS020i+0iQA)
We lose some of the semantics, but gain a functioning definition 😅

luin:
@fabrykowski Oh that's awesome! I didn't know that was possible. Looks like it's only supported from TypeScript 4.6 but seems to be a good progressive enhancement.

PR #1545, opened 2022-03-27
fix: support TypeScript interface as parameters of hmset and mset

### LINKED ISSUE #1536: hmset in typescript object
I tried to insert one simple object in typescript to ioredis hashmap. I define object like this

interface User {
name : string,
age :string
}

let newUser : User = {
name : "suman",
age :"34"
}
How do I write hmset command ?

luin:
Closes #1536

There is a limit on the TypeScript side https://github.com/microsoft/TypeScript/issues/15300#issuecomment-891681234 that you can't assign an interface to a record:

```ts
interface User {
  name: string;
  age: string;
}

let newUser: User = {
  name: "suman",
  age: "34",
};

const record: Record<string, string> = newUser; // Error here
```

Given `hmset` and `mset` already support ES6 `Map`, I think it's should be fine to loosen the typing here for convenience.

It's possible that developers may pass invalid object values (e.g. `redis.hmset("user-hash", new WeakMap([[{ name: "Bob" }, "value"]]))`) but I think that's fine since the usage doesn't make sense and it will trigger an runtime error anyway.

### DIFF
--- a/bin/overrides.js
+++ b/bin/overrides.js
@@ -1,8 +1,7 @@
 const msetOverrides = {
   overwrite: false,
   defs: [
-    "$1(object: Record<string, string | Buffer | number>, callback?: Callback<'OK'>): Result<'OK', Context>",
-    "$1(map: Map<string | Buffer | number, string | Buffer | number>, callback?: Callback<'OK'>): Result<'OK', Context>",
+    "$1(object: object, callback?: Callback<'OK'>): Result<'OK', Context>",
   ],
 };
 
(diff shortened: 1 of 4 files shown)


bin/overrides.js
```diff
@@ -19,15 +18,13 @@ module.exports = {
   hset: {
     overwrite: false,
     defs: [
-      "$1(key: RedisKey, object: Record<string, string | Buffer | number>, callback?: Callback<number>): Result<number, Context>",
-      "$1(key: RedisKey, map: Map<string | Buffer | number, string | Buffer | number>, callback?: Callback<number>): Result<number, Context>",
+      "$1(key: RedisKey, object: object, callback?: Callback<number>): Result<number, Context>",
```
marcbachmann, line 21: Should we keep the map declaration to have better hints?
How is the non-string key support for the type `object`?
```js
      "$1(key: RedisKey, object: object, callback?: Callback<number>): Result<number, Context>",
      "$1(key: RedisKey, map: Map<string | Buffer | number, string | Buffer | number>, callback?: Callback<number>): Result<number, Context>",
```

test/typing/transformers.test-d.ts
```diff
@@ -0,0 +1,31 @@
+import { expectType } from "tsd";
+import Redis from "../../built";
+
+interface User {
+  name: string;
+  title: string;
+}
+
+const user: User = { name: "Bob", title: "Engineer" };
+const map = new Map([["key", "value"]]);
```
marcbachmann, line 10: I think a test with `new Map([[1, "value"]]);` and `new Map([[Buffer.from('key'), "value"]]);` would be good

luin: Added to both typing tests and functional tests.

bin/overrides.js
luin: `Map` is also `object` but yeah let's keep them for documentation purposes. Only utf8 keys are supported for `object` so users have to use `Map` if they need others.

marcbachmann (approved):

PR #1548, opened 2022-03-28
Remove instruction to install @types/node

bttger:
The [upgrade guide](https://github.com/luin/ioredis/wiki/Upgrading-from-v4-to-v5) says that v5 provides TypeScript declarations officially so you can uninstall the types

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,12 +105,6 @@ Medis starts with all the basic features you need:
 $ npm install ioredis
 ```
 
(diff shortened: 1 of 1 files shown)


bttger:
Sorry, of course @types/node is not the same as @types/ioredis :laughing:

luin:
Aha was going to say that. @types/node is needed as it provides the declaration for `Buffer`.

PR #1554, opened 2022-03-30
Hook up the keepAlive after a successful connect

### LINKED ISSUE #1339: when `keepAlive` parameter is set, the TCP keep-alive is enabled but the specified timeout is not applied
**ioredis version:** 4.26.0
**Node.js version:** 14.16.1
**OS**: Ubuntu 18.04.3 LTS

Setting the `keepAlive` option enables the TCP keep-alive mechanism but does not apply the specified timeout.

This can be reproduced with:
```js

const redis = new Redis({
  host: 'localhost',
  keepAlive: 30000, // should set socket option TCP_KEEPIDLE=30
});

redis.on('connect', () => {
  console.log('Connected!');
});
```

Running [`knetstat`](https://github.com/veithen/knetstat) to inspect socket options shows the socket was created with `TCP_KEEPIDLE=60` (which is the Node.js default):
```shell
$ cat /proc/net/tcpstat
Recv-Q Send-Q Local Address   Foreign Address  Stat Diag Options
     0      0 127.0.0.1:40792 127.0.0.1:6379   ESTB      SO_REUSEADDR=0,SO_REUSEPORT=0,SO_KEEPALIVE=1,TCP_KEEPIDLE=60,TCP_KEEPCNT=10,TCP_KEEPINTVL=1,TCP_NODELAY=1,TCP_DEFER_ACCEPT=0
```
---
This is likely due to `socket.setKeepAlive()` being called after the connection has been established.
To test this I have created a socket and tried calling `socket.setKeepAlive()` before and after calling `socket.connect()`.

When `socket.setKeepAlive()` is called **before** `socket.connect()`:
```js
const net = require('net');

const socket = new net.Socket();

socket.setKeepAlive(true, 10000);

socket.connect({
  host: 'localhost',
  port: 6379,
});

socket.on('connect', () => {
  console.log('Connected!');
});
```

`knetstat` reports `TCP_KEEPIDLE=10` - keep-alive is enabled and the correct timeout is applied

---
When `socket.setKeepAlive()` is called **after** `socket.connect()`:
```js
const net = require('net');

const socket = new net.Socket();

socket.connect({
  host: 'localhost',
  port: 6379,
});

socket.setKeepAlive(true, 10000);

socket.on('connect', () => {
  console.log('Connected!');
});
```

`knetstat` reports `TCP_KEEPIDLE=60` - keep-alive is enabled but the Node.js default timeout is applied

marcbachmann:
I haven't tested it yet as I'm running on macos and there's no way to read out the socket options, but this should fix #1339

### Changelog
- 🐛 Call `socket.setKeepAlive(true, ms)` after a successful connect as workaround for a node issue where it won't work when it's called before a successful connect

### DIFF
--- a/lib/Redis.ts
+++ b/lib/Redis.ts
@@ -213,8 +213,19 @@ class Redis extends Commander {
           }
 
           _this.stream = stream;
-          if (typeof options.keepAlive === "number") {
-            stream.setKeepAlive(true, options.keepAlive);
+
+          if (options.noDelay) {
+            stream.setNoDelay(true);
+          }
+
+          // Node ignores setKeepAlive before connect, therefore we wait for the event:
+          // https://github.com/nodejs/node/issues/31663
+          if (typeof options.keepAlive === 'number') {
+            if (stream.connecting) {
+              stream.once(CONNECT_EVENT, () => stream.setKeepAlive(true, options.keepAlive));
+            } else {
+              stream.setKeepAlive(true, options.keepAlive);
+            }
           }
 
           if (stream.connecting) {
@@ -264,10 +275,6 @@ class Redis extends Commander {
           if (!stream.destroyed) {
             stream.once("error", eventHandler.errorHandler(_this));
             stream.once("close", eventHandler.closeHandler(_this));
-          }
-
-          if (options.noDelay) {
-            stream.setNoDelay(true);
           }
 
           const connectionReadyHandler = function () {
--- a/lib/connectors/SentinelConnector/index.ts
+++ b/lib/connectors/SentinelConnector/index.ts
@@ -165,14 +165,15 @@ export default class SentinelConnector extends AbstractConnector {
           resolved.port,
           endpointAddress
         );
+
         if (this.options.enableTLSForSentinelMode && this.options.tls) {
           Object.assign(resolved, this.options.tls);
           this.stream = createTLSConnection(resolved);
+          this.stream.once("secureConnect", this.initFailoverDetector.bind(this));
         } else {
           this.stream = createConnection(resolved);
+          this.stream.once("connect", this.initFailoverDetector.bind(this));
         }
-
-        this.stream.once("connect", () => this.initFailoverDetector());
 
         this.stream.once("error", (err) => {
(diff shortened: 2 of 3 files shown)


marcbachmann:
https://github.com/nodejs/node/issues/31663

marcbachmann:
> Alternatively we could use socket.connect({keepAlive: true, noDelay: true}[, listener])

For the standalone connector:
```diff
diff --git a/lib/connectors/StandaloneConnector.ts b/lib/connectors/StandaloneConnector.ts
index 2a32e39..9c6bdd5 100644
--- a/lib/connectors/StandaloneConnector.ts
+++ b/lib/connectors/StandaloneConnector.ts
@@ -8,6 +8,8 @@ export type StandaloneConnectionOptions = (Partial<TcpNetConnectOpts> &
   Partial<IpcNetConnectOpts>) & {
   disconnectTimeout?: number;
   tls?: ConnectionOptions;
+  noDelay?: boolean;
+  keepAlive: number;
 };

 export default class StandaloneConnector extends AbstractConnector {
@@ -41,6 +43,15 @@ export default class StandaloneConnector extends AbstractConnector {
       Object.assign(connectionOptions, options.tls);
     }

+    if (options.noDelay === true) {
+      connectionOptions.noDelay = true;
+    }
+
+    if (typeof options.keepAlive === 'number') {
+      connectionOptions.keepAlive = options.keepAlive;
+    }
+
+
```

Sentinel is not much more different.

marcbachmann:
Passing connection options during `createConnection` is more straight forward. I've pushed the changes.

luin:
Is this ready for review? I think we also need to change SentinelConnector?

marcbachmann:
image

`noDelay` only got added in node 17. So I'll need to revert that change 😢

marcbachmann:
This is now ready. I've decided to move it back into the redis constructor to keep it in one place with the reconnect handling.
So it should work now for sentinel and regular use. I also found one small issue with sentinel where the failover event listener wasn't set up properly.  That's in the second commit now.

luin:
LGTM! Awesome 👍

PR #1578, opened 2022-05-13
fix: sentinel slaves called replicas in Redis v7

joostdebruijn:
In Redis v5.0 the command `SENTINEL SLAVES` has been changed to `SENTINEL REPLICAS`. The old command was still available as alias, however from v7.0 and on the alias has been removed (which isn't mentioned in the Redis release notes, btw). This breaks the preferred slaves functionality, for example.

This change is thus breaking backwards compatibility with Sentinels running v4.0 or older. Maybe we need/must resolve that in one way or another as compatibility from v2.6.12 is claimed for this package.

### DIFF
--- a/lib/connectors/SentinelConnector/index.ts
+++ b/lib/connectors/SentinelConnector/index.ts
@@ -259,7 +259,7 @@ export default class SentinelConnector extends AbstractConnector {
   private async resolveSlave(
     client: RedisClient
   ): Promise<TcpNetConnectOpts | null> {
-    const result = await client.sentinel("slaves", this.options.name);
+    const result = await client.sentinel("replicas", this.options.name);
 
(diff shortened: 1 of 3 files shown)


joostdebruijn:
@luin I've solved that in the new commit (and I fixed the tests as well). Because Node 12 and 14 don't support Promise.all I added a small helper function to achieve the same behavior.

luin:
Thanks @joostdebruijn! I just did a quick search and found out it was a bug on Redis side and had just been fixed 22 hours ago 😄 https://github.com/redis/redis/pull/10723. Given it's a bug and has been fixed in a pretty small time window, I don't think we should add workarounds for it.

joostdebruijn:
Ah, nice! Let's close this one. :)

PR #1585, opened 2022-05-21
fix: Add back Pipeline#length

### LINKED ISSUE #1584: Typescript woes with Pipeline
ioredis 4, the "Redis.Pipeline" type had a "length" parameter, so I could do things like this:

```
    async executePipeline(pipeline: Redis.Pipeline): Promise<[Error | null, any][]> {
        if (pipeline.length < 1) {
            this.logger.warn('Got zero length pipeline to execute');
        }
        const t0 = performance.now();
        const res = await pipeline.exec();
        const t1 = performance.now();
        Math.random() < 0.1 && this.logger.info({ timeTaken: t1 - t0, length: pipeline.length }, `Executed cache pipeline`);
        return res;
    }
```

In ioredis 5 there doesn't seem to be a way to check the length of the pipeline?  This is a problem because I have several places where I'm doing a bunch of work in a loop, pumping it into a pipeline and when the length hits a threshold doing an `await pipeline.exec()`, then creating a fresh pipeline and the loop continues...

luin:
Closes #1584

### DIFF
--- a/bin/template.ts
+++ b/bin/template.ts
@@ -11,8 +11,9 @@ export interface ResultTypes<Result, Context> {
   pipeline: ChainableCommander;
 }
 
-export interface ChainableCommander
-  extends RedisCommander<{ type: "pipeline" }> {}
+export interface ChainableCommander extends RedisCommander<{ type: "pipeline" }> {
+  length: number;
+}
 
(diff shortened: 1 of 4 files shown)


simllll:
Exactly what I was looking for, the length itself is here, the type is just missing

PR #1599, opened 2022-06-08
hgetallBuffer response type should be a Record, not 2-element array 

richardwu:
Tested this on redis version 7


luin:
Hey @richardwu 👋,

Thanks for the PR. However, [lib/utils/RedisCommander.ts](https://github.com/luin/ioredis/pull/1599/files#diff-efcdb6a0baf8186bd3d2429e33931c544e941eed936d9bde7cd0cc57a1a0cdab) is generated from bin/template.ts, so I think the fix should be made in other places.

PR #1607, opened 2022-06-17
fix: Allow address family to be resolved from DNS

### LINKED ISSUE #1576: Unhandled error event: Error: getaddrinfo ENOTFOUND  - AWS ElasticCache
Hey guys, did anyone have some issue like that to connect in the AWS ElasticCache in the ECS Container?

[ioredis] Unhandled error event: Error: getaddrinfo ENOTFOUND cursobeta-dev-elasticache-redis.ilg7lv.ng.0001.use1.cache.amazonaws.com

Looks like permision, right?

`const pubClient = new Redis({
  host: process.env.AWS_REDIS_HOST,
  port: 6379
});`

richiejp:
It may not be possible to know before hand what address a host will
resolve to.

Possibly this fixes #1576

### DIFF
--- a/lib/redis/RedisOptions.ts
+++ b/lib/redis/RedisOptions.ts
@@ -189,7 +189,6 @@ export const DEFAULT_REDIS_OPTIONS: RedisOptions = {
   // Connection
   port: 6379,
   host: "localhost",
-  family: 4,
   connectTimeout: 10000,
   disconnectTimeout: 2000,
   retryStrategy: function (times) {


richiejp:
@luin Ping

luin:
Sorry for the delay! Should it be considered as a breaking change?

maxpain:
Any updates on this?

Defman:
I have sort of the inverse problem. In our dualstack setup we have aws vpc dns64, this works great with our nat... however our elasticache is placed in a legacy VPC without ipv6. However this should in its own not be a problem. But since the dns server returns both ipv4 and ipv6 addresses for the elasticache cluster dns... ioredis seems to get stuck in `ClusterAllFailedError: Failed to refresh slots cache`. I have tried setting `?family=4` however this does not seem to force ioredis to resolve the hostname to only ipv4. Any suggestions for workarounds, place a dns server in front of the default dns server and have it strip out the ipv6 addresses...

Note: ioredis does not seem to have this problem.

PR #1618, opened 2022-07-09
feat: add mode property to client

joelrbrandt:
Adds a `.mode` getter to the client that reports whether the client is in 'normal', 'subscriber', or 'monitor' mode.


luin:
Hey @joelrbrandt 👋 ,

Thanks for the PR. Can you elaborate a bit more on the use case of this change?

PR #1620, opened 2022-07-14
fix: srandmember with count argument should return array of strings or array of unknown

### LINKED ISSUE #1621: Type error, srandmember with count argument should return array of strings or array of unknown instead of string or array of unknowj
Hey,

In the current type implementation of `srandmember` with `count` argument the return type is `string | unknown[]` instead of `(string | unknown)[]` for returning multiple values

This is the entire type:
`srandmember(key: RedisKey, count: number | string, callback?: Callback<string | unknown[] | null>): Result<string | unknown[] | null, Context>;`

it should be:

`srandmember(key: RedisKey, count: number | string, callback?: Callback<(string | unknown)[] | null>): Result<(string | unknown)[] | null, Context>;`

Opening a PR for a fix

joni7777:
Closes #1621

### DIFF
--- a/bin/returnTypes.js
+++ b/bin/returnTypes.js
@@ -265,7 +265,7 @@ module.exports = {
   sort: "number" | "unknown[]",
   sortRo: "unknown[]",
   spop: (types) => (types.length > 1 ? "string[]" : "string | null"),
-  srandmember: (types) => (types.length > 1 ? "(string | unknown)[] | null" : "string | null"),
+  srandmember: (types) => (types.length > 1 ? "string[]" : "string | null"),
   srem: "number",
   strlen: "number",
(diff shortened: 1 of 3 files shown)


luin:
Hey @joni7777 👋 ,

Good catch! However, RedisCommander.ts is generated from bin/index.js so changes should be made on that file.

joni7777:
@luin awesome, fixing

joni7777:
@luin Fixed it, i think it supports now the two variants (with and without count)

luin:
Made some changes. LMK if they make sense. @joni7777

joni7777:
> Made some changes. LMK if they make sense. @joni7777 

Yea perfectly I think removing the unknown type is much better

PR #1631, opened 2022-08-09
Remove lazyConnect from Cluster redisOptions

shaharmor:
It is already being set by the cluster to `true` and there is no way to override it: https://github.com/luin/ioredis/blob/main/lib/cluster/ConnectionPool.ts#L84


luin:
`lazyConnect` is actually supported (though I can't recall why we want to allow that) since `this.redisOptions` overrides `{ lazyConnect: true }` given how `_.defaults` works.

PR #1655, opened 2022-09-19
Add node 18 to CI

silverwind:
Let's see how this goes.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [12.x, 14.x, 16.x]
+        node: [12.x, 14.x, 16.x, 18.x]
         redis: ['2', '6']
     steps:
       - name: Git checkout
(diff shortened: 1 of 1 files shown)


TysonAndre:
I'm not familiar with this project's test strategy but this looks reasonable with node 18 LTS starting October 25th


Unrelatedly, I see redis 7.0 stable was recently released.

Also, a question for other maintainers: I've used `include` in other projects to reduce the number of jobs while allowing testing more combinations without a large number of excludes. Would it make sense to start testing more redis versions with this?

```yaml
matrix:
  include:
    - node: 12.x
      redis: '2'
    - node: 12.x
      redis: '6'
    - node: 12.x
      redis: '7'
      # ...
```

luin:
> Also, a question for other maintainers: I've used include in other projects to reduce the number of jobs while allowing testing more combinations without a large number of excludes. Would it make sense to start testing more redis versions with this?

@TysonAndre You mean only test all Redis versions with only one Node.js version? That does make sense to me but I'm okay with 4 (Node.js versions) * 3 (Redis versions) so I think we can stay without using `include`.

PR #1668, opened 2022-10-23
Add missing typing for xautoclaim/xreadgroup

nicknotfun:
Was using these in my project and wanted to share more accurate typing for them, in particular the Buffer variants are missing entirely


calebboyd:
I made #1676 and then saw this PR. They're pretty much the same except I think xreadgroup can potentially have null entries if you're reading deleted values from the PEL.

Issue #1672, opened 2022-11-02
Node 18 family parameter fails when passing url


ksmithut:
The following example fails in node 18

```js
import Redis from 'ioredis'

new Redis('redis://localhost:6379?family=6')
```

```
[ioredis] Unhandled error event: TypeError [ERR_INVALID_ARG_VALUE]: The property 'options.family' must be one of: 0, 4, 6. Received '6'
    at lookup (node:dns:179:11)
    at node:net:1170:5
    at defaultTriggerAsyncIdScope (node:internal/async_hooks:466:18)
    at lookupAndConnect (node:net:1169:3)
    at Socket.connect (node:net:1097:5)
    at Object.connect (node:net:219:17)
    at .../node_modules/ioredis/built/connectors/StandaloneConnector.js:58:45
    at process.processTicksAndRejections (node:internal/process/task_queues:77:11)
```

This is because node 18 stopped type coercing the family value. https://github.com/nodejs/node/pull/41431

I'll attempt to submit a PR for this.

ksmithut:
PR submitted: https://github.com/luin/ioredis/pull/1673

ksmithut:
@luin Thanks for the quick turnaround 🍺

PR #1673, opened 2022-11-02
Fixes passing in family parameter in URL in node 18

ksmithut:
Node 18 stopped coercing the type for its options (specifically the ip family option) and instead throws an error. So if you pass in `?family=6` in your Redis URL, it would fail because when it does the DNS lookup passing in the string `'6'`, node would throw an error.

Fixes https://github.com/luin/ioredis/issues/1672


luin:

PR #1687, opened 2022-12-02
docs: update readme to fix slotsRefreshInterval default value

metzgegu:
## Description 

👋 Hello, when upgrading my version of ioredis from v4 to v5 I noticed a desync in the Readme

[Breaking changes](https://github.com/luin/ioredis/blob/main/CHANGELOG.md#breaking-changes) include that 

> `slotsRefreshInterval` is disabled by default, previously, the default value was 5000.

In the [Readme](https://github.com/luin/ioredis#cluster), it is still mentioned that its default value is 5000


luin:

Issue #1692, opened 2022-12-15
Keep getting MOVED errors after enabling autopipelining


gustavotoyota:
For some reason I keep getting MOVED errors after enabling auto-pipelining.
Anyone knows a possible reason for this?

![Screenshot from 2022-12-15 00-05-31](https://user-images.githubusercontent.com/11639451/207770619-6c53700f-a973-4463-98c7-be50cf479c79.png)

I thought the library was supposed to handle these errors automatically.
I'm using some Lua scripts in Redis, maybe it has something to do with that?

gustavotoyota:
I found that my redis cluster is behaving very strangely.
One of the slaves isn't synchronizing with its master at all.
I'll try to fix it and close this issue.

gustavotoyota:
Strangely, I keep receiving MOVED errors even after fixing the slave:
```
2022-12-15T16:23:43.951551038Z 0|index  | You have triggered an unhandledRejection, you may have forgotten to catch a Promise rejection:
2022-12-15T16:23:43.951619937Z 0|index  | WO: MOVED 15327 5.161.75.167:48418
2022-12-15T16:23:43.951631339Z 0|index  |     at R0e (/app/index.js:195:11553)
2022-12-15T16:23:43.951635837Z 0|index  |     at pG (/app/index.js:195:12448)
```

gustavotoyota:
Okay I found something.
Auto pipelining in ioredis separates pipelines per node.
I solved by making ioredis separate pipelines per slot.
I don't know why it didn't happen to anyone else, given Redis itself states that pipelines should always target a single slot.

I will close the issue and open a PR.

PR #1693, opened 2022-12-15
fix: make autopipelining group commands by slot instead of by node

gustavotoyota:
This change solved my problems with auto-pipelining + clusters.
ioredis itself states that pipeline commands should target keys in the same slot:
https://github.com/luin/ioredis/blob/807cfc0e211f72885e00228edc5e72878916a938/lib/Pipeline.ts#L309

### DIFF
--- a/lib/autoPipelining.ts
+++ b/lib/autoPipelining.ts
@@ -146,9 +146,7 @@ export function executeWithAutoPipelining(
   // ioredis will only flatten one level of the array, in the Command constructor.
   const prefix = client.options.keyPrefix || "";
   const slotKey = client.isCluster
-    ? client.slots[
-        calculateSlot(`${prefix}${getFirstValueInFlattenedArray(args)}`)
-      ].join(",")
+    ? calculateSlot(`${prefix}${getFirstValueInFlattenedArray(args)}`)
     : "main";
 
   if (!client._autoPipelines.has(slotKey)) {


gustavotoyota:
This is the issue I opened where I explored the problem and found the solution: https://github.com/luin/ioredis/issues/1692

TysonAndre:
```
    - `retryDelayOnMoved`: By default, this value is `0` (in ms), which means when a `MOVED` error is received, the client will resend
      the command instantly to the node returned together with the `MOVED` error. However, sometimes it takes time for a cluster to become
      state stabilized after a failover, so adding a delay before resending can prevent a ping pong effect.
```

I wonder if the lib/cluster/index.ts handling of `"MOVED"` is wrong for pipelines and it's retrying the entire pipeline? I'm not familiar with how that works

From the code of lib/Pipeline.ts - it looks like the entire pipeline is re-execed on `moved:`, which would make this fix appropriate

----

Note that there's 65536 slots, so this would have some impact on throughput (e.g. a hundred commands might still be a hundred stream writes). I wonder if it'd be better to group the writes from the pipelines (that would go to the same `client` belonging to the cluster)  into a single write() rather than executing multiple write()s to avoid head of line blocking and keep throughput high

TysonAndre:
https://github.com/luin/ioredis/issues/1377#issuecomment-1357875825

Never mind, I've been misunderstanding what autopipelining does. https://en.wikipedia.org/wiki/Protocol_pipelining is used whether or not enableAutoPipelining is enabled, enableAutoPipelining just reduces the tcp overhead by combining smaller writes into a larger write and sends fewer packets as a result

TysonAndre:
> From the code of lib/Pipeline.ts - it looks like the entire pipeline is re-execed on moved:, which would make this fix appropriate

This PR definitely looks appropriate, especially since the error handling is only done when all errors are exactly the same (e.g. the old way wouldn't handle multiple slots being moved).

JTaylor-myenergi:
@TysonAndre @luin Is there any reason this wasn't merged? We have applied this as a local patch to fix the same autopipelining issues with our Redis Clusters.

gianDiazM (approved):

AVVS:
My 2 cents is that you shouldnt use autopipelining if slots are frequently moved. Idea is that you combine requests that happen during the tick into as few large pipes as possible. One node hosts many slots, by default you have 16k slots, so this basically disables benefits of autopipelining as you’d have many small groups instead of few large ones

-1 to this pr

PR #1694, opened 2022-12-16
add type for known event of EventEmitter

trim21:
help callback of pub/sub typed

https://github.com/luin/ioredis#pubsub


luin:
Thanks for the contribution!

PR #1704, opened 2023-01-14
chore: Upgrade dependencies

luin:


package.json
```diff
@@ -79,9 +79,9 @@
     "sinon": "^13.0.1",
     "ts-node": "^10.4.0",
     "tsd": "^0.19.1",
-    "typedoc": "^0.22.12",
+    "typedoc": "^0.22.18",
     "typescript": "^4.6.3",
-    "uuid": "^8.3.0"
+    "uuid": "^9.0.0"
```
abenhamdine, line 84: uuid 9.0.0 drops Node.js 10.x,  IE 11 and Safari 10 support  
Is that intended to update this module ?
see https://github.com/uuidjs/uuid/blob/main/CHANGELOG.md

luin: Yes. ioredis 5 already requires Node.js 12 and also uuid is a dev-only dependency so it should be safe.

PR #1707, opened 2023-01-24
remove extraneous TCP/IPC properties from RedisOptions TS type

### LINKED ISSUE #1680: TS2322 Typescript: 'keepAlive' type 'number' is not assignable to type 'never'
I have the following code:
```
public toClientConfig(): RedisOptions {
        return {
           host: this.host,
           port: this.port,
           db: this.db,
           password: this.password,
           keepAlive: 1,
           lazyConnect: true,
        };
     }
```

I got this error during the build:

```
src/cache/classes/cache-config.ts:27:10 - error TS2322: Type 'number' is not assignable to type 'never'.

27          keepAlive: 1,
            ~~~~~~~~~
node_modules/ioredis/built/redis/RedisOptions.d.ts:19:5
    19     keepAlive?: number;
           ~~~~~~~~~
    The expected type comes from property 'keepAlive' which is declared here on type 'RedisOptions'

```

I think it's a weird type overlap with **TcpSocketConnectOpts** 

```
interface TcpSocketConnectOpts extends ConnectOpts {
        port: number;
        host?: string | undefined;
        localAddress?: string | undefined;
        localPort?: number | undefined;
        hints?: number | undefined;
        family?: number | undefined;
        lookup?: LookupFunction | undefined;
        noDelay?: boolean | undefined;
        keepAlive?: boolean | undefined;
        keepAliveInitialDelay?: number | undefined;
    }
```

How can I solve it?

### LINKED ISSUE #1684: Keep alive property is typed as undefined in RedisOptions
`RedisOptions` has the `keepAlive` property from `CommonRedisOptions` and from `TcpSocketConnectOpts`

The only overlapping type is `undefined` here so this can not be set on the config.

schmod:
Node 18 introduced a new `keepAlive` option in the TCP socket constructor.  

This causes ioredis's TypeScript typings to clash with `@types/node`, because we define `keepAlive` as `number | undefined`, whereas Node defines it as a `boolean`.  

Luckily, this isn't a runtime regression.  We don't pass `keepAlive` into `createConnection`, and in fact only pass in three properties total.  _But_ we pretend that any property of `TcpNetConnectOptions` can be included in `RedisOptions`.  

This PR reduces the surface-area of `RedisOptions` to only include the TCP and IPC connection parameters that we actually use, which results in `keepAlive` once again having the correct TS type.

It should also be less confusing for users, as we've removed several unused properties off of that interface...

Fixes #1680 
Fixes #1684

### DIFF
--- a/lib/connectors/StandaloneConnector.ts
+++ b/lib/connectors/StandaloneConnector.ts
@@ -4,8 +4,10 @@ import { NetStream } from "../types";
 import { CONNECTION_CLOSED_ERROR_MSG } from "../utils";
 import AbstractConnector, { ErrorEmitter } from "./AbstractConnector";
 
-export type StandaloneConnectionOptions = (Partial<TcpNetConnectOpts> &
-  Partial<IpcNetConnectOpts>) & {
+type TcpOptions = Pick<TcpNetConnectOpts, "port" | "host" | "family">;
+type IpcOptions = Pick<IpcNetConnectOpts, "path">;
+
+export type StandaloneConnectionOptions = Partial<TcpOptions & IpcOptions> & {
   disconnectTimeout?: number;
   tls?: ConnectionOptions;
 };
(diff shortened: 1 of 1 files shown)


luin:
Thanks for the contribution!

PR #1712, opened 2023-01-25
Migrate from mocha to Jest

luin:
Jest seems to be more popular than mocha and the API is nicer to work with.


silverwind:
Jest many issues related to ESM. I can recommend [vitest](https://github.com/vitest-dev/vitest) as a Jest replacement, its API is almost completely compatible with Jest.

PR #1729, opened 2023-03-06
chore(tls): update included Redis.com TLS profiles

GertSallaerts:
Redis.com will be switching out their certificates starting on April 15th 2023 with certificates signed by a different CA. This bundle is compatible with both the old as well as the future certificates. It also works for all plans.


GertSallaerts:
Sorry, I didn't see the main branch and was still on master.

Issue #159, opened 2015-09-25
Create keys with expires


pensierinmusica:
Hi,

Is it poissible to send this Redis command with `ioredis` in just one function call?

```
set key 100 ex 10
```

luin:
Yep! `redis.set('key', 100, 'ex', 10)`

pensierinmusica:
Awesome, thanks! Would it make sense to add this example to README? If you like I can create a PR  :)

luin:
That would be great :-). ioredis just passes all arguments directly to the redis server

pensierinmusica:
Done! #160 

  :)

wesleymilan:
I'm receiving an error trying to set key with expires

redis.set('key', 100, 'ex', 10);

Unhandled rejection ReplyError: ERR wrong number of arguments for 'set' command
    at ReplyParser._parseResult (/project/node_modules/ioredis/lib/parsers/javascript.js:39:14)
    at ReplyParser.execute (/project/node_modules/ioredis/lib/parsers/javascript.js:139:20)
    at Socket.<anonymous> (/project/node_modules/ioredis/lib/redis/event_handler.js:94:22)
    at Socket.emit (events.js:107:17)
    at readableAddChunk (_stream_readable.js:163:16)
    at Socket.Readable.push (_stream_readable.js:126:10)
    at TCP.onread (net.js:540:20)

luin:
@wesleymilan You should run a redis server > 2.6.12 in order to support `SET` command with `EX` option. Refer to http://redis.io/commands/set for more information.

aalexgabi:
@luin  Please add this to the API documentation.

luin:
@aalexgabi do you mean adding a note about `set` command with `ex` option is only available in redis > 2.8?

dirkbonhomme:
You can also use this dedicated command http://redis.io/commands/SETEX ;)

aalexgabi:
TL;DR Method signatures should be documented.

Disclaimer: Maybe I didn't see the part of documentation that talks about this when I "scanned" it. I didn't read the entire page. I mainly used mouse scroll and Ctrl + F (searched for "set", "ex", "command", "commands").

[...]

PR #1752, opened 2023-04-25
Create a section about expiration time on keys

ArthurPedroti:
I know it's a section about a specific command in Redis, and I don't know if this library wants to document all Redis commands.

But, as a user of this library, in my research about how I can expire my keys using ioredis, I tried to search on the docs words like "expires", "expiration", and I can't find anything, so I googled and searched on issues to finally find my answer on this issue #159 

I think creating this section will be useful for those who want to use timeout in Redis, one of the most used Redis features for people who use Redis for caching, and I think adding a simple section about it will help those users (like me), which searches on "expiration" words and finds nothing in the docs.

I know you can find it just by searching about "EX", but some users don't mind this, or don't know this specific command in Redis, and take the way to find the answer on the issue again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,16 @@ async function listenForMessage(lastId = "$") {
 listenForMessage();
 ```
 
+## Expiration
+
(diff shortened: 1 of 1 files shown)


luin:
Yeah that makes sense. It was surprisingly such a common question. Thanks for the contribution!

lilpoozie2005 (approved):

PR #1779, opened 2023-06-27
Include nodeId in key in cluster connectionpool, correctly refreshing connections when nodeId changes

### LINKED ISSUE #1778: Redis Cluster not recognizing new cluster nodes with identical hostnames
We're in the process of testing MemoryDB and it's failure cases. We've run into an issue when upgrading a cluster while our application is using ioredis.

## Problem
When changing the instance type of a MemoryDB cluster, an in-place upgrade happens where new nodes are deployed with the same hostnames as the old ones. The old ones start returning `MOVED` to all queries, and after a while they are removed.

The way ioredis reacts to this is it ends up in a loop of retrying commands and receiving `MOVED`. This loop is only broken once the old nodes are removed and the individual connections time out.

## Cause
The `MOVED` responses cause ioredis to refresh it's slots, which ends up resetting the connectionpool with the new slots:
https://github.com/redis/ioredis/blob/96a9f41b003ea8b659da45ba7852ec1eee1ab882/lib/cluster/index.ts#L902

But because the hostname and port are identical to the old nodes[^1], no new connections are made and the old ones are left intact. [^2]

## Solution

I believe this could be fixed by including the nodeId (the third return value from `CLUSTER SLOTS`[^3]) in the node keys, so that a new id will trigger a new connection.

[^1]: https://github.com/redis/ioredis/blob/d1d9dba9eafc574a9d9041fd4bc7cd04f1584159/lib/cluster/util.ts#L25-L29
[^2]: https://github.com/redis/ioredis/blob/06782e681500eae6f3ceafcc6385b9be4fdaf4e3/lib/cluster/ConnectionPool.ts#L125-L135
[^3]: https://redis.io/commands/cluster-slots/

annervisser:
Fixes #1778

### DIFF
--- a/lib/cluster/ConnectionPool.ts
+++ b/lib/cluster/ConnectionPool.ts
@@ -40,7 +40,7 @@ export default class ConnectionPool extends EventEmitter {
    * Find or create a connection to the node
    */
   findOrCreate(node: RedisOptions, readOnly = false): Redis {
-    const key = getNodeKey(node);
+    const key = this.getNodeKey(node);
     readOnly = Boolean(readOnly);
 
     if (this.specifiedOptions[key]) {
@@ -113,7 +113,7 @@ export default class ConnectionPool extends EventEmitter {
     debug("Reset with %O", nodes);
     const newNodes = {};
     nodes.forEach((node) => {
-      const key = getNodeKey(node);
+      const key = this.getNodeKey(node);
 
       // Don't override the existing (master) node
       // when the current one is slave.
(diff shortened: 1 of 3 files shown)


annervisser:
This patch is available as npm package at `@scienta/ioredis`

To replace your current install: `npm install ioredis@npm:@scienta/ioredis`

dariusungur (approved):

drewloomer:
Is there anything stopping this from being merged and released?

PR #1787, opened 2023-07-21
fix: TypeError: Cannot redefine property: default

Shyam-Chen:
```sh
/Users/path/to/node_modules/.pnpm/ioredis@5.3.2/node_modules/ioredis/built/index.js:6
Object.defineProperty(exports, "default", { enumerable: true, get: function () { return Redis_1.default; } });
       ^

TypeError: Cannot redefine property: default
```

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,5 +1,3 @@
-exports = module.exports = require("./Redis").default;
-
 export { default } from "./Redis";
 export { default as Redis } from "./Redis";
 export { default as Cluster } from "./cluster";


tobiasdiez:
@Shyam-Chen how did you fixed this issue? I get the same error:
```
node_modules/ioredis/built/index.js:6:8
      4| exports = module.exports = require("./Redis").default;
      5| var Redis_1 = require("./Redis");
      6| Object.defineProperty(exports, "default", { enumerable: true, get: fun…
       |        ^
```

Shyam-Chen:
@tobiasdiez " am using Vitest, and you can upgrade directly from Vitest v0.29 to v0.34.

tobiasdiez:
Thanks, but that still didn't work for me. I've opened https://github.com/redis/ioredis/issues/1822 to track this.

PR #1794, opened 2023-08-10
Connection URL support for Cluster constructor

### LINKED ISSUE #1519: Connection URL support for Cluster constructor
### Problem

Currently the `Cluster` constructor cannot take a url string as input. Therefore the user has to manually parse a url string obtained from the runtime environment into a structured argument which the `ioredis.Cluster` understands.

## Desired Behaviour

It would save a lot of repetitive and error-prone work if the library itself takes care of the parsing.

steprescott:
Currently to pass multiple startup nodes it can only be done by passing them as separate items in the array.

This PR adds support to allow for multiple hosts to be configured if they appear in a URL string.

### Discussion points

I've tried to be light touch and assume as little as possible but a few points to think over:

- If a url of `rediss://node-1,node-2` was given would you assume all connections are using tls?
- If a url contained options `redis://node-1,node-2/1?key=value` should it copy all options over to all nodes?
- Should we add `| string` to Cluster constructer to allow it to be created using a URL string. `Redis.Cluster("redis://node-1")`

At the moment it does none of these and just takes what is supplied and continues to keep all assumptions made by ioredis as I've only added a parsing check for commas and then separate them into their own `startupNode`.

Happy to discuss your thoughts on the approach.

Resolves https://github.com/redis/ioredis/issues/1519


steprescott:
@luin Sorry if this is not the right approach but do I need to tag you or someone else to start the review process? I wasn't sure after reading the contribution section of the `README.md`.

steprescott:
@luin just bumping this in hope it can be reviewed. If you need anything further from me please do let me know.

elena-kolevska:
Hi @steprescott, I'm sorry for the very late reply and I'm fully aware you might not be interested or able to work on the PR anymore, but I wanted to give an update anyway, for posterity.
There is an ongoing initiative to standardise the URIs across all the Redis client libraries. Once the spec is complete, we can share it here and help adapt the PR to it.

Thank you very much for your contribution 🙏

steprescott:
No problem. Thanks for the reply.

andyjones:
thank you @elena-kolevska, Are there any time frames or places to find out more about the standardisation effort? I couldn't see anything on https://github.com/redis/redis-specifications

We have a self-hosted option for our product that uses Redis so keen to align our implementation with any plans.

elena-kolevska:
Yeah, it's still WIP, but here's the current state: https://github.com/redis/redis-specifications/blob/master/uri/redis.txt

andyjones:
Thank you! Are there any plans or discussions around standardising Redis Cluster and Sentinel urls as they aren't covered by the current state?

For example:

1. will sentinel URLs have a special format? eg. redis+sentinel://
2. will cluster URLs have a separate format? eg. redis+cluster://
3. will multi-host be supported? eg. redis+cluster://redis-0,redis-1,redis-2

Appreciate that discussions are on-going so it may change but any insight is useful!

headlessme:
It would be really useful to have cluster and sentinel support in the URL standardisation, as well as multiple host support. @elena-kolevska we'd be happy to contribute to a standardisation effort, and provide a JS implementation supporting these things, or just provide feedback on the standardisation work towards this.

elena-kolevska:
> It would be really useful to have cluster and sentinel support in the URL standardisation, as well as multiple host support. @elena-kolevska we'd be happy to contribute to a standardisation effort, and provide a JS implementation supporting these things, or just provide feedback on the standardisation work towards this.

@mortensi can you provide some info here? 🙏

PR #1795, opened 2023-08-11
isInt function refactoring

0o001:


NurimOnsemiro:
I wonder if "const x = parseFloat(value);" is also unnecessary.

PR #1841, opened 2023-12-08
fix: send correct command during auto-pipelining of .call() operations

kylecannon:
This PR backports the work by @Fusl into v4 due to the bug existing there as well. https://github.com/redis/ioredis/pull/1579


kylecannon:
any chance we can get this merged and a patch release pushed out?

PR #1854, opened 2024-02-02
Default retry strategy uses exponential backoff and jitter

qsymmachus:
The current default retry strategy increases delay linearly with no jitter. This can lead to a thundering herd problem if many clients lose their connection at once, which is common during events like a Redis upgrade.

A retry strategy that uses exponential backoff plus a random jitter helps mitigate this risk and [is recommended](https://aws.amazon.com/blogs/database/best-practices-redis-clients-and-amazon-elasticache-for-redis/) in many best practices documents for Redis clients, so I propose making this the default.

The default strategy I implemented will do the following:

* Delay will be equal to `(n^2) * 50` ms where `n` is the current retry count, with a maximum value of 2000 ms.
* Jitter is a random value between 0 and 200 ms.

The millisecond values I choose are somewhat arbitrary, and I'm happy to change them if we think they're not right.

### DIFF
--- a/README.md
+++ b/README.md
@@ -761,9 +761,12 @@ using the `retryStrategy` option:
 ```javascript
 const redis = new Redis({
   // This is the default value of `retryStrategy`
-  retryStrategy(times) {
-    const delay = Math.min(times * 50, 2000);
-    return delay;
+  retryStrategy: function (times) {
+    // Generate a random jitter between 0 – 200 ms:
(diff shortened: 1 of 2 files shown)


ddixon-lattice (approved):

qsymmachus:
I light of https://github.com/redis/ioredis/issues/1870, which suggests that `node-redis` should be favored over this package, I'm closing this and opening https://github.com/redis/node-redis/pull/2736

PR #1855, opened 2024-02-09
Utilize `retryStrategy` for connection error

nodegin:
I am hosting my Redis on cloud service which has DDoS prevent mechanism,

Currently I'm using BullMQ as the queue service but it emit tons of request causing my provider throttle my Redis connection.

So I'm getting the following error:

```
ReplyError: Too many requests. Please try again later.
    at parseError (/Users/user/Desktop/project/node_modules/redis-parser/lib/parser.js:179:12)
    at parseType (/Users/user/Desktop/project/node_modules/redis-parser/lib/parser.js:302:14) {
  command: {
    name: 'auth',
    args: [ 'redis-host', 'password' ]
  }
}
```

I have added the following to my ioredis connection params:

```
  reconnectOnError: () => 2 as const,
  retryStrategy(times) {
    const delay = Math.min(times * 50, 2000);
    return delay;
  },
```

which should fixes the issue, but currently ioredis doesn't utilize retryStrategy for connection event.

So this PR is to provide support for that

### DIFF
--- a/lib/Redis.ts
+++ b/lib/Redis.ts
@@ -637,18 +637,27 @@ class Redis extends Commander implements DataHandledable {
         item.command.reject(err);
         break;
       case 2:
-        if (this.status !== "reconnecting") {
-          this.disconnect(true);
+        const self = this;
+        const resendCommand = () => {
+          if (self.status !== "reconnecting") {
+            self.disconnect(true);
+          }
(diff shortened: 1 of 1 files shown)


lukas-becker0:
You could could open the same PR in the iovalkey fork of this repo which seems to be better maintained than this repo.

PR #1864, opened 2024-03-06
fix: Improve cluster connection pool logic when disconnecting

martinslota:
# Motivation and Background

This is an attempt to fix errors occurring when a `connect()` call is made shortly after a `disconnect()`, which is something that [the Bull library does when pausing a queue](https://github.com/OptimalBits/bull/blob/60fa88f08637f0325639988a3f054880a04ce402/lib/queue.js#L869-L871).

Here's a relatively minimal way to reproduce an error:

```js
import IORedis from "ioredis";

const cluster = new IORedis.Cluster([{ host: "localhost", port: 6380 }]);

await cluster.set("foo", "bar");

const endPromise = new Promise((resolve) => cluster.once("end", resolve));
await cluster.quit();
cluster.disconnect();
await endPromise;

cluster.connect();
console.log(await cluster.get("foo"));
cluster.disconnect();
```

Running that script in a loop using 

```sh
#!/bin/bash

set -euo pipefail

while true
do
    DEBUG=ioredis:cluster node cluster-error.mjs
done
```

against the `main` branch of `ioredis` quickly results in this output:

```
/Code/ioredis/built/cluster/index.js:124
                    reject(new redis_errors_1.RedisError("Connection is aborted"));
                           ^

RedisError: Connection is aborted
    at /Code/ioredis/built/cluster/index.js:124:28

Node.js v20.11.0
```

My debugging led me to believe that the existing node cleanup logic in the `ConnectionPool` class leads to race conditions: upon [`disconnect()`](https://github.com/redis/ioredis/blob/ec42c82ceab1957db00c5175dfe37348f1856a93/lib/cluster/index.ts#L283-L304), the [this.connectionPool.reset() call](https://github.com/redis/ioredis/blob/ec42c82ceab1957db00c5175dfe37348f1856a93/lib/cluster/index.ts#L302) will [remove nodes from the pool](https://github.com/redis/ioredis/blob/ec42c82ceab1957db00c5175dfe37348f1856a93/lib/cluster/ConnectionPool.ts#L129) without cleaning up [the event listener](https://github.com/redis/ioredis/blob/ec42c82ceab1957db00c5175dfe37348f1856a93/lib/cluster/ConnectionPool.ts#L90-L96) which may then subsequently issue more than one `drain` event. Depending on timing, one of the extra `drain` events may fire after `connect()` and [change the status to `close`](https://github.com/redis/ioredis/blob/ec42c82ceab1957db00c5175dfe37348f1856a93/lib/cluster/index.ts#L153), interfering with the connection event and leading to the error above.

# Changes

* Keep track of node listeners in the `ConnectionPool` class and remove them from the nodes whenever they are removed from the pool.
* Issue `-node` / `drain` regardless of whether nodes disconnected or were removed through a `reset()` call.
* Within `reset()`, add nodes before removing old ones to avoid unwanted `drain` events.
* Fix one of the listeners by using an arrow function to make `this` point to the connection pool instance.
* Try to fix the script for running cluster tests and attempt to enable them on CI. If this doesn't work out or isn't useful, I'm happy to revert the changes.
* Add a test around this issue. The error thrown in the test on `main` is seemingly different from the error shown above but it still seems related to the disconnection logic and still gets fixed by the changes in this PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,12 +35,12 @@ jobs:
           flag-name: node-${{matrix.node}}
           parallel: true
 
-  # test-cluster:
-  #   runs-on: ubuntu-latest
-  #   steps:
-  #     - uses: actions/checkout@v2
-  #     - name: Build and test cluster
-  #       run: bash test/cluster/docker/main.sh
+  test-cluster:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build and test cluster
+        run: bash test/cluster/docker/main.sh
 
   code-coverage:
     needs: test
--- a/lib/cluster/ConnectionPool.ts
+++ b/lib/cluster/ConnectionPool.ts
@@ -7,7 +7,7 @@ const debug = Debug("cluster:connectionPool");
 
 type NODE_TYPE = "all" | "master" | "slave";
 
-type Node = {
+type NodeRecord = {
   redis: Redis;
   endListener: () => void;
   errorListener: (error: unknown) => void;
@@ -15,7 +15,7 @@ type Node = {
 
 export default class ConnectionPool extends EventEmitter {
   // master + slave = all
-  private nodes: { [key in NODE_TYPE]: { [key: string]: Node } } = {
+  private nodes: { [key in NODE_TYPE]: { [key: string]: NodeRecord } } = {
     all: {},
     master: {},
     slave: {},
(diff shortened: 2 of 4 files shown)


martinslota:
I now created [a separate repository that (hopefully) makes it easy to reproduce the bug](https://github.com/martinslota/redis-connection-race-bug).

We have been using the fix in this branch in production throughout the last roughly 3 months and it has considerably reduced the error rates we are seeing when shutting down Bull queue clients.

martinslota:
I just pushed the fixes identified in https://github.com/valkey-io/iovalkey/pull/5.

PR #1878, opened 2024-04-11
Recommend Node-Redis for new projects, clarify that ioredis will continue to be maintained.

mcohen75:
Based on discussion in https://github.com/redis/ioredis/issues/1870 it seems that Node-Redis should be used when in search of a Node.js client for Redis on a new project. The community would benefit from knowing this, which is why I'd suggest making this prominent as I have here.


kristianjaeger:
Thanks for proposing this.  I'm not sure why the Redis org has not commented / approved??  Cheers.

PavelPashov:
This is now reflected in the README, so I’m closing the PR. Thanks!

PR #1900, opened 2024-07-05
fix: fix issue of cases where prevCondition is a subscriber

prathik:
The current logic does an or operation between condition and prevCondition.

In condition the value of subscriber could be false and in prevCondition it could be a subscription set.

The change addresses this by picking the first non false subscriber.

Added some debug logs to show the case (in the image).

Screenshot 2024-07-05 at 11 40 55 PM


prathik:
@leibale @luin any advice on this?

lukas-becker0:
@prathik
luin is no longer the maintainer AFAIK and the redis team seems to focus on https://github.com/redis/node-redis,
you could move this PR to https://github.com/valkey-io/iovalkey which is the most promising fork IMO.

PR #1929, opened 2024-11-01
Add *Buffer methods typings for XCLAIM/XAUTOCLAIM/XREADGROUP/XPENDING

SiarheiFedartsou:
I noticed that there are no typings for these methods even though they are available.


lib/utils/RedisCommander.ts
```diff
@@ -8579,6 +8579,45 @@ interface RedisCommander<Context extends ClientContext = { type: "default" }> {
     justid: "JUSTID",
     callback?: Callback<unknown[]>
   ): Result<unknown[], Context>;
+  
+  xautoclaimBuffer(
```
SiarheiFedartsou, line 8583: It is simple copy-paste for non-buffer version typings - I assume it is okay, but let me know if it doesn't...

Issue #1942, opened 2025-01-07
Support NOVALUES option to hscanStream


NickZambrano:
Hello,
Since redis 7.4 HSCAN has a NOVALUES option which allow to not return the values of the fields and just return the fields names
Is it possible to implement the novalues option in the hscanStream option ?
See the documentation: https://redis.io/docs/latest/commands/hscan/
"The second element is an Array reply of field/value pairs that were scanned. When the NOVALUES flag (since Redis 7.4) is used, only the field names are returned."

PR #1943, opened 2025-01-07
feat(HscanStream): adding NOVALUES option

NickZambrano:
Adding the option NOVALUES to the hscanStream

In order to close the #1942


tishun (approved):
Looks good, thanks for the contribution!

PR #1956, opened 2025-02-20
feat: Sharded pub/sub support via dedicated subscribers

dmaier-redislabs:
The previous implementation used one subscriber connection per cluster instance, but the command `SSUBSCRIBE <channel>` needs to be executed against a node that is responsible for the hash slot that is derived from the channel name.

This PR adds a ClusterSubscriberGroup whereby each subscriber in the group is responsible for one shard.


lib/cluster/ConnectionPool.ts
```diff
@@ -36,6 +36,55 @@ export default class ConnectionPool extends EventEmitter {
...
+    return false;
+  }
+
+
+  /**
+   * Creates a Redis connection instance from the node options
+   * @param node
+   * @param readOnly
+   */
+  createRedisFromOptions(node: RedisOptions, readOnly: boolean) {
```
bobymicroby, line 64: What do you think about adding a `useShardedPubsub` option? It would be `false` by default, and would control whether we eagerly initialise the new pool of socket connections  ?

lib/cluster/ClusterSubscriberGroup.ts
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Refreshes the subscriber-related slot ranges
+     *
+     * Returns false if no refresh was needed
+     *
+     * @param cluster
+     */
+    refreshSlots(cluster: Cluster) : boolean {
+        //If there was an actual change, then reassign the slot ranges
+        if (this.clusterSlots !== cluster.slots) {
```
bobymicroby, line 109: Since `!==` only compares references, we should add a test to verify the refresh logic and confirm whether we need a deep equality check using `JSON.stringify` .

bobymicroby:
There are some tests that need attention as well:

```  
2) cluster:spub/ssub
       should receive messages:
     Uncaught TypeError: Cannot read properties of undefined (reading 'writable')
      at MockServer.write (test/helpers/mock_server.ts:139:11)
      at /home/runner/work/ioredis/ioredis/test/functional/cluster/spub_ssub.ts:26:13
      at tryCatcher (node_modules/standard-as-callback/built/utils.js:12:23)
      at /home/runner/work/ioredis/ioredis/node_modules/standard-as-callback/built/index.js:21:53
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
  3) cluster:spub/ssub
       supports password:
     Uncaught AssertionError: expected undefined to deeply equal 'abc'
      at MockServer.handler (test/functional/cluster/spub_ssub.ts:66:31)
      at JavascriptRedisParser.returnReply (test/helpers/mock_server.ts:96:54)
      at JavascriptRedisParser.execute (node_modules/redis-parser/lib/parser.js:544:14)
      at Socket.<anonymous> (test/helpers/mock_server.ts:113:16)
      at Socket.emit (node:events:517:28)
      at Socket.emit (node:domain:489:12)
      at addChunk (node:internal/streams/readable:368:12)
      at readableAddChunk (node:internal/streams/readable:341:9)
      at Socket.Readable.push (node:internal/streams/readable:278:10)
      at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
      
     ```

lib/cluster/ConnectionPool.ts
dmaier-redislabs: I think you are right. The way how it is implemented runs now one subscriber connection per shard. Which doubles the number of connections to a cluster. I added an option called "shardedSubscribers". 

BTW: This especially relevant for RESP2. The library doesn't support RESP3 yet. Later when RESP3 might be supported, users can then decide if they want to subscribe on the normal connections or if they want to use dedicated connections for the sharded subscriptions.

I added a new configuration option called `shardedSubscribers`, which defaults to `false`.

lib/cluster/ClusterSubscriberGroup.ts
dmaier-redislabs: Yes, I will investigate that.

dmaier-redislabs: OK, added a deep equality check via JSON.stringify.

PR #1966, opened 2025-03-04
Integrate latest changes from v5 branch, fix CI and tests

tishun:
### Changes to the way we build the app

- `eslint-plugin` and `parser` packages had incompatible versions, due to #1316
- rebuilt the package-lock.json with Node 8 as the new baseline
- Node 6 is failing for unknown reasons, decided to drop support as it is already quite ancient
- Node 20 added to test matrix
- CI now uses the same flow as v5 CI flow, no nested Docker worker

### Test changes
- cluster tests are now manually executed, same as v5 branch
- some tests were disabled as they test v5 functionality
- using grokzen/redis-cluster:7.0.10 because it contains the `protected-mode: no` fix
- fixed a few failing spubsub tests

### Functional changes
- pulled changes related to `sunsubscribe` from the v5 branch


dmaier-redislabs (approved):
LGTM

PR #1972, opened 2025-04-08
feat: added the option for disabling slots refresh, fixed concurrent calls to refreshSlotsCache

htemelski:
Adding the option for disabling slots refresh by setting the slotsRefreshInterval option to a negative value, similar to https://github.com/redis/ioredis/commit/370fa625cd20bfe62f41c38088e596c7a6f0619c#diff-bf62b2774a434c4a07716686a3c1fbd23368a673a7633622bdf1a9a9cd087de5
Backported bugfix: when `refreshSlotsCache` is called concurrently, call the callback only when the refresh process is done #1881


elena-kolevska:
I guess this should go to the v4 branch?

nkaradzhov (approved):
CLGTM

elena-kolevska (approved):

elena-kolevska:
Please update the docs about the '-1' value

tishun (approved):
LGTM

PR #1976, opened 2025-04-12
fix/docs(readme): javascript truthy shenanigans

joachimbulow:
This specific section is confusing. 

`indexOf` returns `-1` when the string is not found, i.e. a `truthy` value.
Hence, the docs suggest that for some reason, the `napMap` application would apply if and only if the `index` of the string `30001` is not 0.

I think it would be a clearer read to use `includes` as it more accurately aligns with an expected usecase

### DIFF
--- a/README.md
+++ b/README.md
@@ -1149,7 +1149,7 @@ const cluster = new Redis.Cluster(
   ],
   {
     natMap: (key) => {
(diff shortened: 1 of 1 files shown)


PavelPashov (approved):
LGTM

Issue #1974, opened 2025-04-11
Release workflow creates `latest` tag on previous version minor releases


elena-kolevska:
[This run](https://github.com/redis/ioredis/actions/runs/14401200085) added a `latest` tag on a 4.30.0 release, while we already have a 5.6.0 version out.

**Expected behaviour:**
A `latest` tag is only added if there aren't any higher versions published (as defined by SemVer)

nkaradzhov:
#1977 provides solution for the v4 branch. **Any other existing major release branch will have to be fixed as well.**
Turns out the semantic-release package supports [maintenance branches](https://github.com/semantic-release/semantic-release/blob/10cafbc2741e462515486e76ca6473a050b24544/docs/usage/workflow-configuration.md#maintenance-branches) out of the box. 
I propose we adopt this approach in the main branch. Benefits will be that any further major releases will work out of the box if we use semantic-release's branch naming convention ( for example 8.x, 9.x, etc ). The change will still have to be cherry-picked in all the previously created branches in order to work for them

@htemelski @elena-kolevska what do you think?

PR #1977, opened 2025-04-15
chore(release): stop overwriting npm `latest` dist-tag

nkaradzhov:
Assigns the v4 branch its own dist-tag (`v4`) and restricts it to the 4.x version range. This prevents v4 maintenance releases from overwriting the latest major version in the default npm channel.

#1974

### DIFF
--- a/.releaserc.json
+++ b/.releaserc.json
@@ -1,5 +1,11 @@
 {
-  "branches": ["v4"],
+  "branches": [
+    {
+      "name": "v4",
+      "tag": "v4",
+      "channel": "v4"
+    }
+  ],
   "verifyConditions": [
     "@semantic-release/npm",
     "@semantic-release/changelog",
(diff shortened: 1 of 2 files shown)


.releaserc.json
```diff
@@ -23,10 +29,7 @@
       ]
     }
   ],
-  "publish": [
-    "@semantic-release/npm",
-    "@semantic-release/github"
-  ],
+  "publish": ["@semantic-release/npm", "@semantic-release/github"],
```
elena-kolevska, line 32: A quick search suggests that we should use [distTag](https://docs.npmjs.com/cli/v9/commands/npm-dist-tag) so that we don't set any version to latest. Can you please verify if this makes sense?

So we would have something like:

```suggestion
  "publish": [
  ["@semantic-release/npm", {
    "npmPublish": true,
    "distTag": "v4-latest"
  }], 
  "@semantic-release/github"
],
```

htemelski-oss:
@nkaradzhov should we merge this PR?

PR #1985, opened 2025-06-10
feat(auth): add support for function for password

slukes:
When working with elasticache with IAM (and probably other providers), it is necessary to use a token that is only valid for 15 minutes as a password, as a result subsequent reconnects need to be done with a newly generated password.

There are at least these two issues discussing this:

https://github.com/redis/ioredis/issues/1738
https://github.com/redis/ioredis/issues/1810

Up until now we had been using an `on('ready,...)` event listener to manage this in our code, but since migrating from bull to bullmq which uses `duplicate` method which does not copy the event listeners and what is more applies it's own on ready listener which makes this option unusable.

Rather than continuing with the (quite hacky) work around suggested in the above two issues, I would like to propose that we accept a function instead of a string for the password.

### DIFF
--- a/lib/Redis.ts
+++ b/lib/Redis.ts
@@ -189,115 +189,122 @@ class Redis extends Commander implements DataHandledable {
 
       const { options } = this;
 
-      this.condition = {
-        select: options.db,
-        auth: options.username
-          ? [options.username, options.password]
-          : options.password,
-        subscriber: false,
-      };
+      this.resolvePassword().then((resolvedPassword) => {
+        this.condition = {
+          select: options.db,
+          auth: options.username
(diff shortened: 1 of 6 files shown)


artman:
Yes please.

slukes:
Hello 👋 any chance of a review please

MykhailoMatiiasevych:
Please review the pr, the feature is very needed

MykhailoMatiiasevych:
Would this work properly in case of reconnects?
I was thinking to bring password resolution closer to AUTH, ie in connectHandler

PavelPashov:
At first glance this looks okay, however the checks are failing

@slukes please make sure the tests pass and then request a review again

PR #1987, opened 2025-06-16
fix(ssubscribe): Force slots refresh on MOVED error when using ssubscribe

htemelski-oss:
When issuing a `ssubscribe` command, the Cluster client creates a "regular" Redis client to establish a connection to the correct shard. This doesn't work if there's a failover since the client doesn't know how to correctly handle a `MOVED` response. This PR handles such events by propagating the `ssubscribe MOVED` error from the `DataHandler`, through the `ClusterSubscriber` and the `ClusterSubscriberGroup` to the `RedisCluster` client and forces slots refresh

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,7 +45,10 @@ jobs:
         with:
           redis-version: latest
 
+      - run: npm run cluster:setup
+
       - run: npm install
       - run: npm run lint
       - run: npm run build
       - run: npm run test
+      - run: npm run test:cluster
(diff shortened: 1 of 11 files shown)


lib/DataHandler.ts
```diff
@@ -74,6 +74,11 @@ export default class DataHandler {
       args: item.command.args,
     };
 
+    if (item.command.name == "ssubscribe" && err.message.includes("MOVED")) {
+      this.redis.emit("moved");
+      return;
+    }
+
```
elena-kolevska, line 81: Can you please add a comment here explaining how we use the standalone client for `ssubscibe` connections

elena-kolevska: When a hashslot is moved to another shard the message that is sent on the pubsub channel is `sunsubscribe`, not `moved`. I suspect this is working because it's caught when the connection is retried. On issuing the `ssubscribe` command, the response is `moved`.

elena-kolevska (changes requested):
We need tests on this. Please fix and enable the existing ones that weren't running before

PR #1991, opened 2025-06-19
Update README.md to accurately describe silent `error` event

loucadufault:
Revise doc changes in #1990

The initial docs (pre that change) for the `error` event were not terribly accurate, part of that was addressed in #1990, but  introduced some inaccuracy for this part.

> emits when an error occurs while connecting.
When `lazyConnect` is false\*, ioredis emits all `error` events silently (only emits when there's at least one listener) so that your application won't crash if you're not listening to the error event.
However when `lazyConnect` is true\** and `redis.connect()` is explicitly called it will return a promise which might reject with an error. If redis.connect()\*** is not called explicitly, ioredis will try to connect automatically on the first command and emit the error event silently.

\*: not only when `lazyConnect` is false, also when it is true.
\**: not only when `lazyConnect` is true, in fact calling redis.connect() at any tick after `new Redis()`, regardless of whether `lazyConnect` is true/false, will reject with the connection error.
\***: and if and only if `lazyConnect` is true (which may be implicit from the second sentence).

In general I am not sure it is worth mentioning the promise rejection behaviour in this event description, maybe we can remove the second sentence altogether and just keep the first and third.


PavelPashov:
Hey @loucadufault this seems like a valid clarification.

PavelPashov (approved):

PR #2027, opened 2025-10-11
docs(example): add example for setting TTL to "HSET Field"

ironnicko:
would make sense to have an example here.


ironnicko:
> Looks good

Thanks for the feedback.

Issue #2026, opened 2025-10-11
Automatic selection of IP family


slagiewka:
The default has been set in the initial commit: https://github.com/redis/ioredis/blob/b8ea9b8367e4eb6550822b74fdc263d318c5286a/lib/redis/mixin/default_options.js#L11

As a result, any project that that effectively uses `ioredis` defaults is only using IPv4 even when IPv6 is available. This gets worse when Redis is available in IPv6 only.

As an operator for OSS projects like [Immich](https://github.com/immich-app/immich/pull/22863) or [Ghostfolio](https://github.com/ghostfolio/ghostfolio/issues/5721) I can't get automatic upgrades to IPv6 - manual intervention is required. If the project is flexible around providing custom DSN this is only partially bad. If project only handles parts of connection settings - updating the project is required.

Database drivers (e.g. for PostgreSQL) or `@redis/client` don't suffer from this limitation and moving to IPv6 is hassle free.

slagiewka:
It is worth noting that node,js has implemented [RFC6724](https://datatracker.ietf.org/doc/html/rfc6724) via c-ares since node v14.

https://github.com/nodejs/node/commit/4cfcaaeef38031ffb4b87296702508b45dbb4222

slagiewka:
At the time of initial commit, node's docs have been informing users that there are two possible values: `4` and `6`. This has been fixed since as this was not correct. https://github.com/nodejs/node/commit/0777e090b4128b582abdee26a3baa70b49a32995

Another honourable mention is `autoSelectFamily` implementing [RFC8305](https://www.rfc-editor.org/rfc/rfc8305.txt) that gets completely ignored if `family` is not `0`. The functionality is in node since v19 and also available in v18.

slagiewka:
Honourable mentions of `ioredis` still confusing users:
https://cycle.io/community/threads/6837b669c09c5118b34d9be3/gotcha-with-redis-and-ipv6
https://community.n8n.io/t/allow-ipv6-to-use-redis-with-provider-internal-network/32439/3
https://station.railway.com/questions/private-network-issue-6e5aba86

PavelPashov:
Hey @slagiewka, thank you for fixing this in #2028, we'll make sure to include it in the next release

RedPhoenixQ:
[...]

PR #2028, opened 2025-10-12
feat: default IP family selection to 0

slagiewka:
A 30th birthday gift to IPv6.

Closes: #2026


PavelPashov:
Hey @slagiewka, do you mind if we consider this a `fix` instead of a `feat`, since I believe this change shouldn't result in a new minor version?

slagiewka:
> Hey @slagiewka, do you mind if we consider this a `fix` instead of a `feat`, since I believe this change shouldn't result in a new minor version?

I'll leave it up to you. In fact, I would consider this fixing behaviour as well.

PR #2029, opened 2025-10-17
fix:maintain_setinfo_command_order

ntvviktor:
Attempt to split the order of `private setStatus()` function to allow `sendCommand` with setting client info metadata before Redis client would execute other users' command.

Try to resolve this issue https://github.com/redis/ioredis/issues/2025

### DIFF
--- a/lib/redis/event_handler.ts
+++ b/lib/redis/event_handler.ts
@@ -120,7 +120,7 @@ function abortError(command: Respondable) {
 // aborting and purging we'll have a queue that looks like this: [0, 1, 2]
 function abortIncompletePipelines(commandQueue: Deque<CommandItem>) {
   let expectedIndex = 0;
-  for (let i = 0; i < commandQueue.length; ) {
+  for (let i = 0; i < commandQueue.length;) {
     const command = commandQueue.peekAt(i)?.command as Command;
     const pipelineIndex = command.pipelineIndex;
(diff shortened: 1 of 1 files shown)


ntvviktor:
Hi @PavelPashov, let me know if this PR can help with the issue

PavelPashov:
> Hi @PavelPashov, let me know if this PR can help with the issue

Hi @ntvviktor, thanks for your help with this issue, however it looks like the change has affected the CI. I’m also exploring a few approaches to resolve it.

ntvviktor:
@PavelPashov Yep, seems like my solution will not work well and compatible with other tests, looking forward to seeing yours, the underlining command executed and events replied from redis server is more complicated than it looks

PavelPashov:
Hey @ntvviktor, feel free to take a peek at #2033, however please note that it might change, let me know if you have any suggestions

ntvviktor:
Hey @PavelPashov, nice catch, btw I just wonder is this due to the root cause that Dragonfly DB is so fast and multi-thread that current approach of executing commands and replies would have problems?

PavelPashov:
> Hey @PavelPashov, nice catch, btw I just wonder is this due to the root cause that Dragonfly DB is so fast and multi-thread that current approach of executing commands and replies would have problems?

 I’m not able to confirm on whether that’s the underlying cause, but it’s an interesting hypothesis.

PR #2044, opened 2025-12-02
Fix Sentinel URL in README

andrewhathaway:
Noticed the Sentinel URL in the README was broken, updated it to match the scaling destination.


PavelPashov (approved):

PR #2050, opened 2025-12-04
fix: fix timeout for blocking commands during disconnections

### LINKED ISSUE #1888: BZPOPMIN will hang for ever when a short disconnection happens and the port is different from 6379
This one was pretty weird but quite serious. If you issue a BZPOPMIN command, and there is a small disconnection while the command is blocking, then it will stay blocking forever, effectively hanging the app that uses the command. 

Now, one of the *insane* things about this issue is that to reproduce it you must use a port different than the standard 6379, and also use docker network disconnect, it is not enough to just stop the docker container running Redis:

Try this code:
bug.mjs:
```ts
import Redis from "ioredis";

const connection = new Redis({
    host: "localhost",
    port: 6380,
});

connection.on("connect", () => console.log("Redis connected!"));
connection.on("ready", () => console.log("Redis ready!"));
connection.on("error", (err) => console.error("Redis error:", err));
connection.on("end", () => console.log("Redis connection ended"));
connection.on("reconnecting", () => console.log("Redis reconnecting..."));

async function test() {
    console.log("Gonna issue a BZPOPMIN command...")
    const result = await connection.bzpopmin("key", 0)
    console.log("Issued BZPOPMIN command!", result)
}

test().catch(console.error);
```

docker-compose.yml:
```yaml
version: "3.7"
services:
  redis:
    image: redis:alpine
    ports:
      - "6380:6379"
```

Start the docker container with:
```
docker-compose up
```

Run the test with:
```sh
node bug.mjs
```
Results in:
```sh
Gonna issue a BZPOPMIN command...
Redis connected!
Redis ready!
```

Now in a different terminal, and depending on where you run the docker (I had my docker compose on directory ioredis-econnreset:
```sh
docker network disconnect ioredis-econnreset_default ioredis-econnreset-redis-1
sleep 3
docker network connect ioredis-econnreset_default ioredis-econnreset-redis-1
```

The Redis network was disconnected and connected again after 3 seconds. Open a Redis cli:
```sh
redis-cli -p 6380
> 127.0.0.1:6380> zadd key 10 test
(integer) 1
127.0.0.1:6380>
```

The terminal with the node app will stay the same, with no error or nothing. If you restart the app and without disconnecting and reconnecting you try again to ZADD to the key you will get instead:
```sh
Gonna issue a BZPOPMIN command...
Redis connected!
Redis ready!
Issued BZPOPMIN command! [ 'key', 'test', '10' ]
```

Let me know if you need more information.

manast:
Fixes #1888

### DIFF
--- a/lib/Redis.ts
+++ b/lib/Redis.ts
@@ -68,6 +68,23 @@ class Redis extends Commander implements DataHandledable {
    * Default options
    */
   private static defaultOptions = DEFAULT_REDIS_OPTIONS;
+  private static readonly TAIL_TIMEOUT_COMMANDS = new Set([
+    "blpop",
+    "brpop",
+    "brpoplpush",
+    "blmove",
+    "bzpopmin",
+    "bzpopmax",
+  ]);
+  private static readonly HEAD_TIMEOUT_COMMANDS = new Set([
+    "bzmpop",
+    "blmpop",
+  ]);
+  private static readonly BLOCK_OPTION_COMMANDS = new Set([
+    "xread",
+    "xreadgroup",
+  ]);
+  private static readonly BLOCKING_TIMEOUT_GRACE_MS = 1000;
 
   /**
    * Create a Redis instance.
@@ -107,6 +124,8 @@ class Redis extends Commander implements DataHandledable {
   private retryAttempts = 0;
   private manuallyClosing = false;
   private socketTimeoutTimer: NodeJS.Timeout | undefined;
+  private blockingCommandTimer: NodeJS.Timeout | undefined;
+  private activeBlockingCommand: Command | undefined;
 
   // Prepare autopipelines structures
   private _autoPipelines = new Map();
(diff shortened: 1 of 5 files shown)


lib/Redis.ts
```diff
@@ -125,6 +144,7 @@ class Redis extends Commander implements DataHandledable {
     this.parseOptions(arg1, arg2, arg3);
 
     EventEmitter.call(this);
+    this.on("close", () => this.clearBlockingCommandTimeout());
```
PavelPashov: It would be better to move this inside the [closerHandler](https://github.com/redis/ioredis/blob/a1e7d0203e0642e9bbe874b1ed21798ee4454a8d/lib/redis/event_handler.ts#L199-L200)

PavelPashov:
Hey @manast, thank you for the PR, however it seems to break the CI and doesn't follow many of the established patterns in the repository so I created #2051 instead.

PavelPashov:
Closing this as #2052 has been merged

PR #2059, opened 2026-01-12
Fixed issue with moved command for replicas

DevandScorp:
Hi, I was fighting with the issue `All keys in the pipeline should belong to the same slots allocation group` for quite a lot of time.
Obviously this validation message can appear when smth is happening on the cluster and ioredis need some time to recalculate slots mapping. But in my case the cluster remained stable, and this validation messages were randomly shown on different instances of the application.
My PR has fixed the issue for the following scenario:
1. My Redis setup is: enabled sharding + read replica for each master node.
2. `scaleReads` = `all` + `enableAutoPipelining` = `true`
3. ioredis is executing a https://redis.io/docs/latest/commands/readonly/ command for all read replicas in order to avoid `MOVED` messages.
4. Sometimes `READONLY` command can be executed after several get operations already made on the replica(just some race condition, which can easily happen when we have `enableOfflineQueue` and `lazyConnect` features enabled). In this case I receive `MOVED <slot-number> <master-node-host>` response.
5. `MOVED` command is not crucial, but the way moved command is handled is. Currently ioredis is rewriting a slot map without checking what is happening in the slot. With read replicas slot map's value is ['master-node-host', ...<replica-nodes-hosts>]. After rewriting it with the master's node address only future validations will fail on `generateMultiWithNodes` before a slots map refresh.

I hope this fix will help a lot of users to stop receiving issues on the same setup I've described.

### DIFF
--- a/lib/cluster/ClusterSubscriberGroup.ts
+++ b/lib/cluster/ClusterSubscriberGroup.ts
@@ -105,12 +105,18 @@ export default class ClusterSubscriberGroup {
   }
 
   /**
-   * Disconnect all subscribers
+   * Disconnect all subscribers and clear some of the internal state.
    */
   stop() {
     for (const s of this.shardedSubscribers.values()) {
       s.stop();
     }
+
+    // Clear subscriber instances and pending operations.
+    // Channels are preserved for resubscription on reconnect.
+    this.pendingReset = null;
+    this.shardedSubscribers.clear();
+    this.subscriberToSlotsIndex.clear();
   }
 
   /**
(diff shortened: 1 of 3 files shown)


DevandScorp:
@PavelPashov hi, possibly my PR should fix this issue https://github.com/redis/ioredis/issues/1734(sorry, I'm not sure that I know the way to attach this issue to the PR).

PavelPashov:
@DevandScorp Thanks for the PR, looks good overall. CI is currently failing due to the new tests. Can you take a look and push a fix?

DevandScorp:
@PavelPashov hi, I made a fix, hopefully it will work.

DevandScorp:
@PavelPashov hi, it seems to me that all tests have passed, is there anything else needed from my side?

PavelPashov (dismissed):

lib/Pipeline.ts
```diff
@@ -164,7 +164,11 @@ class Pipeline extends Commander<{ type: "pipeline" }> {
         cluster.handleError(commonError, this.leftRedirections, {
           moved: function (_slot: string, key: string) {
             _this.preferKey = key;
-            cluster.slots[errv[1]] = [key];
+            if (cluster.slots[errv[1]]) {
+                if (cluster.slots[errv[1]][0] !== key) {
+                  cluster.slots[errv[1]] = [key];
+                }
+            }
```
PavelPashov: Don't we need an `else` statement here to add the slot if it doesn't exist, similar to [this](https://github.com/redis/ioredis/blob/def9804dd44faa13dd57347c7353142ec0bd2d8f/lib/cluster/index.ts#L541)

PR #2064, opened 2026-01-15
fix: fix issue with moved command for replicas

DevandScorp:
Hi, I was fighting with the issue `All keys in the pipeline should belong to the same slots allocation group` for quite a lot of time.
Obviously this validation message can appear when smth is happening on the cluster and ioredis need some time to recalculate slots mapping. But in my case the cluster remained stable, and this validation messages were randomly shown on different instances of the application.
My PR has fixed the issue for the following scenario:
1. My Redis setup is: enabled sharding + read replica for each master node.
2. `scaleReads` = `all` + `enableAutoPipelining` = `true`
3. ioredis is executing a https://redis.io/docs/latest/commands/readonly/ command for all read replicas in order to avoid `MOVED` messages.
4. Sometimes `READONLY` command can be executed after several get operations already made on the replica(just some race condition, which can easily happen when we have `enableOfflineQueue` and `lazyConnect` features enabled). In this case I receive `MOVED <slot-number> <master-node-host>` response.
5. `MOVED` command is not crucial, but the way moved command is handled is. Currently ioredis is rewriting a slot map without checking what is happening in the slot. With read replicas slot map's value is ['master-node-host', ...<replica-nodes-hosts>]. After rewriting it with the master's node address only future validations will fail on `generateMultiWithNodes` before a slots map refresh.

I hope this fix will help a lot of users to stop receiving issues on the same setup I've described.

### DIFF
--- a/test/functional/monitor.ts
+++ b/test/functional/monitor.ts
@@ -15,11 +15,15 @@ describe("monitor", () => {
           return;
         }
         monitor.on("monitor", function (time, args) {
-          expect(args[0]).to.match(/get/i);
-          expect(args[1]).to.eql("foo");
-          redis.disconnect();
-          monitor.disconnect();
-          done();
+          // Filter out handshake commands (client, info, select, auth, readonly)
+          // and keep-alive commands (ping) - only process the actual 'get' command
+          if (args[0] && args[0].toLowerCase() === "get") {
+            expect(args[0]).to.match(/get/i);
+            expect(args[1]).to.eql("foo");
+            redis.disconnect();
+            monitor.disconnect();
+            done();
+          }
         });
 
         await waitForMonitorReady();


DevandScorp:
@PavelPashov pls take a look, now it seems to me that I've fixed all places, where any flaky behavior can happen. Sorry for such a big amount of commits

PavelPashov:
Hey @DevandScorp, the tests don't fail now, however I decided to check them against the `main` branch without the fix and it seems that the second test never fails, while the first one will fail ~ 50% of the time. I'm not convinced that the second test is worth keeping. If you don't mind I'll keep this PR open for now to check if I can reproduce the issue and see if we can have a better test for this.

DevandScorp:
@PavelPashov sure, can you pls attach failed tests, I'll try to figure out the reason of failure

PavelPashov:
> @PavelPashov sure, can you pls attach failed tests, I'll try to figure out the reason of failure

To clarify: the concern isn’t that the tests are failing now, but that they aren’t failing when they should. I would expect the test to reliably fail without your fix. However, even without the fix, `should not throw 'All keys in the pipeline should belong to the same slots allocation group' with scaleReads=all and autopipelining` never fails. That suggests the test is consistently producing a false positive and isn’t validating the intended behavior.

DevandScorp:
@PavelPashov I think I was able to rewrite the tests. They were too complicated and as you mentioned before, one of this tests was not failing at all. So I wrote one test, where I use scaleReads=`slave` to be 100% sure that my request will go to the read replica. After that I execute two pipelines: first one should trigger MOVED command and slots map should be updated. Another piipeline will just trigger the validation. WIth previous implementation it will fail. In my opinion this test is much cleaner.
Unfortunately if slots map is refreshed before second pipeline is executed, the test can pass on the main branch. I don't know how can I prevent such race condition. Maybe you have some thoughts on that.

PavelPashov:
> @PavelPashov I think I was able to rewrite the tests. They were too complicated and as you mentioned before, one of this tests was not failing at all. So I wrote one test, where I use scaleReads=`slave` to be 100% sure that my request will go to the read replica. After that I execute two pipelines: first one should trigger MOVED command and slots map should be updated. Another piipeline will just trigger the validation. WIth previous implementation it will fail. In my opinion this test is much cleaner. Unfortunately if slots map is refreshed before second pipeline is executed, the test can pass on the main branch. I don't know how can I prevent such race condition. Maybe you have some thoughts on that.

Thank you for re-writing the tests, I'm going to review them and check if we can work around the race condition.

DevandScorp:
@PavelPashov hi, I think I was able to finally understand the way I can avoid any race conditions: if I just use pipeline's retry logic on moved, for more than 1 key it will reexecute pipeline for all keys. This tests fails all the time for me on the main branch and does not fail on my branch, so I think this is it. Can you pls take a look and check it from your side? For me it looks like a final solution, which should suit our needs. 
At the same time I'm a little bit concerned about this NAT test, which I've never faced before locally or in CI, probably it is not related to my tests.

DevandScorp:
@PavelPashov hi, is there anything else I could do from my side? Sorry for being too annoying, this fix worked for us  and we will be happy to use it in official ioredis version.

DevandScorp:
Also fixed some monitoring tests, which were flaky

PavelPashov:
> @PavelPashov hi, is there anything else I could do from my side? Sorry for being too annoying, this fix worked for us and we will be happy to use it in official ioredis version.

@DevandScorp Thanks for the update, glad to hear the fix is working for you. I have this queued for re-review and will take a look as soon as I can. If everything checks out, we can get it into an official ioredis release.

PR #2072, opened 2026-01-30
Fix autopipelining to route writes to masters with scaleReads

ivoronin:
When `scaleReads` is set to `slave` or `all` and `enableAutoPipelining` is enabled, writes may end up going to replicas instead of masters.

The problem is that autopipelining groups commands by slot into a single pipeline, and that pipeline goes to whatever node `scaleReads` picks - which can be a replica.

This causes the whole pipeline to fail and retry with `MOVED` redirects until by luck it lands on a master, stacking up latency. In the worst case (after 16 failed attempts) these `MOVED` errors bubble up to user code.

The fix separates read and write commands into different pipelines when `scaleReads` is not master. Read commands get a `:read` suffix on their slot key, writes get `:write`. This way they end up in separate pipelines that route correctly.

### DIFF
--- a/lib/autoPipelining.ts
+++ b/lib/autoPipelining.ts
@@ -1,6 +1,7 @@
 import { isArguments, noop } from "./utils/lodash";
 import * as calculateSlot from "cluster-key-slot";
 import asCallback from "standard-as-callback";
+import { exists, hasFlag } from "@ioredis/commands";
 import { ArgumentType } from "./Command";
 
 export const kExec = Symbol("exec");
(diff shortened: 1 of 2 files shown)


PavelPashov (approved):
This looks good, thank you for your contribution

ivoronin:
@PavelPashov Could I ask you for a minor release?

ivoronin:
@PavelPashov Could I ask you for a minor release?

PavelPashov:
> @PavelPashov Could I ask you for a minor release?

@ivoronin I'm planning to do a minor release next week, as there're other fixes/features that would be great if included

PR #2089, opened 2026-03-11
feat: Implement `TracingChannel` support

logaretm:
This PR implements tracing channels for `command` and `connect` operations. This allows APM libraries (OTEL, Datadog, Sentry) to instrument redis commands **without monkey-patching**.

It also enables users to setup their own custom instrumentations and analytics.

This is a port of redis/node-redis#3195 to ioredis, based on the proposal in redis/node-redis#2590.

### Tracing Channels

All channels in this PR use the Node.js [`TracingChannel`](https://nodejs.org/api/diagnostics_channel.html#class-tracingchannel) API, which provides `start`, `end`, `asyncStart`, `asyncEnd`, and `error` sub-channels automatically.

| TracingChannel | Tracks | Context fields |
|---|---|---|
| `ioredis:command` | Individual Redis command execution from queue to reply (standalone, MULTI, and pipeline) | `command`, `args`, `database`, `serverAddress`, `serverPort`, and optionally `batchMode`, `batchSize` for MULTI/pipeline commands |
| `ioredis:connect` | Client connection lifecycle (initial connections and reconnections) | `serverAddress`, `serverPort`, `connectionEpoch` |

### Context Properties

The payload sent in the trace events closely follows what OTEL currently extracts as attributes in their instrumentations:

| Field | Source | OTEL attribute it enables |
|---|---|---|
| `command` | command name | `db.operation.name` |
| `args` | full args array | `db.query.text` (APM serializes/sanitizes) |
| `database` | selected DB | `db.namespace` |
| `serverAddress` | socket options host (or `path` for IPC) | `server.address` |
| `serverPort` | socket options port (`undefined` for IPC) | `server.port` |

For batch commands we include these properties:

| Field | Source | OTEL attribute it enables |
|---|---|---|
| `batchMode` | `'PIPELINE'` or `'MULTI'` | `db.operation.name` prefix |
| `batchSize` | number of user commands (excludes MULTI/EXEC protocol wrappers) | `db.operation.batch.size` |

For connect events:

| Field | Source | Description |
|---|---|---|
| `connectionEpoch` | internal reconnection counter | `0` for initial connect, `1+` for reconnections. APMs can use this to distinguish initial connections, detect flapping, or correlate traces across reconnection cycles. |

### Backward Compatibility

Zero-cost when no subscribers are registered. Silently skipped on Node.js versions where `TracingChannel` is unavailable (e.g. Node 18).

```ts
const dc = ('getBuiltinModule' in process)
  ? process.getBuiltinModule('node:diagnostics_channel')
  : require('node:diagnostics_channel');
```

## Usage Examples

Consumers will be doing something like this:

```ts
  const dc = require('node:diagnostics_channel');

  // Trace individual Redis commands
  dc.tracingChannel('ioredis:command').subscribe({
    start(ctx) {
      console.log(`→ ${ctx.command}`, ctx.args, `db=${ctx.database}`);
      // APM: create span here
      // ctx.span = tracer.startSpan(`redis ${ctx.command}`, { attributes: {
      //   'db.system': 'redis',
      //   'db.operation.name': ctx.command,
      //   'db.namespace': String(ctx.database),
      //   'server.address': ctx.serverAddress,
      //   'server.port': ctx.serverPort,
      // }});
    },
    asyncEnd(ctx) {
      console.log(`✓ ${ctx.command} done`);
      // ctx.span?.end();
    },
    error(ctx) {
      console.error(`✗ ${ctx.command} failed:`, ctx.error?.message);
      // ctx.span?.setStatus({ code: SpanStatusCode.ERROR });
    },
  });

  // Trace connection
  dc.tracingChannel('ioredis:connect').subscribe({
    start(ctx) {
      console.log(`Connecting to ${ctx.serverAddress}:${ctx.serverPort} (epoch: ${ctx.connectionEpoch})`);
    },
    asyncEnd() {
      console.log('Connected');
    },
    error(ctx) {
      console.error('Connection failed:', ctx.error?.message);
    },
  });

  // Works with all ioredis operations — no monkey-patching needed
  const redis = new Redis();
  await redis.set('foo', 'bar');          // ioredis:command { command: 'SET', args: ['foo', 'bar'] }
  await redis.get('foo');                  // ioredis:command { command: 'GET', args: ['foo'] }

  // Pipeline/MULTI commands include batch metadata
  const pipeline = redis.pipeline();
  pipeline.set('a', '1');
  pipeline.set('b', '2');
  await pipeline.exec();                  // ioredis:command { command: 'SET', batchMode: 'PIPELINE', batchSize: 2 }
```

### Future Improvements

- **Offline queue tracing channel**: A dedicated `ioredis:command:queue` channel could emit events when commands enter the offline queue, allowing APMs to measure queue wait time separately from command execution time and track how many commands hit the offline queue (a signal of connection instability). Note that the existing OTEL monkey-patch instrumentation (`@opentelemetry/instrumentation-ioredis`) has the same gap — it doesn't distinguish queued vs. written commands either.
- **Sentinel-aware server address**: `_getServerAddress()` currently uses configured `host`/`port` options. For sentinel setups, the trace context could reflect the actual resolved endpoint rather than the configured defaults.

---

> [!NOTE]
> **Medium Risk**
> Touches core `Redis.connect()` and `sendCommand()` paths to wrap operations in tracing helpers; while intended to be zero-overhead when unused, mistakes could affect command execution flow or error handling. Adds argument sanitization rules to avoid leaking secrets, but incorrect redaction logic could still expose or over-redact data.
> 
> **Overview**
> Adds Node.js `diagnostics_channel`/`TracingChannel` instrumentation for ioredis operations, emitting telemetry for **individual commands** (`ioredis:command`), **MULTI batches** (`ioredis:batch`), and **connection attempts** (`ioredis:connect`) when subscribers are present (silently no-op otherwise).
> 
> `Redis.connect()` is now traced via `traceConnect`, and `sendCommand()` wraps the written command promise with `traceCommand` exactly once per command (using a new `Command.isTraced` flag) to avoid duplicate events during offline-queue flushes or command resends after reconnect.
> 
> Introduces `lib/tracing.ts` with safe-by-default argument sanitization (redacting sensitive values via `sanitizeArgs`) and exported context types, updates docs with subscription examples, and adds comprehensive functional tests validating tracing behavior, deduplication, error tracing, and lack of unhandled rejections.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 79d3220c1406b0eb78688be2eff06c0e19593223. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -868,7 +868,7 @@ Besides the above connection events, there are several other custom events:
 
 ioredis publishes telemetry through Node.js [`diagnostics_channel`](https://nodejs.org/api/diagnostics_channel.html), allowing APM tools and custom instrumentation to observe commands, connections, and batch operations without modifying application code.
 
-These channels use `TracingChannel#tracePromise()` and emit `start`, `end`, `asyncStart`, `asyncEnd`, and `error` sub-events. Requires Node.js >= 20.13.0; on older versions the channels are silently unavailable (zero overhead). Subscribe via `tracing:<name>:<event>`:
+These channels use `TracingChannel#tracePromise()` and emit `start`, `end`, `asyncStart`, `asyncEnd`, and `error` sub-events. Requires Node.js >= 18.19.0; on older versions the channels are silently unavailable (zero overhead). Subscribe via `tracing:<name>:<event>`:
 
 ```typescript
 import dc from "node:diagnostics_channel";
--- a/README.md
+++ b/README.md
@@ -729,13 +729,16 @@ const stream = redis.zscanStream("myhash", {
   match: "age:??",
 });
 ```
+
 The `hscanStream` also accepts the `noValues` option to specify whether Redis should return only the keys in the hash table without their corresponding values.
+
 ```javascript
 const stream = redis.hscanStream("myhash", {
   match: "age:??",
   noValues: true,
 });
 ```
+
 You can learn more from the [Redis documentation](http://redis.io/commands/scan).
 
 **Useful Tips**
@@ -815,6 +818,7 @@ const redis = new Redis({
 ```
 
 When enabled:
+
 - For commands with a finite timeout (e.g., `blpop("key", 5)`), ioredis sets a client-side deadline based on the command's timeout plus a small grace period (`blockingTimeoutGrace`, default 100ms). If no reply arrives before the deadline, the command resolves with `null`—the same value Redis returns when a blocking command times out normally.
 - For commands that block forever (e.g., `timeout = 0` or `BLOCK 0`), the `blockingTimeout` value is used as a safety net.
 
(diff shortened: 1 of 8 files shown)


logaretm:
@tlhunter @Qard I followed the same patterns here as well.

PavelPashov:
@logaretm Thanks for the PR. I’ll take a look when I get a chance and follow up with feedback.

logaretm:
@PavelPashov Thanks! happy to work as much as needed to get this through as long as no one is opposed to the idea 🙏 

P.S: I created a PR for `node-redis` in https://github.com/redis/node-redis/issues/2590 which had the original issue discussion.

logaretm:
@PavelPashov No rush, but let me know if we can jump on a call to discuss the PR if that may help move things along. Otherwise, I will be waiting for your 👀

Qard (approved):

PavelPashov:
> @PavelPashov No rush, but let me know if we can jump on a call to discuss the PR if that may help move things along. Otherwise, I will be waiting for your 👀

Hey @logaretm,

I'm on the official [Redis discord server](https://discord.gg/s42Y9vKn), feel free to message me @pavelpashov_16203.

I answered in this [thread](https://github.com/redis/node-redis/pull/3195#issuecomment-4110416143) let's continue the github discussion there.

logaretm:
@PavelPashov i brought this PR to match node-redis one, it includes sanitization and the adjustments to the batch channel. Also added a doc section for it.

PavelPashov:
@logaretm I think that a plain pipeline is mostly a transport/write optimization. It changes how commands are flushed, but it does not create a higher-level semantic operation boundary. So I do not think we should promote pipeline membership into batch semantics. (Likely the pipeline batch type will be removed from `node-redis` as well)

`MULTI` is different. In the default `pipeline: true` path, ioredis is explicitly building a local `MULTI ... EXEC` operation and flushing it as one unit. That is the one place where a batch span makes sense semantically.

Because of that, I think the batch tracing should live in [transaction.ts](https://github.com/redis/ioredis/blob/main/lib/transaction.ts), not in the generic pipeline executor. For example:

```ts
      const batchSize = Math.max(this.length - 2, 0);
      for (let i = 0; i < this._queue.length; ++i) {
        this._queue[i].batchMode = "MULTI";
      }
      const promise = traceBatch(
        () => exec.call(pipeline),
        () => buildBatchContext(this.redis, batchSize)
      );
```

This keeps the model straightforward:

- standalone commands: command traces
- commands inside `pipeline()`: still command traces
- `multi()` / `multi({ pipeline: true })`: one `MULTI` batch trace
- `multi({ pipeline: false })`: no batch trace unless we explicitly decide to model connection-scoped transactions separately

I think that separation is both cleaner semantically and simpler to reason about than trying to infer batch semantics from generic pipeline execution.

Bottom line: lets remove batch tracing from [Pipeline.ts](https://github.com/redis/ioredis/blob/main/lib/Pipeline.ts) and keep `MULTI` batch tracing in [transaction.ts](https://github.com/redis/ioredis/blob/main/lib/transaction.ts).

PavelPashov (approved):
LGTM

PR #2093, opened 2026-04-04
fix(test): fully resolve spub/ssub auto-resubscription  
  race condition

rit3sh-x:
## Summary
  - Follow-up to #2092                                                                               
  - Await initial ssubscribe calls so channels are registered before disconnect
  - Poll for stub.callCount >= 3 instead of relying on ping() as a barrier

  ## Test plan
  - [x] All 555 tests passing, 0 failures
  - [x] Stable under nyc coverage overhead

---

> [!NOTE]
> **Low Risk**
> Low risk since changes are limited to functional tests, mainly addressing timing/race issues around async `ssubscribe` and reconnect handling.
> 
> **Overview**
> Fixes `spub/ssub` functional test flakiness by ensuring the publisher connection is always disconnected when both `smessage` and `smessageBuffer` handlers complete.
> 
> Stabilizes the auto-resubscribe test by `await`ing initial `ssubscribe` calls before forcing a reconnect and by asserting the expected `ssubscribe` stub calls after the `ready` event (instead of using extra commands as a timing barrier).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 54097d9f463cdaed1199ec5e374c42f535b0771f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/test/functional/spub_ssub.ts
+++ b/test/functional/spub_ssub.ts
@@ -84,6 +84,7 @@ describe("spub/ssub", function () {
       expect(channel).to.eql("foo");
       expect(message).to.eql("bar");
       if (!--pending) {
+        pub.disconnect();
         redis.disconnect();
         done();
       }
@@ -94,6 +95,7 @@ describe("spub/ssub", function () {
       expect(message).to.be.instanceof(Buffer);
       expect(message.toString()).to.eql("bar");
       if (!--pending) {
+        pub.disconnect();
         redis.disconnect();
         done();
       }
@@ -115,15 +117,10 @@ describe("spub/ssub", function () {
     });
   });
 
-  // TODO ready reconnect in redis stand
   it("should restore subscription after reconnecting(ssubscribe)", (done) => {
-    const redis = new Redis({ port: 6379, host: "127.0.0.1" });
-    const pub = new Redis({ port: 6379, host: "127.0.0.1" });
-    // redis.ping(function (err, result) {
-    //   // redis.on("message", function (channel, message) {
-    //   console.log(`${err}-${result}`);
-    //   // });
-    // });
+    // It defaults to port: 6379, host: 127.0.0.1
+    const redis = new Redis();
+    const pub = new Redis();
     redis.ssubscribe("foo", "bar", function () {
       redis.on("ready", function () {
         // Execute a random command to make sure that `subscribe`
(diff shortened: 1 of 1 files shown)


PavelPashov:
@rit3sh-x thanks, I'll take a look

rit3sh-x:
Hi @PavelPashov, could you please share the status of this PR? Let me know if any changes are needed from my side.

test/functional/spub_ssub.ts
```diff
@@ -165,6 +168,17 @@ describe("spub/ssub", function () {
...
+    // may not have fired all 3 ssubscribe calls yet. Poll until the stub
+    // records all 3 calls; Mocha's test timeout handles the failure case.
+    await new Promise<void>((resolve) => {
+      const check = () => {
+        if (stub.callCount >= 3) return resolve();
+        setTimeout(check, 50);
+      };
+      check();
+    });
+
```
PavelPashov: This has the potential to make the test really slow without much benefit, because `ssubscribe` is called, before `ready` is emitted.

PavelPashov (changes requested):
LGTM

PavelPashov (approved):

test/functional/spub_ssub.ts
rit3sh-x: @PavelPashov thanks, that makes sense, the poll would leak mocha's full timeout on any regression and adds dead latency in the happy path.

Would this approach work instead? Replace the poll with an event-driven signal: have the stub's `callsFake` capture the args and resolve a deferred promise the moment the 3rd call lands. Something like:

```ts
const calls: (string | Buffer)[][] = [];
let resolveAllCalls: () => void;
const allCalls = new Promise<void>((resolve) => {
  resolveAllCalls = resolve;
});

const stub = sinon
  .stub(Redis.prototype, "ssubscribe")
  .callsFake((...args: (string | Buffer)[]) => {
    calls.push(args);
    if (calls.length === 3) resolveAllCalls();
    return Promise.resolve(calls.length);
  });

subscriber.disconnect(true);
await allCalls;

expect(calls[0]).to.deep.equal(["shard1"]);
expect(calls[1]).to.deep.equal(["shard2"]);
expect(calls[2]).to.deep.equal(["shard3"]);
```

That way we get zero latency in the happy path, the `once("ready")` + second `ping()` become unnecessary since the stub directly observes the calls, and mocha's timeout still catches the failure case. Does that shape look right to you, or would you prefer something different?

rit3sh-x:
@PavelPashov  I’ve updated the implementation, could you please review it again?

PavelPashov:
> @PavelPashov I’ve updated the implementation, could you please review it again?

This removes the polling implementation detail, but it still appears to have the same failure mode: if the 3 ssubscribe calls don’t all happen, the test just hangs until timeout. What makes this fundamentally different from the previous version?

rit3sh-x:
You are right, the polling vs deferred-promise change was cosmetic. Both versions hang to the mocha timeout on failure.

Pushed a simpler fix: use `once("ready")` as the signal. Per `lib/redis/event_handler.ts#readyHandler`, the auto-resubscribe loop runs synchronously before `setStatus("ready")` emits (emit is deferred via `process.nextTick`), so by the time the ready listener fires, the stub has already captured all 3 calls. No poll, no deferred promise, and the assertion runs immediately, no extra latency.

Failure case: stub has <3 calls, assertions fail directly at the first mismatched expect, no wait needed. Resolving this thread.

PavelPashov (approved):
LGTM, thank you for the contribution

Issue #2073, opened 2026-02-06
Add support for custom ConnectionPool or Redis class in Cluster


rafaelclp:
Hi, I'd like to request a change that would allow us to support custom `Redis` subclasses when using `ioredis.Cluster`. I'd be happy to open a PR myself if/after we agree on some solution!

## Context

We are extending the ioredis.Redis class to add some logging, error handling and opentelemetry instrumentation:
```typescript
class RedisWithInstrumentation extends Redis { ... }
```

We'll start using `ioredis.Cluster`, and there is no easy way to implement the same instrumentation without monkeypatching.

## Idea

If we could tell `Redis.Cluster` to use a different `ConnectionPool` or `Redis` class, we could just pass in our own:
```typescript
// Option 1
const redisClient = new Redis.Cluster(nodes, {Redis: RedisWithInstrumentation, ...otherClusterOptions});
// Option 2
const redisClient = new Redis.Cluster(nodes, {ConnectionPool: ConnectionPoolWithInstrumentation, ...otherClusterOptions});
```

Option 1 would be the nicest: no extra work, just pass in the same `Redis` subclass we were already using. But option 2 might make more sense, and also solves the problem. Either solution would be great!

PavelPashov:
Hey @rafaelclp, Option 1 is likely the cleanest solution here. Cluster uses multiple Redis instances: clients for each cluster node (via the connection pool), a cluster subscriber client, and when sharded Pub/Sub is enabled sharded subscriber clients per node. Overriding the Redis constructor covers all internal clients, swapping ConnectionPool will not.

Also, what’s missing from the existing OpenTelemetry ioredis instrumentation? Are you already using [@opentelemetry/instrumentation-ioredis](https://www.npmjs.com/package/@opentelemetry/instrumentation-ioredis)? It instruments ioredis by wrapping the underlying `Redis.sendCommand`, so it captures commands in Cluster mode as well. What are you missing beyond that (e.g., logging/error handling)?

rafaelclp:
I definitely would prefer option 1, so it's great you think that's the cleanest! I suggested swapping `ConnectionPool` because the `Redis` objects are created by it, instead of being created directly by `Cluster` - so looking at the implementation, passing in the `ConnectionPool` looks cleaner. But looking from the outside (as a user of the package), indeed passing in `Redis` makes much more sense (I shouldn't have to care about who creates the objects).

[...]

PR #2098, opened 2026-04-10
fix: preserve subclasses in duplicate() (#2053)

### LINKED ISSUE #2053: duplicate method should new subclasses
Image

I think it should be `new this.constructor` instead of `new Redis` in the `duplicate` method; otherwise, subclasses that inherit from Redis will behave incorrectly.

So do Cluster.

rit3sh-x:
Closes #2053.

## Summary

`Redis#duplicate()` and `Cluster#duplicate()` now preserve the actual class of the instance, so subclasses are returned from `duplicate()` instead of plain `Redis` / `Cluster`.

## Root cause

The `applyMixin` utility in `lib/utils/applyMixin.ts` copied every own property from the mixin prototype onto the derived class prototype — including the built-in `constructor` property. When `applyMixin(Redis, EventEmitter)` runs at module load, `EventEmitter.prototype.constructor` overwrites `Redis.prototype.constructor`, so any Redis instance had `r.constructor === EventEmitter` instead of `Redis`.

That is why a naive `new this.constructor(...)` fix for `duplicate()` previously broke Cluster (as noted in Pavel's comment on #2053) — the constructor was wrong at the root.

## Fix

1. **`lib/utils/applyMixin.ts`** — skip the `constructor` key when copying descriptors, using the canonical `Object.getOwnPropertyDescriptors` + `delete` + `Object.defineProperties` pattern. Preserves getters/setters, non-enumerable props, and symbol keys.
2. **`lib/Redis.ts`** — `duplicate()` now uses `this.constructor` and returns `this`.
3. **`lib/cluster/index.ts`** — same treatment for `Cluster.duplicate()`, typed with `ClusterNode[]` / `ClusterOptions` for proper type safety.

## Example

```ts
class InstrumentedRedis extends Redis {
  getTag() { return "instrumented"; }
}

const redis = new InstrumentedRedis(6379);
const dup = redis.duplicate();
// Before: dup instanceof InstrumentedRedis === false (plain Redis)
// After:  dup instanceof InstrumentedRedis === true
// After:  dup.getTag() === "instrumented"
```

## Test plan

New unit test file `test/unit/duplicate.ts` with **15 tests** covering:

- `Redis.prototype.constructor` correctly resolves to `Redis` (applyMixin fix)
- `Cluster.prototype.constructor` correctly resolves to `Cluster`
- `redis.duplicate()` returns a `Redis` instance on plain Redis
- `redis.duplicate()` returns a subclass instance when called on a subclass
- Subclass methods are preserved across `duplicate()`
- Subclass instance fields (set in constructor) are preserved
- Override options are applied to the duplicate
- Deeply-nested subclass hierarchies (A → B → C → Redis) work
- Same coverage for `Cluster` subclasses
- `applyMixin` still mixes in EventEmitter methods (`on`, `emit`, `removeAllListeners`) onto both Redis and Cluster

## Verification

- [x] `npm run build` — PASS
- [x] `npm run test:tsd` — PASS
- [x] Full unit + functional suite: 609 passing, 26 pending, only pre-existing flakes unrelated to this change
- [x] Cluster integration tests (Docker): 17/17 passing
- [x] New duplicate tests: 15/15 passing

## Relation to #2073

Independent but complementary. #2073 proposes a `redisClass` option on Cluster to thread a custom class through all internal connections. This PR fixes the underlying inheritance bug that prevented `new this.constructor(...)` from working in the first place.

Either PR can land first — there are no file-level conflicts (different line ranges in `cluster/index.ts`).

---

> [!NOTE]
> **Low Risk**
> Targeted inheritance/mixin fix with broad unit tests; duplicate() is used internally (e.g. MONITOR, slot refresh) but behavior aligns with documented intent.
> 
> **Overview**
> Fixes **#2053** so `duplicate()` on custom `Redis` / `Cluster` subclasses returns the same class (instrumentation, extra methods, constructor fields) instead of a plain base instance.
> 
> **Root cause:** `applyMixin` copied `EventEmitter.prototype.constructor` onto `Redis`/`Cluster`, so `instance.constructor` was wrong and `duplicate()` could not safely use `new this.constructor(...)`.
> 
> **Changes:** `applyMixin` now skips copying `constructor` when merging prototypes. `Redis#duplicate()` and `Cluster#duplicate()` instantiate via `this.constructor` and are typed to return `this`. New unit tests cover constructors, nested subclasses, overrides, and that EventEmitter APIs still work.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 994ec8cc78a929bec25ac7d1197d08d9145ffbc6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PavelPashov:
Thanks for the PR. I will review it and follow up once I have gone through it.

rit3sh-x:
@PavelPashov any updates?

rit3sh-x:
@PavelPashov any updates?

PR #2105, opened 2026-04-20
fix: add SNI servername for TLS connection from host field

dougEfresh:
If  `servername` field is not explicitly set then use `host` field for SNI identification. 
This allows TLS terminating proxies to determine which cert to serve.

---

> [!NOTE]
> **Low Risk**
> Small, localized change to connection option construction plus targeted tests; behavior only changes when TLS is used without an explicit `servername`.
> 
> **Overview**
> `StandaloneConnector` now auto-populates TLS `servername` from the configured `host` when TLS is enabled and `servername` isn’t explicitly provided, while *skipping IP hosts* to avoid invalid SNI.
> 
> Unit tests were extended to cover the new SNI defaulting behavior for hostname vs IP `host` values.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 61db0c1a89936d35720c2d427c2e05fed1f696b4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/connectors/StandaloneConnector.ts
+++ b/lib/connectors/StandaloneConnector.ts
@@ -1,4 +1,4 @@
-import { createConnection, IpcNetConnectOpts, TcpNetConnectOpts } from "net";
+import { createConnection, IpcNetConnectOpts, TcpNetConnectOpts, isIP } from "net";
 import { connect as createTLSConnection, ConnectionOptions } from "tls";
 import { NetStream } from "../types";
 import { CONNECTION_CLOSED_ERROR_MSG } from "../utils";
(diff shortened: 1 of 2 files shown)


PavelPashov:
@dougEfresh Thanks for the contribution. Could you also add a test that reproduces this issue and verifies that the solution works as expected?

dougEfresh:
@PavelPashov I added a couple of tests

PavelPashov (approved):
Approving this as-is. As a small follow-up, it may be worth applying the same `servername` fallback in the Sentinel connector as well, so TLS behavior stays consistent across connection modes.

lib/connectors/StandaloneConnector.ts
```diff
@@ -41,6 +41,12 @@ export default class StandaloneConnector extends AbstractConnector {
 
     if (options.tls) {
       Object.assign(connectionOptions, options.tls);
+      if ("host" in connectionOptions && !("servername" in connectionOptions)) {
+        const host = (connectionOptions as TcpOptions).host;
+        if (host && !isIP(host)) {
+          (connectionOptions as any).servername = host;
+        }
+      }
     }
```
matheusvellone, line 50: I think adding the `tls.servername` even when `tls` is _undefined_ would be a good improvement. 

In my case, where I'm using only the `tls.servername`, I would need to set an empty `tls: {}` to trigger this `servername` logic, which would be kinda weird.

I believe setting the `servername` for non-TLS connections isn't a breaking change, right?

PR #2126, opened 2026-06-22
feat: add read/write splitting for standalone Redis instances

slukes:
## Why

Enables read/write splitting for non-cluster Redis setups — primarily for ElastiCache read replicas but applicable to any topology with a primary + read replicas. Today, `scaleReads` is cluster-only; standalone users have no equivalent.

## How

Extracts the routing logic from cluster mode into a shared `ReadWriteRouter` utility, then wires it into the standalone `Redis` class:

- **`lib/utils/ReadWriteRouter.ts`** — new shared class with the same read/write routing logic that cluster already uses (`isReadOnly`, round-robin, `slave`/`all`/`master` strategies, function-based routing)
- **`lib/Redis.ts`** — initializes a `ReadWriteRouter` in the constructor; routes commands in `sendCommand` before the existing offline-queue logic; disconnects read instances when the primary disconnects
- **`lib/redis/RedisOptions.ts`** — adds `scaleReads` option (array of endpoints, string strategy, or function) to `CommonRedisOptions`

Array format creates lazy-connecting read `Redis` instances; string/function format delegates to the router directly — same semantics as cluster.

## Changes

- `lib/utils/ReadWriteRouter.ts` — new shared routing utility (extracted from cluster)
- `lib/Redis.ts` — `initializeReadWriteRouter()`, routing in `sendCommand`, cleanup in `disconnect()`
- `lib/redis/RedisOptions.ts` — `scaleReads` option on `CommonRedisOptions`
- `examples/read_write_splitting.js` — usage examples (ElastiCache + local dev)
- `test/functional/read_write_splitting.ts` — functional tests (routing, round-robin, fallback, cleanup)
- `test/unit/read_write_command_classification.ts` — unit tests for read/write command detection

## Evidence of Testing

Functional tests cover:
- Write commands routed to primary, read commands to replicas
- Round-robin distribution across multiple replicas
- Graceful fallback to primary when read replica is unreachable
- Read instance cleanup on `disconnect()`
- Backward compatibility (no `scaleReads` → unchanged behavior)

## Review Focus

- The `ReadWriteRouter` extraction: does the abstraction feel right, or should the logic stay inline?
- The `sendCommand` insertion point (before `blockingTimeout`) — open to moving it
- Whether `scaleReads: 'slave'` is the right alias for standalone (vs. something more neutral like `'replica'`)

---
_This pull request was created with AI assistance._

---

> [!NOTE]
> **Medium Risk**
> Changes core `sendCommand` routing and opens extra connections; replica lag can surface stale reads, though writes and non-readonly commands still hit the primary.
> 
> **Overview**
> **Standalone Redis** now supports **read/write splitting** through the same **`scaleReads`** option shape as cluster mode, aimed at primary + replica setups (e.g. ElastiCache).
> 
> A new **`ReadWriteRouter`** centralizes routing: non–read-only commands stay on the primary; read-only commands (via `@ioredis/commands` flags or `command.isReadOnly`) go to **round-robin** replicas when `scaleReads` is an endpoint array or `'all'` / `'slave'`, with optional **custom function** selection. Replicas that are not **`ready`** are skipped so reads **fall back to the primary** instead of hanging on dead connections.
> 
> **`Redis`** wires this in **`initializeReadWriteRouter()`** (array `scaleReads` spawns child **`Redis`** connections with `readOnly: true` and eager connect), **`sendCommand`** when status is **`ready`**, and **`disconnect`** tears down replica clients. **`CommonRedisOptions`** documents the new `scaleReads` types. Adds **`examples/read_write_splitting.js`** plus functional and unit tests for routing, round-robin, fallback, and command classification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 00967009766eed790a5849f3eb0930a07354fa9c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/read_write_splitting.js
+++ b/examples/read_write_splitting.js
@@ -1,0 +1,85 @@
+const Redis = require("../built/index.js").default;
+
+// Example 1: Using read/write splitting with ElastiCache read replicas (array format)
+const redis = new Redis({
+  // Primary endpoint (for writes)
+  host: 'primary.cache.amazonaws.com',
+  port: 6379,
+  
+  // Read replicas (for reads) - same API as cluster!
+  scaleReads: [
+    { host: 'replica1.cache.amazonaws.com', port: 6379 },
+    { host: 'replica2.cache.amazonaws.com', port: 6379 }
+  ]
+});
+
+// Example 2: Using cluster-style scaleReads (when you manage instances separately)
+const redisWithClusterStyle = new Redis({
+  host: 'primary.cache.amazonaws.com',
+  port: 6379,
+  scaleReads: 'all' // Same as cluster - routes reads to available read instances
+});
+
+async function example() {
+  try {
+    // Write operations go to primary
+    await redis.set('user:1', JSON.stringify({ name: 'John', age: 30 }));
+    await redis.hset('counters', 'views', 100);
+    
+    // Read operations are distributed across read replicas
+    const user = await redis.get('user:1');
+    const views = await redis.hget('counters', 'views');
+    const keys = await redis.keys('user:*');
+    
+    console.log('User:', JSON.parse(user));
+    console.log('Views:', views);
+    console.log('User keys:', keys);
+    
+    // Mixed operations
+    await redis.incr('counters:visits'); // Write - goes to primary
+    const visits = await redis.get('counters:visits'); // Read - goes to replica
+    
+    console.log('Visits:', visits);
+    
+  } catch (error) {
+    console.error('Error:', error);
+  } finally {
+    redis.disconnect();
+  }
+}
+
+// For demonstration with localhost (when you don't have ElastiCache)
+async function localExample() {
+  const localRedis = new Redis({
+    host: '127.0.0.1',
+    port: 6379,
+    scaleReads: [
+      { host: '127.0.0.1', port: 6379 } // Same instance for demo
+    ]
+  });
+  
+  try {
+    await localRedis.set('demo', 'value');
+    const result = await localRedis.get('demo'); // This will use read endpoint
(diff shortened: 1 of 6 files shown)


PavelPashov:
@slukes Thanks for the PR. I’ll take a look and review it. In the meantime, please check why the new tests are failing.

slukes:
Rebased onto latest `main` and pushed a fix-up covering the CI failures from the previous run:

- Removed 5 incorrect test assertions (`INFO`/`PING`/`ECHO`/`TIME`/`LASTSAVE` aren't flagged `readonly` by Redis and were never meant to be routed to replicas)
- Fixed tests that accessed a non-existent `readInstances` property instead of `readWriteRouter.getReadInstances()`
- Implemented the actual fallback-to-primary behavior: read replicas now connect eagerly, and routing only targets a replica once its status is `ready`, otherwise it falls back to the primary. Previously a downed replica would just queue commands against a dead connection with no fallback.
- Replaced random `sample()` selection with genuine round-robin (the `roundRobinIndex` field existed but was unused)
- Fixed a race in the "route commands" test that could double-count `sendCommand` invocations when a command is queued while the connection is still opening

Converting back to draft until the new CI run is green.

slukes:
Rebased onto latest `main` (past the RESP3 change in #2127) — one small conflict in the constructor overloads, resolved by keeping the new RESP3-typed overloads and re-adding the `readWriteRouter` field declaration alongside them. Still a single commit, 629 insertions, verified against a local Redis instance (112/112 tests passing across repeated runs, build/lint/tsd clean).

PR #2137, opened 2026-07-09
fix(cluster): support zero-argument sunsubscribe with sharded subscribers

### LINKED ISSUE #2133: Cluster#sunsubscribe() without channels fails when `shardedSubscribers` is enabled
# Description

Calling `sunsubscribe()` with no channel arguments appears to fail in Redis Cluster mode when `shardedSubscribers: true` is enabled.

Redis supports `SUNSUBSCRIBE` with no shard-channel arguments (https://redis.io/docs/latest/commands/sunsubscribe/). In that form, the command should unsubscribe the client from all previously subscribed shard channels.

This seems related to #1759, but the issue here is narrower: channel-specific `SUNSUBSCRIBE <channel>` can be routed by slot, while zero-argument `SUNSUBSCRIBE` has no channel/key from which ioredis can compute a target slot.

# Reproduction

You'll need to install both `ioredis` (version 5.11.1) and testcontainers (any version). I used testcontainers so that I don't need to launch a separate redis cluster process, but it should have no effect on the bug

```js
const { GenericContainer, Wait } = require('testcontainers');

const REDIS_CLUSTER_IMAGE =
  process.env.REDIS_CLUSTER_IMAGE ??
  'harbor.jscrambler.com/proxy_cache/grokzen/redis-cluster:7.0.7';
const CLUSTER_PORTS = [7000, 7001, 7002, 7003, 7004, 7005];

async function main() {
  let container;
  let redis;

  try {
    container = await startRedisClusterContainer();
    redis = new Redis.Cluster(container.nodes, {
      natMap: container.natMap,
      shardedSubscribers: true,
    });

    await new Promise((resolve) => {
      redis.once('ready', resolve);
    });

    const channel = `ioredis-repro:${Date.now()}`;

    console.log('SSUBSCRIBE');
    await redis.ssubscribe(channel);

    console.log('SUNSUBSCRIBE WITH CHANNEL ARGS');
    await redis.sunsubscribe(channel);

    console.log('SUNSUBSCRIBE WITH NO CHANNEL ARGS');
    await redis.sunsubscribe();
  } finally {
    redis?.disconnect();
    await container?.redisContainer.stop({ remove: true, removeVolumes: true });
  }
}

async function startRedisClusterContainer() {
  const redisContainer = await new GenericContainer(REDIS_CLUSTER_IMAGE)
    .withExposedPorts(...CLUSTER_PORTS)
    .withStartupTimeout(120000)
    .withWaitStrategy(Wait.forListeningPorts())
    .start();

  await waitForRedisClusterReady(redisContainer);

  const host = redisContainer.getHost();
  const [networkName] = redisContainer.getNetworkNames();
  const containerIp = redisContainer.getIpAddress(networkName);
  const nodes = CLUSTER_PORTS.map((port) => ({
    host,
    port: redisContainer.getMappedPort(port),
  }));
  const natMap = Object.fromEntries(
    CLUSTER_PORTS.map((port) => [
      `${containerIp}:${port}`,
      { host, port: redisContainer.getMappedPort(port) },
    ]),
  );

  return { redisContainer, nodes, natMap };
}

async function waitForRedisClusterReady(redisContainer) {
  const startedAt = Date.now();

  while (Date.now() - startedAt < 120000) {
    const result = await redisContainer.exec([
      'redis-cli',
      '-p',
      CLUSTER_PORTS[0].toString(),
      'cluster',
      'info',
    ]);

    if (result.exitCode === 0 && result.output.includes('cluster_state:ok')) {
      return;
    }

    await new Promise((resolve) => {
      setTimeout(resolve, 200);
    });
  }

  throw new Error('Timeout waiting for Redis Cluster to become ready');
}

main().catch((error) => {
  console.error(error);
  process.exitCode = 1;
});

```

# Expected behavior

`redis.sunsubscribe()` with no arguments should unsubscribe from all currently subscribed shard channels, matching Redis’ `SUNSUBSCRIBE [shardchannel ...]` semantics.

### Actual behavior

`redis.sunsubscribe()` with no arguments fails in cluster mode with `shardedSubscribers: true`.

Please replace this section with the exact error output from the repro, for example:

```text
SSUBSCRIBE
SUNSUBSCRIBE WITH CHANNEL ARGS
SUNSUBSCRIBE WITH NO CHANNEL ARGS
TypeError: Cannot read properties of undefined (reading '0')
    at ClusterSubscriberGroup.getResponsibleSubscriber (<redacted>/node_modules/.pnpm/ioredis@5.9.3/node_modules/ioredis/built/cluster/ClusterSubscriberGroup.js:70:48)
    at tryConnection (<redacted>/node_modules/.pnpm/ioredis@5.9.3/node_modules/ioredis/built/cluster/index.js:439:62)
    at EventEmitter.sendCommand (<redacted>/.pnpm/ioredis@5.9.3/node_modules/ioredis/built/cluster/index.js:424:9)
    at EventEmitter.sunsubscribe (<redacted>/node_modules/.pnpm/ioredis@5.9.3/node_modules/ioredis/built/utils/Commander.js:90:25)
    at main (<redacted>/scripts/repro-ioredis-sharded-sunsubscribe.js:33:17)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

### Notes

This appears to be a cluster routing edge case. For `SSUBSCRIBE` and channel-specific `SUNSUBSCRIBE <channel>`, ioredis can compute a slot from the channel name and route the command to the relevant sharded subscriber connection. For zero-argument `SUNSUBSCRIBE`, there are no keys/channels, so there is no slot to compute.

### Environment

* ioredis version: 5.11.1
* Node.js version: 22.21.1
* Redis version/image: `grokzen/redis-cluster:7.0.7` but any should do
* Mode: Redis Cluster
* Options: `shardedSubscribers: true`

Develop-KIM:
Fixes #2133.

## Problem

With `shardedSubscribers: true`, calling `cluster.sunsubscribe()` with no arguments throws:

```
TypeError: Cannot read properties of undefined (reading '0')
    at ClusterSubscriberGroup.getResponsibleSubscriber
```

A zero-argument `SUNSUBSCRIBE` has no channel to hash, so `command.getSlot()` is `null`. The routing still calls `getResponsibleSubscriber(null)`, which indexes `clusterSlots[null]` and blows up.

## Fix

Redis treats a zero-argument `SUNSUBSCRIBE` as "unsubscribe this connection from all of its shard channels". With sharded subscribers those channels can live on multiple shards, so there is no single slot to route to. This special-cases the zero-argument form and fans a zero-argument `SUNSUBSCRIBE` out to every sharded subscriber that currently holds channels, clears the tracked channels, and resolves with `0`.

I also pulled the "does this subscriber still hold channels" check out of `shouldStartSubscriber` into a small `hasSubscribedChannels` helper so both paths share it.

## Verification

- Added a functional test in `test/functional/cluster/spub_ssub.ts` (MockServer) asserting the zero-argument `sunsubscribe()` resolves and reaches both shard connections.
- Checked end-to-end against a real 3-master cluster: subscribed one channel per shard, then called `sunsubscribe()`. Before this change it threw and messages kept arriving on all shards; after, it resolves with `0` and no further messages arrive.
- `npm run lint` and `npm run build` are clean.

I used Claude Code (Opus) to help write this and have reviewed it myself. One design call worth your input: it currently surfaces whatever `Promise.all` throws if a shard fails mid-unsubscribe — happy to switch to best-effort (`allSettled`) if you would prefer.

---

> [!NOTE]
> **Medium Risk**
> Touches sharded Pub/Sub command routing and subscription bookkeeping; incorrect fan-out or channel clearing could leave ghost subscriptions or failed unsubscribes, though behavior is covered by new tests.
> 
> **Overview**
> Fixes a crash when **`cluster.sunsubscribe()`** is called with no channel names while **`shardedSubscribers: true`**. Routing used a null slot and blew up in **`getResponsibleSubscriber`**.
> 
> **`Cluster.sendCommand`** now detects zero-key **`sunsubscribe`** and delegates to **`ClusterSubscriberGroup.sunsubscribeAll()`**, which fans out zero-argument **`SUNSUBSCRIBE`** to every sharded subscriber that still has tracked channels, awaits ready shards with **`allSettled`**, clears local channel state on full success, and rejects (keeping channels) if any ready shard fails—while still issuing **`SUNSUBSCRIBE`** on non-ready connections so reconnect does not silently resubscribe.
> 
> **`hasSubscribedChannels`** is extracted and shared with **`shouldStartSubscriber`** for lazy subscriber startup. Tests cover multi-shard unsubscribe, non-ready connections, and partial failure/retry behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e3bb774fda42e9bd373adede0e27cac0a819c5c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lib/cluster/ClusterSubscriberGroup.ts
```diff
@@ -121,6 +121,38 @@ export default class ClusterSubscriberGroup {
...
+      const redis = subscriber.getInstance();
+      if (redis && redis.status === "ready") {
+        pending.push(redis.sunsubscribe());
+      }
+    }
+
+    // Drop the tracked channels so a later reconnect doesn't resubscribe them.
+    this.channels.clear();
+
+    await Promise.all(pending);
```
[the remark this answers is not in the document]
Develop-KIM: Good catch — confirmed. On reconnect `autoResubscribe` replays `SSUBSCRIBE` from each connection's own subscription set, not the group's tracked channels, so skipping a non-ready shard left it subscribed server-side while we reported everything unsubscribed.

Fixed in 83e1085: `SUNSUBSCRIBE` now goes to every subscriber that still holds channels — ioredis queues it on a non-ready connection and clears that connection's shard subs when it flushes on reconnect. Only ready connections are awaited so a reconnecting shard can't hang the caller, and per-shard errors are swallowed (this also settles the `Promise.all` vs best-effort question from the PR description). Added a regression test that fails without the change.

lib/cluster/ClusterSubscriberGroup.ts
```diff
@@ -121,6 +121,54 @@ export default class ClusterSubscriberGroup {
...
+        pending.push(settled);
+      }
+    }
+
+    // Drop the tracked channels so the group no longer treats them as active.
+    this.channels.clear();
+
+    await Promise.all(pending);
+
+    return 0;
```
PavelPashov: I think this should reject when a shard's SUNSUBSCRIBE fails instead of swallowing and resolving `0`, the client may still be subscribed on that shard, and the app gets no signal. Since zero-arg SUNSUBSCRIBE is idempotent, a retry is safe: collect results with `Promise.allSettled`, reject with an error naming the failed nodes, and move  `channels.clear()` after the await so a retry still sees those shards in `hasSubscribedChannels`.

lib/cluster/ClusterSubscriberGroup.ts
```diff
@@ -121,6 +121,54 @@ export default class ClusterSubscriberGroup {
...
+        continue;
+      }
+
+      // Issue SUNSUBSCRIBE even when the connection isn't ready. ioredis queues
+      // it and, on reply, drops these channels from the connection's own
+      // subscription set — the set `autoResubscribe` replays on reconnect.
+      // Skipping a non-ready connection would leave it subscribed there, so it
+      // would silently resume delivering shard messages once it reconnects even
+      // though we reported everything as unsubscribed. Swallow per-connection
+      // errors so one failing shard doesn't reject the whole fan-out.
```
PavelPashov: Nit: the comments here are quite heavy. Could we trim to one or two lines.

PavelPashov (changes requested):

lib/cluster/ClusterSubscriberGroup.ts
Develop-KIM: Agreed. Switched to `allSettled`, reject with the failed nodes, and moved `channels.clear()` after the await so a retry still hits them.

One thing I kept: only awaiting the ready connections, otherwise a reconnecting or dead shard hangs the call. So a non-ready shard won't show up in the reject. Can await all if you'd prefer.

lib/cluster/ClusterSubscriberGroup.ts
Develop-KIM: Trimmed to two lines.

Develop-KIM:
Merged latest main to clear the conflict — my branch collided with the RESP2/3 test refactor and the tsconfig `lib` change. I kept the parametrized `should receive messages`/`should works`/`supports password` tests and re-added the three zero-arg SUNSUBSCRIBE tests; the whole `cluster:spub/ssub` suite is green locally (14 passing). The earlier review changes (allSettled + reject naming the failed shard, trimmed comments) are already in, so this should be ready for another look when you get a chance.

Develop-KIM:
@PavelPashov ping — everything from your review is in, and the branch has been conflict-free with green CI since I merged main on Jul 24. Happy to rebase or adjust anything if you'd rather it landed differently; otherwise it's ready whenever you have time for another look.

PavelPashov:
> @PavelPashov ping — everything from your review is in, and the branch has been conflict-free with green CI since I merged main on Jul 24. Happy to rebase or adjust anything if you'd rather it landed differently; otherwise it's ready whenever you have time for another look.

PR #2138, opened 2026-07-14
test(DataHandler): cover deeply nested reply causing decoder stack overflow

### LINKED ISSUE #2108: Deeply nested RESP aggregates can trigger uncaught RangeError and crash the process
I found a denial-of-service issue in `ioredis` caused by deeply nested RESP array replies.

## Summary:
  A malicious Redis server, compromised cluster node, or on-path attacker can send a deeply nested RESP aggregate reply that triggers unbounded recursion in the bundled `redis-parser`.
  This leads to `RangeError: Maximum call stack size exceeded` escaping through the response parsing path and crashing the consuming Node.js process.

 ## Affected code path:
  - `lib/DataHandler.ts` passes socket data directly into `parser.execute(data)`
  - `redis-parser/lib/parser.js` recursively descends through nested arrays via `parseArray -> parseArrayElements -> parseType -> parseArray`
  - there is no nesting-depth bound, and the runtime exception is not converted into a controlled parser error

## Observed behavior:
  - shallow nesting succeeds
  - sufficiently deep nesting throws `RangeError: Maximum call stack size exceeded`
  - this reproduces both in direct parser mode and through the public ioredis socket path

## Tested on:
  - repository: `redis/ioredis`
  - commit: `4760e0a19c194f29f4feb703003dcf046e4509cd`

## Reproduction summary:
  - depth 4: parses successfully
  - depth 1200: parses successfully
  - depth 10000: throws `RangeError: Maximum call stack size exceeded`

## Impact:
  This is a remote denial of service against applications using ioredis when the Redis response path is attacker-controlled.

## Suggested fix:
  - enforce a maximum aggregate nesting depth in `redis-parser`, or replace recursive aggregate parsing with an explicit stack
  - additionally, `ioredis` should defensively convert runtime exceptions from `parser.execute(data)` into a controlled connection/parser failure rather than allowing an uncaught exception
  to terminate the process

GiHoon1123:
Fixes #2108.

`redis-parser` (and its replacement, the RESP3 Decoder from #2127) recurses
once per level of RESP aggregate nesting with no depth limit, so a
sufficiently deeply nested reply throws a RangeError instead of reporting a
parser error.

#2127 already wraps the decoder call in try/catch and routes failures
through the existing returnFatalError -> recoverFromFatalError path, so
this no longer needs a code change. There was no test pinning that behavior
down for this specific case, so this adds one.

Verified by temporarily removing the try/catch and confirming the new test
fails with the expected RangeError.

---

> [!NOTE]
> **Low Risk**
> The change is narrowly scoped defensive error handling around inbound parsing; the diff here is test-only, with production behavior already aligned to the existing fatal-recovery path.
> 
> **Overview**
> This PR’s **visible diff** adds a **regression test** in `DataHandler` for issue #2108: feeding a reply nested ~50,000 `*1` arrays must **not** throw synchronously and must invoke **`recoverFromFatalError`** once with a **`RangeError`** (“Maximum call stack size exceeded”) and `{ offlineQueue: false }`.
> 
> The accompanying fix (described in the PR, not in the snippet above) wraps stream **`decoder.write`** in **`try/catch`** and forwards unexpected errors through the existing **`returnFatalError` → `recoverFromFatalError`** path so parser stack overflows recover instead of killing the Node process.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 831b5d85abfd68b7ead17c19e15c48a53b9995ef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/test/unit/DataHandler.ts
+++ b/test/unit/DataHandler.ts
@@ -266,6 +266,29 @@ describe("DataHandler", () => {
     expect(setup.redis.recoverFromFatalError.called).to.be.false;
     expect(setup.redis.commandQueue.length).to.equal(1);
   });
+
+  it("recovers instead of crashing on a reply that overflows the decoder's call stack | https://github.com/redis/ioredis/issues/2108", () => {
+    const setup = setupDataHandler();
+
+    // The decoder recurses once per level of RESP aggregate nesting, with no
+    // depth limit, so a sufficiently deeply nested reply (from a
+    // compromised/malicious server, or a corrupted stream) throws a
+    // synchronous RangeError instead of reporting a parser error.
+    let deeplyNested = ":1\r\n";
+    for (let i = 0; i < 50000; i++) {
+      deeplyNested = "*1\r\n" + deeplyNested;
+    }
+
+    expect(() => setup.write(deeplyNested)).to.not.throw();
+
+    expect(setup.redis.recoverFromFatalError.calledOnce).to.be.true;
+    const [commandError, err, options] =
+      setup.redis.recoverFromFatalError.firstCall.args;
+    expect(commandError).to.be.instanceOf(RangeError);
+    expect(err).to.equal(commandError);
+    expect(commandError.message).to.include("Maximum call stack size exceeded");
+    expect(options).to.deep.equal({ offlineQueue: false });
+  });
 });
 
 function setupDataHandler(parserOptions = { stringNumbers: false }) {


PavelPashov:
@GiHoon1123 thank you for this PR, I'll take a look, however please check #2127 as it might make sense to adapt the change for that PR.

GiHoon1123:
Thanks for the pointer! I took a look at #2127 — its `DataHandler.ts` already wraps `decoder.write()` in a try/catch that routes into `returnFatalError` the same way, so this exact crash looks like it's already handled there for the new RESP3 decoder path.

Since #2127 is still in progress, this PR still seemed worth having as a fix for the currently released behavior - `main` still calls `redis-parser` directly with no such guard. Happy to close this in favor of #2127 once it lands, or adjust the scope here if you'd prefer a different approach - whichever is more useful, just let me know.

PavelPashov (approved):
Thanks, looks good

PR #2140, opened 2026-07-15
test: run the standard functional suite against an external Redis Enterprise database

kiryazovi-redis:
## Motivation

The functional test suite assumes a passwordless Redis on `localhost:6379`. This change lets the same suite also run against an **external / managed Redis Enterprise** database - remote host, authentication, and optional TLS - **without changing the existing local behaviour**.

## What changed

- **Opt-in endpoint discovery.** When `RE_CLUSTER=true`, `test/helpers/global.ts` resolves the target database's host, port, username, password and TLS from `REDIS_ENDPOINTS_CONFIG_PATH` (the same endpoints-config format already consumed by the scenario tests) and applies it to `DEFAULT_REDIS_OPTIONS`, so `new Redis()` connects to the managed database. `RE_DB_NAME` selects a named database (defaults to `standalone`). The destructive `afterEach` cleanup (`CLIENT KILL` / `SCRIPT FLUSH`) is reduced to `FLUSHALL`, which the default user can run on a dedicated BDB. **When `RE_CLUSTER` is unset, behaviour is unchanged** (`localhost:6379`, no auth).
- **RE skips with rationale** for specs that cannot run against a managed Enterprise database:
  - the AR.* array preview commands and `INCREX` (OSS previews absent on RE),
  - the vector set commands (not available across all managed RE versions, e.g. `VRANGE` on RE 8.0.x),
  - the `duplicate()` default-port assertion (hardcoded `6379`),
  - the batch-before-ready transaction test (`CONFIG GET` inside `MULTI` is restricted for the default user).

## Validation

Ran the curated functional + command suite against managed Redis Enterprise: **green across RE 8.0.16, 8.0.22 and 100.0.20**. Locally: 187 passing, 63 pending, 0 failing. With `RE_CLUSTER` unset the suite still targets `localhost:6379` unchanged.

---
This change was prepared with AI assistance and reviewed before submission.

---

> [!NOTE]
> **Low Risk**
> Test-only wiring and conditional skips; no library or runtime behavior changes for end users.
> 
> **Overview**
> Adds an **opt-in** path (`RE_CLUSTER=true`) so the standard functional tests can target a **managed Redis Enterprise** database instead of passwordless `localhost:6379`, without changing default local runs.
> 
> New `test/helpers/re-config.ts` reads `REDIS_ENDPOINTS_CONFIG_PATH` (and optional `RE_DB_NAME`) to resolve host, port, and credentials. `test/helpers/global.ts` applies those to `DEFAULT_REDIS_OPTIONS` for `new Redis()` and uses **FLUSHALL-only** teardown instead of `CLIENT KILL` / `SCRIPT FLUSH`, which managed RE often blocks.
> 
> Several specs **skip on RE** when they assume OSS-only commands, hardcoded port `6379`, or `CONFIG GET` inside `MULTI`: array/`INCREX`/vector-set suites, `duplicate` default-port check, and one transaction `#exec` test.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5f834cd851a2fc86134078fd41e538bc3d97fc84. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/test/helpers/global.ts
+++ b/test/helpers/global.ts
@@ -1,8 +1,41 @@
 import * as sinon from "sinon";
 import Redis from "../../lib/Redis";
+import { DEFAULT_REDIS_OPTIONS } from "../../lib/redis/RedisOptions";
+import { isReCluster, loadREConnection } from "./re-config";
+
+// When RE_CLUSTER=true, point the default connection used by `new Redis()` at the
+// managed Redis Enterprise database resolved from REDIS_ENDPOINTS_CONFIG_PATH.
+// Tests that pass an explicit host/port (mock servers, deliberate failure ports)
+// are unaffected. When RE_CLUSTER is unset, behaviour is unchanged (localhost:6379).
+if (isReCluster()) {
+  const re = loadREConnection();
+  DEFAULT_REDIS_OPTIONS.host = re.host;
+  DEFAULT_REDIS_OPTIONS.port = re.port;
+  if (re.username) {
+    DEFAULT_REDIS_OPTIONS.username = re.username;
+  }
+  if (re.password) {
+    DEFAULT_REDIS_OPTIONS.password = re.password;
+  }
+  if (re.tls) {
+    DEFAULT_REDIS_OPTIONS.tls = { rejectUnauthorized: false };
+  }
+}
 
 afterEach((done) => {
   sinon.restore();
+
+  if (isReCluster()) {
+    // A managed Redis Enterprise database does not permit CLIENT KILL / SCRIPT FLUSH
+    // for the default user, and killing connections would disrupt the shared proxy;
+    // FLUSHALL alone is enough to isolate tests on a dedicated BDB.
+    new Redis().flushall().then(
+      () => done(),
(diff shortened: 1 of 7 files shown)


test/helpers/global.ts
```diff
@@ -1,8 +1,41 @@
...
+  if (isReCluster()) {
+    // A managed Redis Enterprise database does not permit CLIENT KILL / SCRIPT FLUSH
+    // for the default user, and killing connections would disrupt the shared proxy;
+    // FLUSHALL alone is enough to isolate tests on a dedicated BDB.
+    new Redis().flushall().then(
+      () => done(),
+      (err) => done(err)
+    );
+    return;
+  }
```
PavelPashov, line 41: Could we disconnect the cleanup client in a `finally` block and ensure managed test clients are closed after each test? Without `CLIENT KILL`, these connections accumulate across the suite and may exhaust the database’s connection limit.

PR #2148, opened 2026-07-24
fix: clear stale socket timeout on reconnect

### LINKED ISSUE #2147: socketTimeout timer survives reconnect and destroys the replacement stream
## Description

When a command starts a `socketTimeout` timer on connection A and A closes before replying, the timer survives reconnection. If connection B becomes ready before A's original deadline, that stale callback destroys B through the current `this.stream` reference—even when B has already returned data successfully.

This makes B's timeout depend on A's deadline instead of B's own activity.

## Reproduction

The following uses the default `retryStrategy` (first retry after 50 ms):

```js
const net = require("node:net");

const socketTimeout = 250;
let connections = 0;
let closes = 0;
let droppedFirstPing = false;
const errors = [];
const sockets = new Set();

const server = net.createServer((socket) => {
  sockets.add(socket);
  const connection = ++connections;
  socket.on("close", () => sockets.delete(socket));
  socket.on("data", (data) => {
    if (!data.toString().toLowerCase().includes("ping")) return;

    if (connection === 1 && !droppedFirstPing) {
      droppedFirstPing = true;
      setTimeout(() => socket.destroy(), 20);
      return;
    }

    socket.write("+PONG\r\n");
  });
});

server.listen(6380, "127.0.0.1", async () => {
  const redis = new Redis({
    host: "127.0.0.1",
    port: 6380,
    lazyConnect: true,
    protocol: 2,
    disableClientInfo: true,
    enableReadyCheck: false,
    socketTimeout,
  });

  redis.on("error", (error) => errors.push(error.message));
  redis.on("close", () => closes++);

  await redis.connect();
  console.log(await redis.ping()); // PONG from connection B
  await new Promise((resolve) => setTimeout(resolve, socketTimeout + 100));

  console.log({ connections, closes, errors, status: redis.status });

  redis.disconnect();
  for (const socket of sockets) socket.destroy();
  server.close();
});
```

Observed on `v5.11.1` / Node.js `v24.15.0`:

```text
A closes:       ~27 ms
B becomes ready: 80 ms
B returns PONG:  80 ms
socketTimeout:  255 ms
C becomes ready: 308 ms
```

```js
{
  connections: 3,
  closes: 2,
  errors: [
    "Socket timeout. Expecting data, but didn't receive any in 250ms."
  ],
  status: "ready"
}
```

The same failure reproduced 3/3 with a fixed 10 ms retry delay. A healthy single-connection control remained ready with no timeout error.

## Expected behavior

Closing connection A should cancel A's socket timeout. Connection B should start and clear its own timeout based only on B's data activity. Since B returns `PONG`, it should remain connected after A's original deadline.

## Cause

`setSocketTimeout()` stores one timer on the Redis instance. Its callback destroys `this.stream`, which is evaluated when the callback fires. The close path does not clear that timer, while `sendCommand()` refuses to install B's timer because the stale timer reference is still defined.

DalecB:
This pull request resolves #2147.

The socket timeout timer was stored on the Redis instance and survived after its original stream closed. Because its callback accessed `this.stream` when it fired, it could destroy a replacement connection that had already returned data successfully.

The fix clears the timer when the connection closes and captures the owning stream when the timer is created.

A regression test covers the close → reconnect → successful reply → old deadline interleaving.

## Validation

- `test/functional/socketTimeout.ts`: 4 passing
- `npm run build`
- `npm run lint`

---

> [!NOTE]
> **Medium Risk**
> Touches connection lifecycle and stream teardown in a hot path used whenever socketTimeout is set; behavior change is narrowly scoped but affects all reconnect scenarios with that option.
> 
> **Overview**
> Fixes a reconnect bug when **`socketTimeout`** is enabled: the client-wide timer could outlive the socket that created it and later tear down a **new** stream after a successful reply.
> 
> **`setSocketTimeout`** now binds the timeout and its one-shot `data` listener to the **stream instance** that was active when the timer was armed, instead of reading **`this.stream`** when the timer fires. **`closeHandler`** clears **`socketTimeoutTimer`** as soon as the connection closes so deadlines from the old socket are not carried into the next connect.
> 
> A functional test simulates first-socket hang/destroy, reconnect, and waits past the old timeout to assert the replacement stays **`ready`** with no spurious socket-timeout errors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e4d8e7c96df58eb8b311570684009c4364b8a908. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/lib/Redis.ts
+++ b/lib/Redis.ts
@@ -685,8 +685,9 @@ class Redis<ReplyMapping extends ReplyMappingMode = "legacy">
   }
 
   private setSocketTimeout() {
+    const stream = this.stream;
     this.socketTimeoutTimer = setTimeout(() => {
-      this.stream.destroy(
+      stream.destroy(
         new Error(
           `Socket timeout. Expecting data, but didn't receive any in ${this.options.socketTimeout}ms.`
         )
@@ -696,7 +697,7 @@ class Redis<ReplyMapping extends ReplyMappingMode = "legacy">
 
     // this handler must run after the "data" handler in "DataHandler"
     // so that `this.commandQueue.length` will be updated
-    this.stream.once("data", () => {
+    stream.once("data", () => {
       clearTimeout(this.socketTimeoutTimer);
       this.socketTimeoutTimer = undefined;
       if (this.commandQueue.length === 0) return;
--- a/lib/redis/event_handler.ts
+++ b/lib/redis/event_handler.ts
@@ -343,6 +343,12 @@ export function closeHandler(self) {
     const prevStatus = self.status;
     self.setStatus("close");
 
+    // The timeout belongs to the stream that just closed and must not survive reconnect.
+    if (self.socketTimeoutTimer !== undefined) {
+      clearTimeout(self.socketTimeoutTimer);
+      self.socketTimeoutTimer = undefined;
+    }
+
     if (self.commandQueue.length) {
       abortIncompletePipelines(self.commandQueue);
     }
--- a/test/functional/socketTimeout.ts
+++ b/test/functional/socketTimeout.ts
@@ -1,6 +1,8 @@
 import { expect } from 'chai';
 import { Done } from 'mocha';
+import { Socket } from 'net';
 import Redis from '../../lib/Redis';
+import MockServer from '../helpers/mock_server';
 
 describe('Redis Connection Socket Timeout', () => {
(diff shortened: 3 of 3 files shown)


nkaradzhov (approved):
Thanks @DalecB CLGTM

PR #2150, opened 2026-07-27
fix: decode RESP3 doubles with Number

### LINKED ISSUE #2149: RESP3 double decoding diverges from Number(): max double becomes Infinity, sub-1e-17 values become 0
Reviewing the RESP3 decoder on main (5d0862e, not yet released), the `,` double path builds the
mantissa by accumulating `digit * 10^-k` from a 17-entry multiplier table and then applies the
exponent as a separate multiply (decoder.ts L422-463 and L481-500). That is not the correctly-rounded
conversion `Number(str)` does, and three cases come out wrong.

Driving the real `Decoder` class with single-value wire frames, decoder vs `Number()`:

  ,1.7976931348623157e308   ->  Infinity   (expected 1.7976931348623157e+308)
  ,5e-324                   ->  0          (expected 5e-324)
  ,0.000000000000000001     ->  0          (expected 1e-18)
  ,0.30000000000000004      ->  0.3000000000000001

The first is Number.MAX_VALUE: the accumulated mantissa lands a few ULP high and `* 10**308` tips it
past the max. The middle two are the `decimalIndex < MULTIPLIERS.length` guard silently dropping
every digit past the 17th, so a plain-notation value below 1e-17 loses all of its digits. The last is
ordinary accumulation drift at the 16th digit.

Splitting the same frames across two `write()` calls at every byte boundary gives the same results,
so it is the arithmetic rather than the streaming continuations. `inf`, `-inf`, `nan`, `(` bignums
and `:` integers all decode correctly.

Scores from ZSCORE / ZINCRBY / GEODIST / TS.* come through this path once RESP3 is on.

The straightforward fix is to collect the token to the CRLF and let `Number()` do the conversion,
keeping the existing integer fast path. I realise the multiplier table is there for speed, so I'd
want to benchmark before/after rather than hand you a regression — happy to send that plus a boundary
table test as a PR if you want it. If you'd rather keep the fast path and only fall back for tokens
with an exponent or more than 17 significant digits, that works too.

Same accumulation block exists in node-redis's packages/client/lib/RESP/decoder.ts, which is released
— worth a look there as well.

GiHoon1123:
Fixes #2149.

RESP3 double decoding currently builds the number manually by accumulating decimal digits and then applying the exponent. That does not always match JavaScript's `Number()` conversion, so a few boundary values decode incorrectly:

- `1.7976931348623157e308` becomes `Infinity`
- `5e-324` becomes `0`
- `0.000000000000000001` becomes `0`
- `0.30000000000000004` rounds to `0.3000000000000001`

This changes RESP3 double decoding to reuse the existing string decoding path and lets `Number()` perform the numeric conversion. RESP3 spellings for `inf`, `+inf`, `-inf`, and `nan` are still handled explicitly.

I also checked a small local decoder loop: short literals like `,1.5\r\n` were slightly slower, while the boundary cases above were faster with this change.

Tests added for the boundary values. The existing decoder test helper covers single chunk, byte-by-byte, and split-at-every-byte-boundary decoding for each case.

### DIFF
--- a/lib/resp/decoder.ts
+++ b/lib/resp/decoder.ts
@@ -34,11 +34,6 @@ const ASCII = {
   "+": 43,
   "-": 45,
   "0": 48,
-  ".": 46,
-  i: 105,
-  n: 110,
-  E: 69,
-  e: 101,
 } as const;
 
 export const PUSH_TYPE_MAPPING = {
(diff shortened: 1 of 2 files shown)


nkaradzhov (approved):
@GiHoon1123, thanks for the fix!  LGTM.

GiHoon1123:
Thanks for the review!

GiHoon1123:
Just checking in on this one. It looks like the PR has approval and the checks are green, but no rush at all. Please let me know if there’s anything else needed from my side.

PavelPashov:
> Just checking in on this one. It looks like the PR has approval and the checks are green, but no rush at all. Please let me know if there’s anything else needed from my side.

Merged now, thank you for the contribution

PR #2163, opened 2026-08-01
fix(types): allow number for zrange stop argument

### LINKED ISSUE #2162: ioredis 6.0.0 zrange argument type issue
In the previous ioredis v5.11.1, zrange accepted a stop argument of type number.
The current ioredis v6.0.0 doesn't allow it. I had to cast to string for it to pass type checking.
Redis seems to allow numeric values for both start and stop params, so is it just a typing error?

lazerg:
v5 accepted a numeric `stop` in `zrange`, but v6 dropped `number` from its type, so a call like `redis.zrange(key, 0, -1)` no longer typechecks and has to be cast to a string. The type map behind the commander generator adds `number` to bounds named `start`, but not `stop`, even though Redis takes numeric values for both. Adding `stop` to that list brings the numeric bound back and puts `zrange` in line with sibling commands like `zrangebyscore` and `zrangestore`, which already type both bounds as `number | string`.

Fixes #2162

---

> [!NOTE]
> **Low Risk**
> Type-only widening for generated Redis command signatures; no runtime behavior change.
> 
> **Overview**
> Restores numeric typing for the **`stop`** bound on **`zrange`**, so calls like `redis.zrange(key, 0, -1)` typecheck again without string casts.
> 
> The commander generator’s **`bin/typeMaps.js`** already widens bound-like argument names (e.g. **`start`**) to `string | Buffer | number`; **`stop`** is added to that name list, which regenerates **`RedisCommander.ts`** so every **`zrange`** / **`zrangeBuffer`** overload accepts **`number`** for **`stop`** as well as **`start`**. **`test/typing/commands.test-d.ts`** adds **`expectType`** coverage for numeric and string bounds.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26cd2c135e3c513fb27dbb657b37a210bf31362a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PavelPashov (approved):
Thank you for your contribution! Looks good.

PR #2172, opened 2026-08-28
fix(cluster): vendor cluster-key-slot to fix empty hash tag slot bug

mishan:
Redis treats an empty hash tag ("{}") as no tag at all: it hashes the whole key and stops searching. cluster-key-slot's single-pass scanner instead falls through and keeps scanning for the next "}", hashing whatever sits between - silently misrouting commands and, in Pipeline's same-slot gate, both wrongly rejecting same-slot pipelines and wrongly accepting cross-slot ones. Measured against a live Redis 8.0.6 cluster, this diverges from real slot assignment 460 times across an 8016-key corpus.

Upstream (invertase/cluster-key-slot) was archived in March 2026 and had been unmaintained for years before that, and ioredis pins the exact version "1.1.1" rather than a range, so no upstream fix would reach users regardless. The function is ~40 lines plus a lookup table with no dependencies, well within vendoring range.

Vendor the implementation into lib/utils/calculateSlot.ts with the fix applied: latch the whole-key fallback once an empty tag is seen, so a later "}" elsewhere in the key can no longer be mistaken for a tag close. Repoint Command, Pipeline, autoPipelining, and ClusterSubscriberGroup at the vendored module, drop the npm dependency, and add unit tests covering the empty-tag cases plus parity with upstream's own test suite.

I posted a full write-up of the bug here https://github.com/mishan/intraslot/blob/main/contrib/empty-hashtag-slot-bug/README.md

---

> [!NOTE]
> **Medium Risk**
> Changes core cluster slot hashing used for every keyed command, pipeline, and autopipeline routing; behavior shifts only for keys with empty `{}` tags, with broad test coverage mitigating regressions.
> 
> **Overview**
> Replaces the **`cluster-key-slot`** dependency with an in-tree **`lib/utils/calculateSlot`** module (vendored from 1.1.1) and wires **Command**, **Pipeline**, **autoPipelining**, and **ClusterSubscriberGroup** to import it locally. The npm dependency is removed from **`package.json`**.
> 
> The vendored hasher fixes **empty hash tags (`{}`)**: Redis hashes the **entire key** and stops tag scanning, while the upstream library could keep scanning and assign the wrong slot—breaking cluster routing and pipeline same-slot checks. The fix sets a **`done`** latch after an empty tag so later `{`/`}` pairs are not treated as tags.
> 
> New **`test/unit/calculateSlot.ts`** covers empty-tag cases (values checked against Redis 8), **`generateMulti`**, and parity with upstream’s hash tests. Cluster functional tests and the benchmark fixture now import the local module.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2821ab403a81cb02ea4fe768714dc42b2839ac70. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PavelPashov (approved):

PR #2176, opened 2026-09-02
feat(cluster): allow configuring node role for Pub/Sub subscriber

ebubekiryigit:
### Problem

Classic Pub/Sub in Redis Cluster uses one dedicated subscriber connection. ioredis currently selects the node for that connection from all discovered cluster nodes:

```ts
sample(this.connectionPool.getNodes());
```

This means the subscriber may connect to either a master or a replica.

The need for this option surfaced in a dedicated-server deployment where one replica had less stable networking than the master nodes. The cluster was already configured with:

```ts
scaleReads: "master"
```

It was reasonable to assume that this would also keep the classic Pub/Sub subscriber on a master. However, `scaleReads` only controls ordinary read-command routing; it does not participate in subscriber node selection. As a result, the long-lived subscriber connection could still be assigned to the less reliable replica.

This distinction is not obvious from the existing API and can affect other deployments where replicas have different network paths, firewall rules, maintenance policies, or availability characteristics.

### Solution

This pull request introduces a new Cluster option:

```ts
const cluster = new Redis.Cluster(nodes, {
  subscriberNodeRole: "master",
});
```

The supported values are:

- `"all"`: select from all discovered nodes.
- `"master"`: select only from master nodes.
- `"slave"`: select only from replica nodes.

The option is applied whenever ioredis needs to select a new classic Pub/Sub subscriber connection.

It remains independent from `scaleReads`:

```ts
const cluster = new Redis.Cluster(nodes, {
  // Routes ordinary read-only commands such as GET and MGET.
  scaleReads: "master",

  // Selects a master for the dedicated SUBSCRIBE/PSUBSCRIBE connection.
  subscriberNodeRole: "master",
});
```

`subscriberNodeRole` does not affect `PUBLISH` routing or Redis 7 sharded Pub/Sub subscribers.

### Connection stability

A role change alone does not replace an otherwise live subscriber connection. The existing connection remains in use until the normal subscriber lifecycle requires a new selection, such as when its node leaves the connection pool or the connection ends.

This avoids unnecessary subscriber replacement, which could otherwise introduce duplicate delivery or message gaps during subscription restoration.

When no node currently matches the configured role:

- the cluster remains ready;
- ordinary commands continue working;
- ioredis does not silently fall back to an ineligible node;
- subscription commands reject with an actionable `AbortError`;
- a later eligible node can be selected through the existing topology discovery flow.

### Backward compatibility

The default is:

```ts
subscriberNodeRole: "all"
```

This preserves the existing node-selection behavior for users who do not configure the option. There is no intentional behavior change unless `subscriberNodeRole` is explicitly set.

### Related work

https://github.com/redis/ioredis/pull/2080 addresses subscriber recovery and topology-related behavior. This pull request is complementary: it adds an explicit policy for choosing which node role may host the classic Pub/Sub subscriber connection.

---

> [!NOTE]
> **Medium Risk**
> Changes classic cluster Pub/Sub connection timing and subscribe error paths; default `"all"` preserves prior behavior, but explicit role settings can block subscriptions until topology matches.
> 
> **Overview**
> Adds **`subscriberNodeRole`** (`"all"` | `"master"` | `"slave"`, default **`"all"`**) so cluster clients can pin the dedicated classic **`SUBSCRIBE`/`PSUBSCRIBE`** connection to masters or replicas. Selection uses `connectionPool.getNodes(role)`; it is separate from **`scaleReads`** and does not affect **`PUBLISH`** or sharded Pub/Sub.
> 
> **Connect and topology behavior:** For restricted roles, subscriber startup waits until the first slots refresh (and reconnect paths start the subscriber after refresh) so role metadata is reliable; **`selectSubscriberIfNeeded()`** runs after **`CLUSTER SLOTS`** when nodes are reclassified without **`+node`**. Partial **`+node`** events no longer pick a restricted-role subscriber early.
> 
> **Failure mode:** If no eligible node exists, the cluster stays ready for normal commands but subscribe commands **`AbortError`** with a clear message instead of falling back to the wrong role. An existing subscriber is not swapped solely because its node’s role changed in topology.
> 
> README, TypeScript options, and functional/unit tests cover selection, rejection, reconnect, and duplication of the option.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ca26f31ff1fe9f5acb9262f28c295a37668af1a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PavelPashov (approved):
Looks good, thank you for the contribution.

PR #2180, opened 2026-09-05
fix: align auto-pipeline callbacks for dynamic commands

### LINKED ISSUE #1931: AutoPipelining: results[i] is not iterable
## What

We are using https://github.com/animir/node-rate-limiter-flexible with ioredis to implement rate limiting across our web endpoints. Yesterday, we observed a series of errors that caused our server to raise an uncaughtException on a global level and fully crash.

Within our endpoint, we got the first error:
```
TypeError: pipeline[functionName] is not a function
    at /app/node_modules/.pnpm/ioredis@5.4.1/node_modules/ioredis/built/autoPipelining.js:155:31
    at new Promise (<anonymous>)
    at executeWithAutoPipelining (/app/node_modules/.pnpm/ioredis@5.4.1/node_modules/ioredis/built/autoPipelining.js:144:33)
    at EventEmitter.rlflxIncr (/app/node_modules/.pnpm/ioredis@5.4.1/node_modules/ioredis/built/utils/Commander.js:114:63)
    at RateLimiterRedis._upsert (/app/node_modules/.pnpm/rate-limiter-flexible@5.0.3/node_modules/rate-limiter-flexible/lib/RateLimiterRedis.js:108:28)
    at /app/node_modules/.pnpm/rate-limiter-flexible@5.0.3/node_modules/rate-limiter-flexible/lib/RateLimiterStoreAbstract.js:203:12
    at new Promise (<anonymous>)
    at RateLimiterRedis.consume (/app/node_modules/.pnpm/rate-limiter-flexible@5.0.3/node_modules/rate-limiter-flexible/lib/RateLimiterStoreAbstract.js:195:12)
    at d.checkRateLimit (/app/web/.next/server/chunks/6702.js:1:2410)
    at d.rateLimitRequest (/app/web/.next/server/chunks/6702.js:1:2040)
```

I noticed that rlflxlncr is a function that is defined within the node-rate-limited-flexible library: https://github.com/animir/node-rate-limiter-flexible/blob/master/lib/RateLimiterRedis.js#L35, but cannot tell if there is a correlation.
```typescript
this.client.defineCommand("rlflxIncr", {
  numberOfKeys: 1,
  lua: this._incrTtlLuaScript,
});
```

About .7ms later, I see two exceptions being raised. Both with the following stacktrace:
```
TypeError: results[i] is not iterable (cannot read property undefined)
    at /app/node_modules/.pnpm/ioredis@5.4.1/node_modules/ioredis/built/autoPipelining.js:64:25
    at tryCatcher (/app/node_modules/.pnpm/standard-as-callback@2.1.0/node_modules/standard-as-callback/built/utils.js:12:23)
    at /app/node_modules/.pnpm/standard-as-callback@2.1.0/node_modules/standard-as-callback/built/index.js:22:53
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```
followed by `Signal:  uncaughtException`.

## Context / Versions

ioredis@5.4.1
rate-limiter-flexible@5.0.3
redis v7 (Managed ElastiCache Redis)

We ran this setup for about 2 months without errors, i.e. I expect this to be an edge case. It happened about 20s after the server started, i.e. it may be an issue on startup, not in ongoing operation.

We create our Redis client like this:
```typescript
const defaultRedisOptions: Partial<RedisOptions> = {
  maxRetriesPerRequest: null,
  enableAutoPipelining: env.REDIS_ENABLE_AUTO_PIPELINING === "true",
};

export const redisQueueRetryOptions: Partial<RedisOptions> = {
  retryStrategy: (times: number) => {
    // Retries forever. Waits at least 1s and at most 20s between retries.
    logger.warn(`Connection to redis lost. Retry attempt: ${times}`);
    return Math.max(Math.min(Math.exp(times), 20000), 1000);
  },
  reconnectOnError: (err) => {
    // Reconnects on READONLY errors and auto-retries the command.
    logger.warn(`Redis connection error: ${err.message}`);
    return err.message.includes("READONLY") ? 2 : false;
  },
};

export const createNewRedisInstance = (
  additionalOptions: Partial<RedisOptions> = {},
) => {
  const instance = env.REDIS_CONNECTION_STRING
    ? new Redis(env.REDIS_CONNECTION_STRING, {
        ...defaultRedisOptions,
        ...additionalOptions,
      })
    : env.REDIS_HOST
      ? new Redis({
          host: String(env.REDIS_HOST),
          port: Number(env.REDIS_PORT),
          password: String(env.REDIS_AUTH),
          ...defaultRedisOptions,
          ...additionalOptions,
        })
      : null;

  instance?.on("error", (error) => {
    logger.error("Redis error", error);
  });

  return instance;
};
```

## Assumptions

I assume that we fail to send the `rlflxIncr` command to Redis, but ioredis somehow still registers a callback for it. When it comes to processing the result, we try to access the result set at a position that is not filled, because redis never received this command. Then it fails out of the core request loop and raises an uncaught exception.

## What I expect to happen

No errors being raised or errors being thrown within the core application loop.

boyam01:
## Summary

Fixes #1931.

An auto-pipeline can be created by an earlier command in the same event-loop turn. If `defineCommand()` is called afterwards (for example, by a rate-limiter integration), the client has the new command but the existing pipeline does not. Calling it rejects with `pipeline[functionName] is not a function`. The callback was already appended, so processing the remaining batch can also throw `results[i] is not iterable` or misalign replies.

## Changes

- Resolve custom and added builtin commands from the current client, executing them with the existing pipeline as their receiver. Ordinary builtin commands continue to use the pipeline method.
- Register the result callback only after the command has been queued successfully.
- Cover late registration on standalone and Cluster clients, command redefinition, Buffer results, synchronous enqueue failure, and callback results after a real `NOSCRIPT` retry.

No public API, dependency, generated command declarations, or auto-pipeline exclusion rules are changed.

## Verification

Tested against base `101fa6265d0dcc9fb5e0204680c33d59f0209e94` using Node.js 24.19.0 and Redis 7.4.11 on Linux (WSL2):

- Reproduced the original missing-method and callback/result errors before the fix; the redefinition regression also returned the old script result before the fix.
- Full non-Cluster Mocha suite: **2,209 passing, 455 pending, 0 failures**. All six new regressions executed.
- Real six-node Redis Cluster suite: **62 passing, 4 pending, 0 failures**.
- Existing server-version/feature skips remain unchanged; pending tests are not counted as passed.
- TypeScript compilation, all tsd tests, the repository ESLint command, changed-file zero-warning lint, changed-file Prettier, and `git diff --check` passed. Repository-wide lint still reports 23 pre-existing warnings in three untouched files.
- Smoke-tested the compiled `built/` package, not just ts-node: standalone and real Cluster late Lua commands, Buffer and callback APIs, plus standalone redefinition.
- Independent read-only review found no blocking defect. CodeRabbit CLI 0.7.6 reviewed all four changed files and completed with **0 findings**.

Earlier Windows-to-WSL full-suite runs timed out in cleanup/connection hooks, and an unmodified-base run also timed out. The passing full-suite results above use native Linux Node and Redis together; no timeout, skip, or cleanup hook was changed to obtain them. This is one local Node/Redis combination, not a claim that the entire upstream CI matrix has run.

## AI assistance

This contribution was developed with OpenAI Codex assistance and reviewed with CodeRabbit. The root cause was reproduced against the real command path, and the changes were validated with the tests above.

---

> [!NOTE]
> **Medium Risk**
> Touches core auto-pipeline batching and callback/result pairing; behavior change is narrow (dynamic commands + enqueue ordering) but affects concurrent command scheduling.
> 
> **Overview**
> Fixes auto-pipelining when **custom or late-added commands** are used on a pipeline that was already created in the same tick (e.g. `get` then `defineCommand`).
> 
> **`executeWithAutoPipelining`** now resolves **Lua scripts** (`scriptsSet`) and **`addBuiltinCommand`** names from the **client**, while still enqueueing on the existing pipeline via `command.call(pipeline, …)`. Built-in commands keep using the pipeline method. The per-command **callback is registered only after enqueue succeeds**, so a synchronous enqueue failure cannot leave an extra callback and misalign replies.
> 
> New tests cover late custom commands (standalone and Cluster), Buffer variants, **redefinition** (queued calls keep their script version), **SCRIPT FLUSH** / NOSCRIPT retry with callbacks, sync enqueue failure, and late builtins.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9cc2e643396cd02af35f3ff75ad0a052c2c2bb93. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PavelPashov (approved):

PR #2187, opened 2026-09-10
fix: handle rejections from auto-resubscribe and select on reconnect

ramin-010:
SELECT is guarded when it runs during the handshake and unguarded when the same command runs from readyHandler.

sendHandshake runs each connect-path command as send().catch(errorHandler), and the handler registered for SELECT in getHandshakeCommands is (err) => self.silentEmit("error", err). But the SELECT, SUBSCRIBE, PSUBSCRIBE and SSUBSCRIBE calls that readyHandler issues to restore client state after a reconnect are fire-and-forget — seven call sites with no rejection handler.

So if one of these fails during resubscribe, for example because of NOPERM or the connection dropping again, it becomes an unhandled rejection. On Node 15+ that terminates the process by default, and it does so even when the user has an error listener registered on the client — the listener receives nothing.

I changed the reconnect path to handle these rejections with silentEmit("error", err), which is already what the handshake path does for the same commands.

I added tests for the reconnect cases: SUBSCRIBE, PSUBSCRIBE, and SELECT through both the subscriber branch and the final re-select. On main, each fails twice over — an unhandled rejection fires, and the error listener gets nothing. The remaining three call sites (SSUBSCRIBE, and the two SELECT calls in the resend and offline-queue loops) are structurally identical and aren't separately covered.

I also fixed the ssubscribe stub in spub_ssub.ts, since a bare sinon.stub() returns undefined instead of the promise the real method returns. The sibling stubs in the cluster tests already return a real promise via callsFake.

One consequence worth mentioning is that a failed SELECT on reconnect now surfaces as an error event rather than terminating the process, so a connection can carry on pointed at the wrong database. That isn't a new policy — SELECT during the handshake already routes to silentEmit, so this only makes the reconnect path match the handshake path.

I kept the changes limited to the reconnect path. ClusterSubscriberGroup already attaches a .catch() to its ssubscribe resubscribe but logs via debug() with a separate TODO about emitting an error event, so I left that unchanged. I went with silentEmit here because it's what this file already does for the equivalent SELECT, but I'm happy to switch to debug() if you'd rather the two paths match while that TODO is open.

---

> [!NOTE]
> **Medium Risk**
> Changes reconnect error handling for pub/sub and DB selection; failures now emit `error` instead of crashing, but the client may continue on the wrong DB if SELECT replay fails (same behavior as handshake SELECT already had).
> 
> **Overview**
> After reconnect, **`readyHandler`** replays `SELECT`, `SUBSCRIBE`, `PSUBSCRIBE`, and `SSUBSCRIBE` (plus `SELECT` in the resend/offline-queue paths) without handling promise rejections. Failures (e.g. NOPERM) could become **unhandled rejections** and crash Node 15+ even when the client has an `error` listener.
> 
> This PR attaches **`.catch((err) => self.silentEmit("error", err))`** to those seven call sites so reconnect replay matches the handshake path for `SELECT`.
> 
> New unit tests in **`test/unit/resubscribe.ts`** cover failing replay for SUBSCRIBE, PSUBSCRIBE, and SELECT (subscriber and non-subscriber paths) under RESP2 and RESP3, asserting no unhandled rejection and that the error event carries the server message. **`spub_ssub.ts`** stubs **`ssubscribe`** with **`.resolves()`** so it behaves like the real async API during auto-resubscribe tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e9758f634b5c2e062d902c3c86332029db012fd9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


test/unit/resubscribe.ts
```diff
@@ -0,0 +1,144 @@
...
+      return "OK";
+    });
+    server.on("connect", () => connections++);
+    return server;
+  }
+
+  function client(port: number) {
+    return new Redis({
+      port,
+      protocol: 2,
```
PavelPashov: Is there a reason these tests explicitly use RESP2? Would it make sense to cover RESP3 as well, since it’s now the default?

PavelPashov (approved):

test/unit/resubscribe.ts
ramin-010: Covered both protocols now — details in the comment below.

PR #2191, opened 2026-09-10
fix(types): add smessage and smessageBuffer event overloads

### LINKED ISSUE #2190: smessage and smessageBuffer are not typed on Redis#on / Redis#once
**Versions:** ioredis `main` (6.0.0), Node.js 22, Redis 7.x, standalone

Sharded pub/sub listeners get untyped callback parameters, while the equivalent non-sharded events are typed:

```ts
redis.on("message",  (channel, message) => {}); // channel: string, message: string
redis.on("smessage", (channel, message) => {}); // channel: any,    message: any
```

**Expected:** `smessage` gives `(channel: string, message: string)` and `smessageBuffer` gives `(channel: Buffer, message: Buffer)`, matching the `message` / `messageBuffer` pair.

**Actual:** both parameters are `any`, so there is no type safety or completion.

`DataHandler` emits `smessage` and `smessageBuffer`, but the `on` / `once` overloads on `Redis` only cover `message`, `messageBuffer`, `pmessage` and `pmessageBuffer`. The sharded events fall through to the `(event: string | symbol, listener: (...args: any[]) => void)` signature.

This may have gone unnoticed because the type tests for that surface have never run: `tsd` is configured with `directory: "test/typing"` and only collects `*.test-d.ts`, but the file is named `events.test-.ts`, so it has been inert since #1694 added it. Renaming it makes the existing assertions execute again; I checked and they pass unchanged.

Happy to open a PR adding the two overload pairs plus the rename so the new assertions actually run. Would you prefer that as one PR or two?

ramin-010:
Closes #2190.

Adds `smessage` and `smessageBuffer` overloads for `on` and `once`, with
assertions for both callback parameter types.

Without the overloads `npm run test:tsd` reports 8 errors of the form
"not identical to argument type `any`"; with them it passes.

Also renames `events.test-.ts` to `events.test-d.ts` for consistency with the
other files in that directory — as you noted, tsd already collects it, so this
activates nothing.

---

> [!NOTE]
> **Low Risk**
> Type-only changes and typing tests; no runtime subscriber or emit behavior is modified.
> 
> **Overview**
> Adds **TypeScript event overloads** on `Redis` for `on`/`once` when listening to **`smessage`** and **`smessageBuffer`**, matching the existing `message` / `pmessage` patterns (string vs `Buffer` channel and payload).
> 
> Extends **`test/typing/events.test-d.ts`** with `tsd` assertions so subscriber callbacks for those events are typed as `string` or `Buffer` instead of falling through to `any` and failing `test:tsd`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba9d3c7cd8e8cfacc87bea94d5b9479a39ea1422. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PavelPashov (approved):
Looks good, thank you for the contribution.

PR #2192, opened 2026-09-10
fix(cluster): recover from failed readonly role changes

MarthalaJagruthiReddy:
Fixes #2086

## Summary

I fixed a Redis Cluster failover issue where a node could remain in the pool after failing to enter READONLY mode, causing stale read routing and repeated MOVED errors.

The change:

- Fails connection setup when the READONLY handshake fails instead of exposing the node as ready.
- Disconnects a pooled node when a runtime READONLY or READWRITE role switch fails.
- Adds standalone, unit, and cluster regression tests covering loading/failover behavior.

## Validation

I ran:

- `npm run build`
- `npm run lint`
- `npm run test:tsd`
- Focused unit and mock-server suites for standalone and cluster behavior.

Docker/real-Redis suites were not run because Docker and Redis are unavailable in the workspace.


nkaradzhov (changes requested):
Thanks — this fixes both swallow sites from #2086 and matches what redis-rs, go-redis, redis-py, and node-redis do. Good test coverage.

Two changes before merge:

1. In ConnectionPool, the new .catch calls redis.disconnect(true) without a guard. A stale rejection can disconnect a node whose newer role change already succeeded. Return early when the connection left the pool or redis.options.readOnly changed, and add a regression test.
2. readOnly: true against a non-cluster server now reconnects forever on "ERR This instance has cluster support disabled". Treat that reply as warn-and-continue, like handleAuthError does for benign auth errors.

PR #2194, opened 2026-09-11
fix: reject unfulfilled commands dropped on reconnect

### LINKED ISSUE #2193: Command promise never settles after reconnect when autoResendUnfulfilledCommands is false
**Versions:** ioredis `main` (6.0.0), Node.js 22, Redis 7.0, standalone

When `autoResendUnfulfilledCommands: false` is set, a command that is in flight when the
connection drops is never settled. The promise does not resolve or reject, and
`redis.disconnect()` does not settle it either — it hangs for the lifetime of the process
while `redis.status` reads `ready`.

**Expected:** the command rejects, as it does when the connection is closed by other means.

**Actual:** the promise stays pending forever.

**Reproduction** (real Redis on 127.0.0.1:6379):

```ts
import Redis from "ioredis";

async function run(autoResend: boolean) {
  const redis: any = new Redis({ autoResendUnfulfilledCommands: autoResend });
  redis.on("error", () => {});
  await redis.ping();

  let settled = "PENDING";
  redis.blpop("nonexistent-key", 30).then(
    (v: any) => (settled = "resolved " + JSON.stringify(v)),
    (e: Error) => (settled = "rejected: " + e.message)
  );
  await new Promise((r) => setTimeout(r, 200));

  redis.stream.destroy();                        // connection drops mid-command
  await new Promise((r) => setTimeout(r, 1500)); // reconnect completes
  const q = redis.commandQueue.length;

  redis.disconnect();                            // should settle anything tracked
  await new Promise((r) => setTimeout(r, 500));

  console.log(`autoResend: ${autoResend} | commandQueue after reconnect: ${q} | blpop: ${settled}`);
}

(async () => { await run(true); await run(false); process.exit(0); })();
```

Output:

```
autoResend: true  | commandQueue after reconnect: 1 | blpop: rejected: Connection is closed.
autoResend: false | commandQueue after reconnect: 0 | blpop: PENDING
```

After the reconnect the command is in `commandQueue` (0), `offlineQueue` (0) and
`prevCommandQueue` (null) — nowhere, so nothing can reach it.

On close from `ready`, the in-flight queue is stashed as `prevCommandQueue`. On the next
`ready`, if `autoResendUnfulfilledCommands` is false it is discarded with
`self.prevCommandQueue = null` without rejecting, and `flushQueue` only walks `offlineQueue`
and `commandQueue`. A successful TCP reconnect is required — before `resetCommandQueue()`,
`prevCommandQueue` is the same array as `commandQueue`, so `flushQueue` still reaches it.
That is why the server-never-returns case rejects correctly via `maxRetriesPerRequest`.

Happy to open a PR rejecting the stashed commands rather than dropping them. What error would
you want them rejected with — `Connection is closed.`, as `disconnect()` uses, or something
more specific to the dropped-resend case?

ramin-010:
Closes #2193.

On close from `ready`, in-flight commands move to `prevCommandQueue`. A successful
reconnect replaces `commandQueue`, so with `autoResendUnfulfilledCommands: false`
the ready handler dropped the stash without rejecting it, and `flushQueue` only
walks `offlineQueue` and `commandQueue` — leaving those promises unsettleable, even
by `disconnect()`.

Each stashed command is now rejected with `abortError`, as you suggested, so they
fail the same way as the other connection-drop paths.

Regression test added to the existing `autoResendUnfulfilledCommands` block in
`test/functional/connection.ts`: an in-flight `BLPOP`, a destroyed stream, and an
assertion that the promise rejects after reconnect. On `main` it times out at 8s;
with the change it passes in ~250ms.


nkaradzhov (approved):
Thanks for the thorough issue and the fix, LGTM.
I would still defer the final decision to @PavelPashov

PavelPashov (approved):

PR #2198, opened 2026-09-17
fix: clean up Sentinel failover subscriptions after quit

### LINKED ISSUE #1362: 4.27.0 introduced a change that prevents the event loop from being cleaned up properly

maximilliangrand:
Closes #1362.

With Sentinel `failoverDetector: true`, `quit()` acknowledges queued commands and closes the Redis connection, but its separate Sentinel subscriptions remain connected. These sockets can prevent jobs and workers from exiting.

Tie each failover detector's cleanup to its owning Redis stream's `close` event, capturing that detector before awaiting subscription setup. Make cleanup idempotent so explicit disconnect, failover, and stream closure can converge without repeatedly disconnecting already closed subscription clients. The command connection still completes its normal QUIT handshake.

Regression tests use the existing local TCP Sentinel/master fixtures with RESP2 and RESP3. Both protocol cases fail on the unchanged base: GET succeeds, QUIT returns `OK`, Redis reaches `end`, but two Sentinel subscription sockets remain open. Assertions run before fixture teardown, so teardown cannot conceal the leak.

Coverage also includes QUIT while the subscription acknowledgment is pending,
repeated cleanup after failover, and the existing failover/reconnection cases.
All five new regression tests fail against unchanged base `d95d05a`; all pass
with this fix.

Validation on Node.js 26.7.0, macOS arm64:

- 36 Sentinel functional and SentinelConnector unit tests passed, using local TCP fixtures. The suite was independently replayed with the same result.
- `npm run build` and `npm run test:tsd`: passed.
- `npm run lint`: zero errors; 23 warnings in unchanged files.
- Changed-file Prettier checks and `git diff --check`: passed.
- Independent review of ownership, pending subscription and repeated cleanup: no blocking findings.

For the isolated fixture run, a local Mocha hook supplied the repository's
`sinon.restore()` cleanup without loading the general helper that clears a Redis
database. The full real-Redis and Cluster matrix was not run locally and remains
for CI.

AI assistance was used for reproduction, implementation, tests, and independent lifecycle review.

---

> [!NOTE]
> **Medium Risk**
> Changes Sentinel failover lifecycle and shutdown wiring; scope is limited to optional failoverDetector mode with idempotent cleanup.
> 
> **Overview**
> Fixes a connection leak when Sentinel **`failoverDetector: true`** clients call **`quit()`**: the main Redis stream closes after QUIT, but separate Sentinel **`+switch-master`** subscription sockets stayed open and could block process shutdown.
> 
> **`SentinelConnector`** now hooks each detector instance to its owning command stream’s **`close`** event so subscription clients are torn down when QUIT closes the connection (a path that does not go through the connector’s **`disconnect()`**). **`FailoverDetector.cleanup()`** is **idempotent** via an **`isCleanedUp`** guard so failover-driven disconnect, explicit cleanup, and stream closure can all run without double-disconnecting sentinel clients.
> 
> New **RESP2/RESP3** functional tests assert subscriptions close after quit (including while subscribe is still pending), plus a unit test that repeated cleanup after failover only disconnects once.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8a32d3d7936e3a93224d1282ab7a2cf6b9bd9ff7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PavelPashov (approved):
@maximilliangrand Looks good, thank you for the contribution

PR #2199, opened 2026-09-17
fix: preserve empty keys and match patterns in scan streams

jakezwang:
Scan streams drop empty string keys and `MATCH` patterns because they check whether the values are truthy. As a result, `hscanStream('')` sends an invalid command, and `scanStream({ match: '' })` scans every key instead of matching the empty key.

Check for `undefined` so empty strings reach Redis. Add regression tests for all six keyed stream variants and both SCAN variants.

Validation: the eight regression tests fail before the fix. `npm test` passes against Redis 8.4.0 (2,409 tests, 290 pending, plus type tests); lint, build, and formatting checks for the changed files pass. The eight scenarios also pass against a real Redis instance.

---

> [!NOTE]
> **Low Risk**
> Narrow behavior fix in scan stream argument building, covered by new regression tests; no auth or data-path changes.
> 
> **Overview**
> **Scan streams** no longer treat empty `key` or `match` values as “missing.” `ScanStream` now checks `key` and `match` with `!== undefined` instead of truthiness, so `""` is forwarded to Redis (e.g. `hscanStream('')` and `scanStream({ match: '' })`).
> 
> Adds unit tests that stub the underlying scan commands and assert the argument lists for `hscan`/`sscan`/`zscan` (string and Buffer streams) and `scan`/`scanBuffer` with an empty `MATCH`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a775f28590bcf849e3dc0382951260c74c29540f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PavelPashov (approved):
@jakezwang looks good, thank you for the contribution.

